PR #6, opened 2014-08-03
Send only updated/changed torrents to the client

bendikro:
This branch contains the following modifications:
- Client can request to get only the torrents that were updated since last time
  - This reduces the amount of network data that needs to be transfered between the client and server greatly. With thousands of torrents, the overhead of processing the entire torrent list to be sent and received regularly causes the major part of the CPU usage as seen by profiling.
- The filtering of the torrents has been split between the server and client. With thousands of torrents, the workload of filtering the torrents each time the list is sent becomes a performance bottleneck. This has been changed so that the server only filters on the owner to make sure only the torrents the user is allowed to fetch are returned. The rest of the filtering is performed on the client side.
  - Much faster to change the list on the client side as it no longer has to ask the server.
  - The filtering functionality is implemented with PyDbLite.
    - This should make it possible to implement advanced filtering quite easily.
    - The filtering is much more efficient than the current code
- [ ] Write more unit tests
- [ ] The code changes for the webui should work but the code should be reviewed

Feedback is very appreciated. :wink:  @gazpachoking

Note: PyDbLite is a separate project: https://pypi.python.org/pypi/PyDbLite
# Changes on the server side
### deluge/core/core.py

Changed functions:
- **get_torrents_status:** 
  - Added keyword argument **only_updated** which makes **get_torrents_status** only return the torrents that have been updated since the last call.
  - Argument **filter_dict** is now obsolete.
  - Argument **diff** is now obsolete.
    Removed functions:
  - **get_filter_tree** - The filtertree is now created on the client side
### deluge/core/filtermanager.py

Most of the functionality of `filtermanager.py` has been moved to the client side (`torrentfilter.py`). `filtermanager.py` now only filters on the owner so the client will only receive the torrents it is authorized to see.
It filters on owner using the new databases in deluge/include/pydblite.
### deluge/core/rpcserver.py

Changed functions:
- **connectionLost**
  - Now emits `ClientDisconnectedEvent` when client disconnects
### deluge/core/torrent.py

Changed functions:
- **set_tracker_status**
  Now emits `TorrentTrackerStatusChangedEvent` when torrent status is set
### deluge/core/torrentmanager.py

Added **status_dicts** which contains all the torrents for each session. When a torrent is updated (either by libtorrent or a plugin) it is removed from the cache to mark it as updated.

Changed functions:
- **handle_torrents_status_callback**
  - Creates a status based on the **only_updated argument**, so if **only_updated** is True, only the torrents that do not exist in the cache (for this session_id) are returned.
- **on_alert_state_update**
  - Removes the updated torrents from the cache, and emits `TorrentsUpdatedEvent`

New Functions:
- **update_torrent_cache(torrent_ids)**
  - Removes the **torrent_ids** from the cache
# Changes on the client side
### deluge/ui/gtkui/torrentview.py

Changed to now only update the necessary rows.
The main difference is that **self.status** now always contains all the torrents, not only the torrents that are displayed. Which torrents that are currently displayed is stored in the `set` `self.visible_torrents` which is updated in **_on_get_torrents_status**.

Changed functions:
- **on_get_torrents_status**
  - Updates self.visible_torrents no contain the currently visible rows.
- **update_view**
  - Added arguments:
  - **only_updated_ids**
    - The torrent_ids that have been updated with new data. If None, all have been updated.
  - **to_show**
    - The **torrent_ids** to show (change filter column value to True)
  - **to_hide**
    - The **torrent_ids** to hide (change filter column value to False)

Removed functions:
- on_torrentstatechanged_event
  - No longer needed as this is handled by sessionproxy.
- **mark_dirty**
  - The dirty functionality was not used anymore (since the last major torrentview changes) so it was completely removed. 
### deluge/ui/sessionproxy.py

Has been completely changed. It no longer keeps track of cache times for each value. This was extremely expensive, and is no longer needed as the torrents that have been updated on the server side are now updated automatically.
**self.updated_ids** contains the torrent ids that have been updated from core and not yet updated in the ui.

All of the filtering previously performed in filtermanager is now performed by **self.torrentfilter**. This reduces the workload for the daemon tremendously and makes it much faster on the client side to get the filtered results.

Changed functions:
-  **get_torrents_status**
  - The biggest change is that it now takes the status dictionary as input and updates the status with the torrents as needed. It returns which torrents were updated and which torrents that should be shown or hidden in the view.

New arguments:
- **torrents_state**
  - Contains state info for the client 
- **only_updated=False**
  - Give only the updated torrents
- **from_cache=True**
  - If it should wait for the data from core, or return data from the cache.
### deluge/ui/torrentfilter.py

This does all the filtering on the client side which was previously done in `filtermanager.py` on the server side.
`torrentfilter.py` uses PyDbLite (which uses SQLite if available, else pure Python) to perform all the filtering. The filtering code should be cleaner and simpler than the old code, and also _much_ more efficient with many torrents.
### deluge/plugins/Label/deluge/plugins/label/{core.py,gktui/**init**.py}

Changed to work with the new filtering

Changed to handle “hide owners” in sidebar:
deluge/ui/gtkui/filtertreeview.py
deluge/ui/gtkui/glade/main_window.ui
deluge/ui/gtkui/menubar.py
deluge/ui/web/server.py
deluge/ui/gtkui/gtkui.py
# Changes for both client and server
### deluge/event.py
- Added events
  - TorrentTrackerStatusChangedEvent
  - ClientDisconnectedEvent
## PyDbLite for efficient filtering

Added files:
- deluge/filterdb.py
  - Contains Classes to create and setup the PyDbLite database

### DIFF
--- a/deluge/tests/test_json_api.py
+++ b/deluge/tests/test_json_api.py
@@ -10,7 +10,7 @@
 import json as json_lib
 
 from mock import MagicMock
-from twisted.internet import defer
+from twisted.internet import base, defer
 from twisted.web import server
 from twisted.web.http import Request
 
@@ -29,6 +29,8 @@ from .common_web import WebServerMockBase
 from .daemon_base import DaemonBase
 
 common.disable_new_release_check()
+
+base.DelayedCall.debug = True
 
 
 class JSONBase(BaseTestCase, DaemonBase):


deluge/core/torrentmanager.py
```diff
@@ -1213,11 +1267,49 @@ def torrents_status_update(self, torrent_ids, keys, diff=False):
...
         return d
+
+    def on_client_disconnected(self, session_id):
+        """
+        Clean up any torrents status requests stored for this session id
+
+        """
+        i = 0
+        removed = 0
+        while i < len(self.torrents_status_requests):
```
cas--: I would tend to use a for loop with range here and no need for `removed` as can be `orig_len - new_len`

deluge/core/torrentmanager.py
```diff
@@ -1166,6 +1182,21 @@ def on_alert_performance(self, alert):
             else:
                 log.warning("send_buffer_watermark reached maximum value: %s Bytes", max_send_buffer_watermark)
 
+    def clear_from_torrent_cache(self, torrent_ids):
+        """Remove the torrent ids from self.status_dicts
+        This indicates that the torrent id has been updated.
+
+        Accepts torrent id (str) or a list of torrent ids
```
cas--: Should be only one type accepted so `list` would be fine.

deluge/core/torrentmanager.py
```diff
@@ -1166,6 +1182,21 @@ def on_alert_performance(self, alert):
             else:
                 log.warning("send_buffer_watermark reached maximum value: %s Bytes", max_send_buffer_watermark)
 
+    def clear_from_torrent_cache(self, torrent_ids):
```
cas--: I am not sure the method name is obvious for it's task.

deluge/core/torrentmanager.py
```diff
@@ -192,8 +198,8 @@ def archive_file(filename):
 
         # Save the state periodically
         self.save_state_timer.start(200, False)
-        self.save_resume_data_timer.start(190, False)
-        self.prev_status_cleanup_loop.start(10)
+        self.save_resume_data_timer.start(190, False)  # The resume data for torrents that need it
+        self.save_all_resume_data_timer.start(900, False)  # Save resume data for all torrents
```
cas--: These comments should be moved to where the timers are defined. Should actually be in init docstring but there isn't one yet it...

deluge/core/torrentmanager.py
```diff
@@ -119,6 +120,7 @@ def __init__(self):
 
         self.torrents_status_requests = []
         self.status_dict = {}
+        self.status_dicts = {}
```
cas--: This will need a comment, also is the naming going to cause confusion?

deluge/event.py
```diff
@@ -104,6 +104,20 @@ def __init__(self, torrent_id, state):
         self._args = [torrent_id, state]
 
 
+class TorrentTrackerStatusChangedEvent(DelugeEvent):
```
cas--: Maybe could be just `TorrentTrackerStatusEvent`?

deluge/event.py
```diff
@@ -104,6 +104,20 @@ def __init__(self, torrent_id, state):
...
+    """
+    Emitted when a torrents tracker status changes.
+    """
+    def __init__(self, torrent_id, status):
+        """
+        :param torrent_id: the torrent_id
+        :type torrent_id: string
+        :param status: the new status
+        :type status: string
```
cas--: Need to use new style docstring...

deluge/core/torrentmanager.py
bendikro: Agreed

deluge/core/torrentmanager.py
bendikro: for loop won't work as `self.torrents_status_requests` is changed inside the loop. The while is needed to avoid incrementing `i` when an element is removed. `orig_len - new_len` is a good point.

cas--:
```
orig_len = len(self.torrents_status_requests)
for i in range(orig_len):
...
```

deluge/event.py
bendikro: Ah, yeah, those doc strings :-)

deluge/event.py
bendikro: Renamed

deluge/core/torrentmanager.py
bendikro: Moved into separate class

deluge/core/torrentmanager.py
bendikro: Done

deluge/core/torrentmanager.py
bendikro: Renamed to mark_dirty

deluge/core/torrentmanager.py
bendikro: Implemented clearer while with if/else in function clear_status_requests

deluge/core/torrentmanager.py
```diff
@@ -1223,47 +1213,176 @@ def separate_keys(self, keys, torrent_ids):
...
+    def torrents_status_updated(self, updated_ids):
+        """
+        Update with new status from libtorrent
+
+        """
+        self.last_state_update_ts = time.time()
+        self.tm.torrents_status.mark_dirty(updated_ids)
+        self.handle_torrents_status_callback()
+
+    def handle_torrents_status_callback(self, request=None):
```
cas--: Do you think the name of this method `handle_torrents_status_callback` would read better as something like `request_torrents_status`? Looking at it's usage, it's not a specific handle/callback method.

deluge/core/torrentmanager.py
```diff
@@ -1223,47 +1213,176 @@ def separate_keys(self, keys, torrent_ids):
...
+
+        self.torrents_status.torrents_status_updated(updated_ids)
+
+    def torrents_status_update(self, torrent_ids, keys, only_updated=False):
+        return self.torrents_status.torrents_status_update(torrent_ids, keys, only_updated)
+
 
-        # Get the torrent status for each torrent_id
+class TorrentsStatus(object):
+
```
cas--: Handy to have a docstring summarising this class.

bendikro: Done

deluge/ui/web/js/deluge-all/Deluge.js
```diff
@@ -126,7 +126,7 @@ Ext.apply(Deluge, {
         var progressWidth = ((width / 100.0) * progress).toFixed(0);
         var barWidth = progressWidth - 1;
         var textWidth = ((progressWidth - modifier) > 0 ? progressWidth - modifier : 0);
-        return String.format(Deluge.progressTpl, text, width, barWidth, textWidth);
+		return String.format(Deluge.progressTpl, text, width, barWidth, textWidth);
```
cas--: extra indent

deluge/ui/web/json_api.py
```diff
@@ -251,10 +252,13 @@ def _on_json_request_failed(self, reason, request):
...
-        response = json.dumps(response)
-        request.setHeader("content-type", "application/x-json")
-        request.write(compress(response, request))
-        request.finish()
+        try:
+            response = json.dumps(response)
+            request.setHeader("content-type", "application/x-json")
+            request.write(compress(response, request))
+            request.finish()
+        except TypeError, e:
```
cas--: If the TypeError come from json.dumps then try should only wrap that with request in else. Also old style exception, needs `as ex` for develop branch.

deluge/ui/web/js/deluge-all/Deluge.js
bendikro: Emacs defaults to tabs, so tried to fix it manually, but missed this one.

deluge/ui/web/json_api.py
bendikro: Done

cas--:
Can you add pydblite to DEPENDS file

deluge/ui/web/js/deluge-all/UI.js
```diff
@@ -157,7 +158,20 @@ deluge.ui = {
         this.oldFilters = this.filters;
         this.filters = filters;
 
-        deluge.client.web.update_ui(Deluge.Keys.Grid, filters, {
+        //var keys = Deluge.Keys.Grid
```
cas--: delete?

deluge/ui/web/json_api.py
bendikro: request.finish() can actually raise a RuntimeError, so changed this in latest update.

DEPENDS
```diff
@@ -20,6 +21,7 @@
  * python-notify (optional)
  * pygame (optional)
  * python-appindicator (optional)
+ * pydblite
```
cas--: This line is not needed, it's already a requirement for common/base files

bendikro: Fixed

cas--:
I have cherrypicked the update_state commit

cas--:
Rebased and squashed fixups upon develop

PR #8, opened 2014-08-21
Feature/develop/2122 2133/listen on changes

cas--:
This is an unfinished branch but looking for input on the direction of the changes. 

There are two detailed tickets that are required reading to understand the changes:

http://dev.deluge-torrent.org/ticket/2133
http://dev.deluge-torrent.org/ticket/2122

The code changes I have made are straightforward and don't affect the UIs at all but I have some unanswered questions...
- Do we need to expose `listen_use_sys_port` and/or `listen_reuse_port` to the user? 
- I included some basic logic to reuse port if only a single port range is supplied. This might be unwanted functionality.
- Should we remove the port range as suggested in the tickets and only have a single port entry?
- Do we also get rid of the random port option and simply generate a port if `listen_ports` is empty or zero?

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -59,16 +59,18 @@ DEFAULT_PREFS = {
     "download_location": deluge.common.get_default_download_dir(),
     "listen_ports": [6881, 6891],
     "listen_interface": "",
-    "listen_use_sys_port": True,
-    "listen_reuse_port": False,
+    "random_port": True,
+    "listen_random_port": None,
+    "listen_use_sys_port": False,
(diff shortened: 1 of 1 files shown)


SamB:
... Wouldn't you just want a text field to allow entering a specific port number, plus a button to set it to a new random port?

sjnewbury:
Maybe I'm mistaken, but doesn't specifying a port range allow parallel incoming connection handling?

cas--:
Would need to clarify with libtorrent dev but the previous versions only listened on one port iirc. The UI settings were changed to simplify things for the user. It is still possible to specify a range of ports in the core.conf.

PR #15, opened 2014-10-02
Added right click popup in list of torrents to add (to copy torrent title)

bendikro:
Right clicking a torrent in the list of torrents to add will popup a
menu that allows the user to copy the torrent title to the clipboard.


cas--:
I added the editable cell property instead: 9e13f671ee8f86c0fa85f68df251e11243dcef18

PR #21, opened 2014-11-02
[GTKUI] optimize adding torrents with thousands of files

rmmh:
Adding a torrent with ~9200 files in one folder was painfully slow. Most of this time was spent constructing the TreeStore.

Before: 1m20s 
After: 3s

Ticket: [#2670](http://dev.deluge-torrent.org/ticket/2670)


netswap:
I am afraid that I should not post or request something here, as nor can I register on the forum, neither can I register on Deluge's Trac...
...but: How is the state of this fix? Might we see this fix being introduced in a near-future update? Many people, including me, have been waiting for this for quite some time -- personally this is *the* reason why I switched clients, as I love my torrents filled with thousands of images to be loaded faster than now, waiting hours to begin the transfer, and hours to see the file list. :/

rmmh:
This was merged into the develop branch. It could be cherrypicked into the 1.3-stable branch very easily, since the patch is minimal. @cas-- would you accept a backport PR?

PR #26, opened 2015-02-19
[#2733] [Core] Fix on_alert_performance - UnicodeDecodeError

bendikro:
Fix for http://dev.deluge-torrent.org/ticket/2733#ticket


cas--:
Thank you

PR #28, opened 2015-05-04
Added quotes around the path arg for explorer.exe

bertyhell:
If the filename has spaces this otherwise causes the file not to be selected (untested)


cas--:
If this is untested why is it being submitted?

bertyhell:
Cause I couldn't get the app to compile

cas--:
I cannot see any issues with selecting filenames with spaces, please open a bug report with traces/logs.

PR #29, opened 2015-05-15
[2750] Fixed: Torrents forget paused state after a force recheck.

Kreidos:
Fix for ticket #2750: http://dev.deluge-torrent.org/ticket/2750

Solves the issue of torrents not returning to paused state after a force re-check. Possibly due to an underlying issue with libtorrent, but this seems to get around the problem.


cas--:
This need more investigation because all that is happening is calling handle.pause twice so not merging this particular change.

PR #30, opened 2015-06-22
[#2295] [WebUI] Increased lifespan of display settings

bendykst:
Display settings for the WebUI are persisted using cookies created by
Ex.state.CookieProvider. When no expiration date is provided, a default
value of (now + 7 days) is used. This causes display settings to be
lost frequently.

This fix adds an 'expires' parameter with a value of (now + 10 years).
This change does not affect the lifespan of the session cookie, which
is created by a separate system.

### DIFF
--- a/deluge/ui/web/js/deluge-all/Deluge.js
+++ b/deluge/ui/web/js/deluge-all/Deluge.js
@@ -31,7 +31,14 @@
  */
 
 // Setup the state manager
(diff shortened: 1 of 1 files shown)


bendikro:
Sounds reasonable. @cas-- ?

cas--:
This is a good idea, thanks. Merged: a39ebae0cd43b979128fb7b59922e2376ec42235

PR #31, opened 2015-06-24
Update gitattributes and ignore

doadin:
The win32/osx folder is needed for building and should be included.
However add .tmp to ignored list so that a file made during build does
not get included. Same with the web folders.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,3 @@
-/win32/ export-ignore
-/osx/ export-ignore
 docs/build/ export-ignore
 docs/source/ export-ignore
 /tests/ export-ignore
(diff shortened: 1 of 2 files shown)


cas--:
Sorry those files will not be included as export is for source tarball creation, packaging scripts are not part of that.

doadin:
what about the webui part eg. deluge/ui/web/js/deluge-all/ export-ignore . That is source?

cas--:
Yes but it is built before export usually, it's not ideal and I will look to improve that.

doadin:
Thank you. I know for me it would be easier to be able to have a export with all the files needed. I can see how not including the packaging parts makes sense but the webui is different story. Thank you for your consideration.

PR #32, opened 2015-07-04
[#2743][GTKUI] Convert  to pygi/gtk3

doadin:
Packaging for windows still needs to be fixed but is the last item i found in my testing.


cas--:
I see a lot of commented out code and TOFIX so this is not ready to merge. There is also no commit history but need to know why certain changes were made, especially those not directly related to conversion. In addition it is failing TravisCI...

bendikro:
@doadin All commented out code needs to be removed, so that the diff only shows the changes that are supposed to be there. After that, it's easier to go over and look at the changes and see if anything should be done differently.

One thing I'm wondering is how good the support for pygi and GTK3+ is on older systems. It seems like there are differences between Ubuntu 12.04 that travis-ci uses, and Ubuntu 14.04. So before pygi conversion can be merged in, we need to know what systems the next major Deluge version should support, and what the status is for pygi support on different systems.

cas--:
Just removing commented out code is not the solution, why is it commented
out and what is the replacement code...

There are other more immediate questions such as does this break all 3rd
party plugins?

doadin:
I had squashed the billion commits i had made to make this cleaner. The flake8 fail is due to a missed squash and is a easy to resolve fail. The other two test fails are due to errors in the tests themself as me and bendikro have found you can see in his fork has has made adjustments toward fixing them. These error are also in the current dev branch, hence why they also show here. This commit is to deluge code it self no to travisci tests. And i consider that seperate kind of like you consider the packaging different for the source tarball. It might be there but its not my concern. However i would be happy to continue to work with you and bendikro to fix them but i think this should be a different pull request as it is a seperate issue. As this code for the majority only affects code in the gtkui folder obviously since its a change in the gtkui... 

Some code just needs to be removed some needs to be tested on other platforms hence why i have pull requested. I have tested the client pretty thoroughly on windows and have found no additional errors in the the process going from dev branch this is based on to this commit/pygi/gtk3. Both versions of the gtkui work 100% the same on windows. If you would also have read through it you would notice some of the commented out code has replacement already and again sorry but i missed a bit of cleanup in the process of squashing. 

Of all the commented code id say 1/3 of it can just be removed as it does already have replacement and just needs to be cleaned, another 1/3 is code that has comment to it and needs review as the comment says, and the last 1/3 is code that is for other platforms and again needs review.

Obviously the only third party plugins it breaks is ones with gtkui's. And even then its a small update in most cases just to reflect the change in imports. As you can notice in the changes to the default plugins because i did update those so they work with the convert as they are a part of deluge. I would be 100% willing to go through current third party plugins and work with any dev this is no longer working on plugins or needs help trying to figure out how to update.

With the issue of third party plugins in mind maybe you want to put off merging this change till like deluge 2.0 so devs have time to update and people can be made aware of this change. I simply made this pull request because its 100% in a working state on the platform i tested on and got 0 feedback from the ticket i opened. Dispite having mentioned  part of the concerns you brought up here. Other than bendikro im not sure what other devs are here to help with testing and as windows is the only platform i can test on currently this is all i got. Again this works 100% the same as the current dev branch and yes there are still some issues i can tell you that right now however again these are issues that are in current code and not due to this convert and needs to be addressed else where.

Again ill appologize upfromt if this seems like a rant but seems like my ticket could have been answered and resolved part of this. Even though i still dont mean to rant. I just want to cover all points. And get the feed back i was looking for. Like always though i appreciate the comments and the help pointing out issues i missed in this commit whether its just code i didnt clean or something i missed.

PR #33, opened 2015-07-15
Develop tests travis xvfb isort trackericon

bendikro:
Fixes for tests:
* Xvfb works again
* Updates for latest isort (4.0.0)
* Tracker icon tests fixed


cas--:
Thanks, there is a bug in those new isort options so we shall continue skipping gtkui.py :)

PR #34, opened 2015-07-15
Update Translations

doadin:


bendikro:
@doadin Have these been translated with an automated translation tool, such as google translate?

doadin:
something like that. with poedit which uses a mix of machine and human translation and already tanslated words from the translation file.

doadin:
I was just testing probably don't want to accept this at least not some of it. Some of the translations are basically non-existent and imo should just be removed. 3% or less translated is not good.Anyways probably not the best translations but idk how you guys feel about it but something is better than nothing imo if youre gonna keep the translation as an option.

bendikro:
There is no point is having a translation for a language that doesn't cover larger (most) parts of the text. Also, I don't think it's a good idea to use automated translation without manually being able to verify that the result is correct. It's better to have a proper English fallback text on phrases that aren't translated than to have a badly translated phrases.

Also, you keep pissing off cas every time you create a pull requests for branches that aren't ready for merging. It's better to create a ticket (http://dev.deluge-torrent.org/wiki/Development#BugTracking) and link to your github branch if you have questions :-)

doadin:
yea i meant to publish this one not push i instantly was like ut oh...he gonna hate me even more.

cas--:
All translation is done through Launchpad: http://dev.deluge-torrent.org/wiki/Development/Translation

I have just uploaded the latest template for develop here: https://translations.launchpad.net/deluge/develop

pereorga:
Hi @cas-- , sorry for hijacking this thread, there is no information in the readme and the link [0] in the documentation [1] is broken. 

I would like to contribute to the Catalan translation. Is Launchpad still used for translations? The translation file is very outdated, for example, "Files" is not translated in https://github.com/deluge-torrent/deluge/blob/develop/deluge/i18n/ca.po#L4233 but the translation exists in Launchpad since 2007: https://translations.launchpad.net/deluge/develop/+pots/deluge/ca/1188/+translate

[0] https://deluge.readthedocs.io/en/develop/contributing/translations.html
[1] https://dev.deluge-torrent.org/wiki/Development

cas--:
@pereorga Yes. The deluge 2.0 translation files are rather outdated and should be updated using 1.3 as a basis. I have not had the time to sort it out but you can update Catalan

I have updated the documentation link.

pereorga:
Great!

Should I work in Launchpad then? It's a bit confusing because current Launchpad data does not match these files. I'm also not familiarized with the workflow followed in Launchpad, but ideally, I would like to compile and open Deluge with the current and updated translation and start improving the translation from there.

cas--:
> It's a bit confusing because current Launchpad data does not match these files.

Which file? The git repository is always behind Launchpad as we need to wait for translations: https://deluge.readthedocs.io/en/latest/contributing/translations.html#translation-process

The Launchpad files can be exported: https://translations.launchpad.net/deluge/2.0/+pots/deluge-2.0/+export

pereorga:
Nevermind, I see that you just updated some translations Launchpad. I'll work there, thanks!

pereorga:
I just completed the Catalan translation: https://translations.launchpad.net/deluge/2.0/+pots/deluge

@cas-- please, let me know if anything else is needed.

PR #38, opened 2015-08-29
Cleaned up code

dvska:


.idea/dictionaries/dvs.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
```
cas--: Has this file been included by accident?

cas--:
Thanks for this, quite helpful. 

The only thing I would change is the use of the staticmethod decorator in plugins. I don't see any benefit to using it.

PR #40, opened 2015-08-31
Prevent crashes on FIPS 140-2 compliant systems

caspear:
Replace weak hashing functions, key sizes, and random number
generation techniques with less weak versions to prevent
crashes when running with the fips module loaded.

Specifically:
Stop using md5 wherever possible. It is a deprecated hash.
Stop using 1024bit RSA keys. Things are unlikely to support such short lengths for much longer.
Stop using 5 year expiry. 3 years and 3 months will soon be the longest accepted expiry time of a certificate.
Do not use random as a source of crypto-strength random data. Use os.urandom instead.
Use 32 bytes of random data instead of 5 when creating a session key. This should make it a little harder to guess.

### DIFF
--- a/deluge/core/rpcserver.py
+++ b/deluge/core/rpcserver.py
@@ -530,10 +530,10 @@ def generate_ssl_keys():
     """
     This method generates a new SSL key/cert.
     """
-    digest = "md5"
+    digest = "sha256"
     # Generate key pair
     pkey = crypto.PKey()
-    pkey.generate_key(crypto.TYPE_RSA, 1024)
+    pkey.generate_key(crypto.TYPE_RSA, 3072)
 
     # Generate cert request
     req = crypto.X509Req()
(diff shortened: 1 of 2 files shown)


deluge/core/rpcserver.py
```diff
@@ -530,10 +530,10 @@ def generate_ssl_keys():
     """
     This method generates a new SSL key/cert.
     """
-    digest = "md5"
+    digest = "sha256"
     # Generate key pair
     pkey = crypto.PKey()
-    pkey.generate_key(crypto.TYPE_RSA, 1024)
+    pkey.generate_key(crypto.TYPE_RSA, 3072)
```
cas--: Would a 2048 bit key be sufficient?

caspear: Short answer: Yes.

Longer answer: [Wikipedia](https://en.wikipedia.org/wiki/Key_size#Asymmetric_algorithm_key_lengths) suggests 3072 is sufficient for 128bit symmetric parity. 2048 is claimed to be sufficient until 2030 (i.e. a change would need to be made in 2027 to bump the strength again), but given that 3072 is not significantly larger and offers reasonable protection against non-quantum attackers beyond 2030, is there any reason to use a shorter key length?

cas--:
Thanks, merged with a change to use 2048 key for now: [431357f6230f91e8]

PR #43, opened 2015-09-19
Increased window size of 'Edit trackers' dialog.

Arzie:
I always resize the window to be a bit bigger after opening it, so I thought: why not make a PR for this?

When trying to register for Trac, I got a captcha error, so I can't create a ticket for this.
Also, Travis complains, but it looks like a separate issue (gobject can't be found).


cas--:
Thanks for the PR but doubling the width of the dialog to 800px is not the correct fix here. The best solution is to store the user size request as is done with the Preference dialog. 

The Trac issue and workaround is documented in the forum: http://forum.deluge-torrent.org/viewtopic.php?f=7&t=51737

Arzie:
I agree that storing it as a preference is even better. On the other hand, it's not a bad first step to update the default values to more sensible ones, right?

PR #46, opened 2015-12-04
Fix #1260: Handle redirection better with httpdownloader

bendikro:


cas--:
Merged: 02f6bfd578f3a5a

PR #47, opened 2015-12-04
Defer save state function to separate thread

bendikro:
With large amounts of torrents, saving the state file becomes
a performance bottleneck, mainly due to the required processing
in pickle.dump. When run in the main thread, the server will
hang and be unresponsive for a significant time.

Solve this issue by running the save state job in a separate thread.

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -16,7 +16,7 @@ import os
 import shutil
 import time
 
-from twisted.internet import reactor
+from twisted.internet import defer, reactor, threads
 from twisted.internet.defer import Deferred, DeferredList
 from twisted.internet.task import LoopingCall
 
@@ -104,6 +104,7 @@ class TorrentManager(component.Component):
         # Create the torrents dict { torrent_id: Torrent }
         self.torrents = {}
         self.queued_torrents = set()
+        self.is_saving_state = False
 
         # This is a map of torrent_ids to Deferreds used to track needed resume data.
         # The Deferreds will be completed when resume data has been saved.
@@ -194,6 +195,7 @@ class TorrentManager(component.Component):
         self.save_resume_data_timer.start(190, False)
         self.prev_status_cleanup_loop.start(10)
 
+    @defer.inlineCallbacks
     def stop(self):
         # Stop timers
         if self.save_state_timer.running:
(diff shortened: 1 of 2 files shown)


cas--:
Did you investigate using the reset timer thread option?

deluge/core/torrentmanager.py
```diff
@@ -647,6 +645,26 @@ def create_state(self):
         return state
 
     def save_state(self):
+        """
+        Run the save state task in a separate thread to avoid blocking main thread.
+
+        If a save task is already running, this call is ignored.
+
+        """
+        d = Deferred()
```
cas--: Is this old code? I see d being assigned again below.

bendikro: Yes indeed. Will remove.

deluge/core/torrentmanager.py
```diff
@@ -206,18 +208,14 @@ def stop(self):
             self.prev_status_cleanup_loop.stop()
 
         # Save state on shutdown
-        self.save_state()
-
+        result = yield self.save_state()
```
cas--: result is unused

deluge/core/torrentmanager.py
```diff
@@ -206,18 +208,14 @@ def stop(self):
...
 
-        def remove_temp_file(result):
-            """Remove the temp_file to signify successfully saved state"""
-            if result and os.path.isfile(self.temp_file):
-                os.remove(self.temp_file)
-
-        d = self.save_resume_data(flush_disk_cache=True)
-        d.addCallback(remove_temp_file)
-        return d
+        log.info("Saving resume data...")
```
cas--: log line should be in method

deluge/core/torrentmanager.py
```diff
@@ -206,18 +208,14 @@ def stop(self):
...
+        result = yield self.save_resume_data(flush_disk_cache=True)
+        """Remove the temp_file to signify successfully saved state"""
```
cas--: convert to # comment

deluge/core/torrentmanager.py
bendikro: Indeed.

deluge/core/torrentmanager.py
bendikro: Fixed

cas--:
Merged: d13fca251ed57cf8

PR #54, opened 2015-12-18
[#2786] [GKTUI] Connection Manager Dialog Wont Show, Errno 11001

bendikro:


cas--:
Should verify input IP with is_ip()

PR #55, opened 2015-12-19
[Core] Implement async_add_torrent in torrentmanager

bendikro:


deluge/core/alertmanager.py
```diff
@@ -50,11 +50,10 @@ def __init__(self):
         # handlers is a dictionary of lists {"alert_type": [handler1,h2,..]}
         self.handlers = {}
         self.delayed_calls = []
-        self.wait_on_handler = False
 
     def update(self):
         self.delayed_calls = [dc for dc in self.delayed_calls if dc.active()]
-        self.handle_alerts(wait=self.wait_on_handler)
```
cas--: We probably discussed this before but what was the reason for removing this?

bendikro: We haven't discussed this before. The reason for adding the self.wait_on_handler variable in the first place was reduce the startup load time by avoiding processing all the torrent added events generated while adding the torrents to the session.
When using async_add instead, it's all asynchronous, so there is no longer the same need to postpone handling the events.

cas--: Ah you are reverting a change you made a while ago, that might be what I was thinking of. Do you think we need to retain the `wait` argument for `handle_alerts()`?

bendikro: I guess the wait parameter can be removed as it's not used.

deluge/core/torrentmanager.py
```diff
@@ -384,25 +382,31 @@ def add(self, torrent_info=None, state=None, options=None, save_state=True,
...
-        component.pause("AlertManager")
-
+        d = Deferred()
         try:
-            handle = self.session.add_torrent(add_torrent_params)
-            if not handle.is_valid():
-                raise InvalidTorrentError("Torrent handle is invalid!")
+            self.torrents_loading[torrent_id] = (d, options, state, filename, magnet, resume_data, filedump, save_state)
+            self.session.async_add_torrent(add_torrent_params)
         except (RuntimeError, InvalidTorrentError) as ex:
```
aresch: We probably don't need to catch InvalidTorrentError here anymore as the raise was removed.

deluge/core/torrentmanager.py
```diff
@@ -384,25 +382,31 @@ def add(self, torrent_info=None, state=None, options=None, save_state=True,
...
-        # We need to pause the AlertManager momentarily to prevent alerts
-        # for this torrent being generated before a Torrent object is created.
-        component.pause("AlertManager")
-
+        d = Deferred()
         try:
-            handle = self.session.add_torrent(add_torrent_params)
-            if not handle.is_valid():
-                raise InvalidTorrentError("Torrent handle is invalid!")
+            self.torrents_loading[torrent_id] = (d, options, state, filename, magnet, resume_data, filedump, save_state)
```
aresch: Should we check if torrent_id is already in torrents_loading to prevent a possible race condition adding the same torrent multiple times very quickly?

deluge/core/torrentmanager.py
bendikro: Indeed.

deluge/core/torrentmanager.py
bendikro: Not likely an issue, but added a test for this.

cas--:
Thanks for this! Merged 5d1aff157e4dac

PR #56, opened 2016-01-14
[Tests] Fix for flake8 in tox

bendikro:
Force install flake8 in tox to avoid the system flake8 being used if
available.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -90,15 +90,26 @@ commands =
     python -c "import subprocess, sys; output = subprocess.check_output('isort -q --diff --recursive deluge docs packaging *.py', shell=True); print output; sys.exit(len(output) != 0)"
 
 [testenv:flake8]
+# Force installing flake8 package or else it will use the system installed if it exists.
+# The flake8 executable has the python path used at install time hardcoded, so if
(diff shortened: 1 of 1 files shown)


cas--:
This seems rather hacky, is there not a more elegant way to solve this? :)

bendikro:
What's so hacky? Force installing is the only way to ensure it will not use the system installed package.

tox.ini
```diff
@@ -90,15 +90,26 @@ commands =
     python -c "import subprocess, sys; output = subprocess.check_output('isort -q --diff --recursive deluge docs packaging *.py', shell=True); print output; sys.exit(len(output) != 0)"
 
 [testenv:flake8]
+# Force installing flake8 package or else it will use the system installed if it exists.
+# The flake8 executable has the python path used at install time hardcoded, so if
+# using the system flake8 the wrong libraries are imported.
```
cas--: I think this would be a little more concise:

"Forcing install of flake8 with pip as system installed flake8 uses a hardcoded python path which can import the wrong libraries"

cas--:
Merged 9d13234e23bb6

PR #57, opened 2016-01-14
UI option handling fixes and cleanup

bendikro:
Includes the four commits from https://github.com/jamielennox/deluge/commits/detach-ui as well as a new commit that replaces the deprecated optparse with argparse.


cas--:
There are a lot of console commits, seems to have strayed a bit from UI options...

bendikro:
Yes, I will remove those and merge with the other console changes for a separate PR

bendikro:
All the console specific commits have now been removed.

cas--:
fyi there are a few scrutinizer issues needed fixed

deluge/ui/console/commands/status.py
```diff
@@ -16,60 +16,50 @@
...
         def on_torrents_status(status):
             self.torrents = status
-            if self.status is not None and self.connections is not None and self.torrents is not None:
-                self.print_status()
 
         def on_torrents_status_fail(reason):
-            self.torrents = -1
-            if self.status is not None and self.connections is not None and self.torrents is not None:
-                self.print_status()
+            log.info("Failed to retrieve session status: %s", reason)
```
cas--: log.warn or log.error instead?

deluge/ui/console/main.py
```diff
@@ -10,165 +10,58 @@
...
+        try:
+            # Handle epilog manually to keep the text formatting
+            epilog = self.epilog
+            self.epilog = ""
+            help = super(ConsoleCommandParser, self).format_help()
+            if epilog is not None:
+                help += epilog
+            self.epilog = epilog
+            return help
+        except Exception as ex:
```
cas--: Do we know the exception to be raised?

deluge/ui/console/main.py
```diff
@@ -10,165 +10,58 @@
...
+            log.info("Exception: %s", ex)
+            log.exception(ex)
```
cas--: two too many log lines

deluge/ui/console/main.py
```diff
@@ -10,165 +10,58 @@
...
+        """
+        Differs from ArgumentParser.format_help by adding the raw epilog
+        as formatted in the string. Default bahavior mangles the formatting.
+
+        """
+        try:
+            # Handle epilog manually to keep the text formatting
+            epilog = self.epilog
+            self.epilog = ""
+            help = super(ConsoleCommandParser, self).format_help()
```
cas--: pylint complains about redefining builtin `help`

deluge/commonoptions.py
```diff
@@ -104,8 +105,8 @@ def __init__(self, *args, **kwargs):
                                 help="Set the log level: %s" % ", ".join(deluge.log.levels))
         self.group.add_argument("-q", "--quiet", action="store_true", default=False,
                                 help="Sets the log level to 'none', this is the same as `-L none`")
-        self.group.add_argument("-r", "--rotate-logs", action="store_true", default=False,
-                                help="Rotate logfiles.")
+        self.group.add_argument("-r", "--rotate-logs", action="store", nargs="?", const="50M",
+                                help="Rotate logfiles with optional max file size..")
```
cas--: I think reword this "Enable logfile rotation (optional max file size, default 50MB)"

What do you think about renaming the command to `--logrotate`?

deluge/log.py
```diff
@@ -107,7 +107,7 @@ def findCaller(self):  # NOQA
 }
 
 
-def setup_logger(level="error", filename=None, filemode="w"):
+def setup_logger(level="error", filename=None, filemode="w", max_size="50M"):
```
cas--: Shouldn't really have a human-readable format as arg for a method

deluge/common.py
```diff
@@ -465,6 +465,67 @@ def fdate(seconds, date_only=False, precision_secs=False):
...
+    'customary':     ('B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y'),
+    'customary_ext': ('byte', 'kilo', 'mega', 'giga', 'tera', 'peta', 'exa',
+                      'zetta', 'iotta'),
+    'iec':            ('Bi', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'),
+    'iec_ext':        ('byte', 'kibi', 'mebi', 'gibi', 'tebi', 'pebi', 'exbi',
+                       'zebi', 'yobi'),
+}
+
+
+def human2bytes(s):
```
cas--: I like the idea of this but it fails if you enter an integer or '50MB'

deluge/ui/console/commands/status.py
bendikro: Change to warn

deluge/ui/console/main.py
bendikro: Hmm...not really.  Agree that it should not catch Exception. Tbh, it could just be part of dev testing and may not be needed now.

deluge/ui/console/main.py
bendikro: Will fix.

deluge/log.py
bendikro: What would you sugggest?

cas--: it's probably reasonable to expect an integer that would be megabytes

deluge/ui/console/main.py
bendikro: remove log.exception(ex)

deluge/commonoptions.py
bendikro: Fixed

deluge/ui/console/main.py
bendikro: Removed the try/except.

deluge/log.py
```diff
@@ -128,7 +128,7 @@ def setup_logger(level="error", filename=None, filemode="w"):
     if filename and filemode == "a":
         handler = logging.handlers.RotatingFileHandler(
             filename, filemode,
-            maxBytes=50 * 1024 * 1024,  # 50 Mb
+            maxBytes=common.human2bytes(max_size),
```
cas--: Just to clarify I meant that anything only ui code should be using human2bytes.

deluge/common.py
cas--: I think we remove this for now and create a ticket to revisit inclusion.

deluge/commonoptions.py
```diff
@@ -0,0 +1,143 @@
...
+                                help="Show program's version number and exit")
+        self.group.add_argument("-c", "--config", action="store", metavar='<config>',
+                                help="Set the config directory path")
+        self.group.add_argument("-l", "--logfile", action="store", metavar='<logfile>',
+                                help="Output to designated logfile instead of stdout")
+        self.group.add_argument("-L", "--loglevel", action="store", choices=deluge.log.levels, metavar='<level>',
+                                help="Set the log level: %s" % ", ".join(deluge.log.levels))
+        self.group.add_argument("-q", "--quiet", action="store_true", default=False,
+                                help="Sets the log level to 'none', this is the same as `-L none`")
+        self.group.add_argument("-r", "--logrotate", action="store", nargs="?", const="50M",
```
cas--: Sorry I meant to add remove the `-r` as it's not an option that needs a shorthand. 

Perhaps split the max size to `--logrotatesize`?

bendikro: Isn't an optional argument ok?

cas--: I guess so, perhaps it just needs better worded help hint.

deluge/log.py
bendikro: Having a human readable has the advantage that it is human readable. Having int in megabytes means you cannot set log size smaller than 1MB.

cas--: I know but who would need to rotate logs at less than 1MB? I still maintain that you should not have human readable format as a input for an internal method, it should all be converted upon user input.

bendikro: I actually agree the function should take an integer, but it doesn't really a need a default value?
I'd say the value should be in bytes as a user may need less than MB even if it sounds strange.
Also considering the convert function outputs in bytes anyways, there is nothing to gain by converting that to MBs?

bendikro: Btw, does the convert functions look ok?

aresch: Yes, make it an integer of bytes.

cas--:
With reference to the new `ui/entry.py`, perhaps create a `core/entry.py` for the daemon start code?

deluge/log.py
```diff
@@ -107,15 +107,18 @@ def findCaller(self):  # NOQA
...
     Sets up the basic logger and if `:param:filename` is set, then it will log
     to that file instead of stdout.
 
-    :param level: str, the level to log
-    :param filename: str, the file to log to
+    Args:
+        level (str): The log level to use (Default: "error")
+        filename (str, optional): The log filename. Default is None meaning log to terminal
+        filemode (str): The filemode to use when opening the log file
+        logrotate (int, optional): The size of the logfile when issuing a log rotation
```
cas--: docstring needs `bytes` specified

deluge/log.py
```diff
@@ -125,13 +128,10 @@ def setup_logger(level="error", filename=None, filemode="w"):
...
-    if filename and filemode == "a":
+    if filename and logrotate:
         handler = logging.handlers.RotatingFileHandler(
-            filename, filemode,
-            maxBytes=50 * 1024 * 1024,  # 50 Mb
-            backupCount=3,
-            encoding="utf-8",
-            delay=0
+            filename, filemode, maxBytes=logrotate,
+            backupCount=3, encoding="utf-8", delay=0
```
cas--: `delay=0` is a default value so can be omitted

deluge/log.py
```diff
@@ -125,13 +128,10 @@ def setup_logger(level="error", filename=None, filemode="w"):
...
 
-    if filename and filemode == "a":
+    if filename and logrotate:
         handler = logging.handlers.RotatingFileHandler(
-            filename, filemode,
-            maxBytes=50 * 1024 * 1024,  # 50 Mb
-            backupCount=3,
-            encoding="utf-8",
-            delay=0
+            filename, filemode, maxBytes=logrotate,
```
cas--: Looking at docs filemode must be ignored as it's a named arg `mode` and defaults to `a` anyway so can be omitted.

deluge/log.py
cas--: Do we think a backupcount of 3 is enough?

deluge/commonoptions.py
```diff
@@ -104,37 +105,42 @@ def __init__(self, *args, **kwargs):
                                 help="Set the log level: %s" % ", ".join(deluge.log.levels))
         self.group.add_argument("-q", "--quiet", action="store_true", default=False,
                                 help="Sets the log level to 'none', this is the same as `-L none`")
-        self.group.add_argument("-r", "--rotate-logs", action="store_true", default=False,
-                                help="Rotate logfiles.")
+        self.group.add_argument("--rotate-log", action="store", nargs="?", const="50M",
+                                help="Enable logfile rotation (optional max file size, default: %(const)s)")
```
cas--: I think mention either here or in manpage about the backupcount being a set value

bendikro:
Doesn't the daemon actually fit in the UI category? While it's not a deluge client, it's still an interface to controlling deluge?

deluge/log.py
bendikro: Will fix

deluge/log.py
bendikro: No idea, should ideally be customizable.

deluge/log.py
bendikro: Not following...

cas--: > class logging.handlers.RotatingFileHandler(filename, mode='a', maxBytes=0, backupCount=0, encoding=None, delay=0)

deluge/log.py
cas--: I was thinking that but perhaps 5 would be better for now.

bendikro: Sure

deluge/log.py
bendikro: I see. Just remove the filemode argument to RotatingFileHandler then.

cas--: Yep (along with delay) perhaps I should have just written it that way first of all :D

deluge/log.py
bendikro: Done

deluge/log.py
bendikro: Fixed

deluge/commonoptions.py
bendikro: Fixed

PR #58, opened 2016-02-03
feature request: Show total amount uploaded across all torrents

Lovatoty:
Ideally it would show how much data was uploaded across all torrents, instead of just the torrents in the queue.


aresch:
This pull request doesn't make any sense.  Please don't try to merge 1.3-stable into develop, only merge the commits necessary for your feature.

PR #59, opened 2016-04-03
Misc fixes files_tab.py, torrent.py and core.py

bendikro:
Mostly fixes:
* files_tab.py: Rewritten/cleanup and fixed bugs and added unit tests
* torrent.py:
  - Allow rename folder to empty string to remove
  - Set a default status message to avoid None value
* core.py: rename_files and rename_folder now returns deferreds

### DIFF
--- a/deluge/component.py
+++ b/deluge/component.py
@@ -183,6 +183,9 @@ class Component(object):
         d.addCallback(on_stop)
         return d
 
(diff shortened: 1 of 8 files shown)


cas--:
Thanks merged! https://git.io/vV7TG

PR #61, opened 2016-04-05
Develop sessionproxy diff torrent finished time

bendikro:


bendikro:
This issue will be fixed by [PR6](https://github.com/deluge-torrent/deluge/pull/6) so applying this will not be necessary.

PR #66, opened 2016-04-16
Develop async add torrent fix

bendikro:


deluge/ui/gtkui/torrentview.py
```diff
@@ -663,9 +663,8 @@ def on_columns_changed_event(self, treeview):
         self.save_state()
 
     def on_torrentadded_event(self, torrent_id, from_state):
-        if not from_state:
```
cas--: Does removing this check not cause other potential issues?

bendikro: Difference after async add commit is that events are fired while torrents are added to session at startup. Before the commit, classic mode added the torrent rows on torrentview._on_session_state callback. Now, the _on_session_state returns empty in classic mode so no torrents are added. In thin client mode, all torrents are added from the _on_session_state callback.

cas--:
Merged: 3b84eb635cf3

PR #67, opened 2016-04-16
[GTKUI] Fix #2802: GTKUI classic mode shutdown procedure is broken

bendikro:
As explained in ticket, current shutdown routine in GTKUI is broken: http://dev.deluge-torrent.org/ticket/2802

I think this fixes the issues, but only tested on Ubuntu 15.10

### DIFF
--- a/deluge/ui/gtkui/gtkui.py
+++ b/deluge/ui/gtkui/gtkui.py
@@ -13,12 +13,13 @@ from __future__ import division
 
 import logging
 import os
+import signal
 import sys
 import time
 
 import gobject
 import gtk
-from twisted.internet import gtk2reactor
+from twisted.internet import defer, gtk2reactor
 from twisted.internet.error import ReactorAlreadyInstalledError
 from twisted.internet.task import LoopingCall
 
@@ -251,35 +252,47 @@ class GtkUI(object):
         self.rpc_stats = LoopingCall(self.print_rpc_stats)
 
         # Twisted catches signals to terminate, so have it call a pre_shutdown method.
-        reactor.addSystemEventTrigger("before", "shutdown", self.pre_shutdown)
+        reactor.addSystemEventTrigger("before", "gtkui_close", self.close)
+
         reactor.callWhenRunning(self._on_reactor_start)
+        self.closing = False
+
+        def gtkui_sigint_handler(num, frame):
+            log.debug("SIGINT signal caught - firing event: 'gtkui_close'")
+            reactor.callLater(0, reactor.fireSystemEvent, 'gtkui_close')
+
+        signal.signal(signal.SIGINT, gtkui_sigint_handler)
 
         # Initialize gdk threading
         gtk.gdk.threads_enter()
         reactor.run()
-        self.shutdown()
+        # Reactor is not running. Any async callbacks (Deferreds) can no longer
+        # be processed from this point on.
         gtk.gdk.threads_leave()
 
     def shutdown(self, *args, **kwargs):
-        log.debug("gtkui shutting down..")
-
-        component.stop()
-
-        # Process any pending gtk events since the mainloop has been quit
-        while gtk.events_pending():
-            gtk.main_iteration(0)
-
+        log.debug("GTKUI shutting down...")
         # Shutdown all components
-        component.shutdown()
+        if self.started_in_classic:
+            return component.shutdown()
 
+    @defer.inlineCallbacks
+    def close(self):
+        if self.closing:
+            return
+        self.closing = True
         # Make sure the config is saved.
         self.config.save()
-
-    def pre_shutdown(self, *args, **kwargs):
-        """Modal dialogs can prevent the application exiting so destroy mainwindow"""
         # Ensure columns state is saved
         self.torrentview.save_state()
+        # Shut down components
+        yield self.shutdown()
+
+        # Modal dialogs can prevent the application exiting so destroy mainwindow
+        # Must do this here to avoid hang when closing with SIGINT (CTRL-C)
         self.mainwindow.window.destroy()
+
+        reactor.stop()
 
     def print_rpc_stats(self):
         if not client.connected():
--- a/deluge/core/daemon.py
+++ b/deluge/core/daemon.py
@@ -154,7 +154,8 @@ class Daemon(object):
 
     def _shutdown(self, *args, **kwargs):
         log.info("Deluge daemon shutting down, waiting for components to shutdown...")
-        return component.shutdown()
+        if not self.classic:
+            return component.shutdown()
 
     @export()
     def get_method_list(self):
--- a/deluge/ui/gtkui/mainwindow.py
+++ b/deluge/ui/gtkui/mainwindow.py
@@ -215,7 +215,7 @@ class MainWindow(component.Component):
         def quit_gtkui():
             def stop_gtk_reactor(result=None):
                 try:
-                    reactor.stop()
+                    reactor.callLater(0, reactor.fireSystemEvent, 'gtkui_close')
                 except ReactorNotRunning:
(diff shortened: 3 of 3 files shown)


deluge/ui/gtkui/gtkui.py
```diff
@@ -239,35 +240,51 @@ def nsapp_open_file(osxapp, filename):
...
-    def shutdown(self, *args, **kwargs):
-        log.debug("gtkui shutting down..")
-
-        component.stop()
 
         # Process any pending gtk events since the mainloop has been quit
         while gtk.events_pending():
-            gtk.main_iteration(0)
+            gtk.main_iteration_do(block=False)
+            time.sleep(0.0001)
```
cas--: Is this very short sleep required?

deluge/ui/gtkui/gtkui.py
```diff
@@ -239,35 +240,51 @@ def nsapp_open_file(osxapp, filename):
 
         # Twisted catches signals to terminate, so have it call a pre_shutdown method.
         reactor.addSystemEventTrigger("before", "shutdown", self.pre_shutdown)
```
cas--: Does this get triggered now?

deluge/core/torrentmanager.py
```diff
@@ -693,7 +693,7 @@ def _save_state(self):
                 os.fsync(_file.fileno())
         except (OSError, cPickle.PicklingError) as ex:
             log.error("Unable to save %s: %s", filename, ex)
-            return
+            return False
```
cas--: Are these torrentmanager changes related to this PR?

deluge/ui/gtkui/gtkui.py
```diff
@@ -239,35 +240,51 @@ def nsapp_open_file(osxapp, filename):
...
         """Modal dialogs can prevent the application exiting so destroy mainwindow"""
+        # Make sure the config is saved.
+        self.config.save()
         # Ensure columns state is saved
         self.torrentview.save_state()
         self.mainwindow.window.destroy()
 
+    @defer.inlineCallbacks
+    def close(self):
+        if self.closing:
```
cas--: Probably better with `if not self.closing` and drop the `return`

deluge/core/torrentmanager.py
bendikro: Umm...no
I checked the return value during testing, but it's not needed for the PR. Would be reasonable to return a value or success/failure? Not strictly used for anything at the moment tho.

deluge/ui/gtkui/gtkui.py
bendikro: Not sure, just found this: http://stackoverflow.com/questions/2844902/python-four-steps-setup-with-progressbars
I was going to ask whether this code is even necessary at all...?

deluge/ui/gtkui/gtkui.py
bendikro: I've combined the two functions.

deluge/ui/gtkui/gtkui.py
bendikro: I'd say it's best to avoid indentations of possible?

deluge/ui/gtkui/gtkui.py
cas--: I'm sure I looked into it before but forgotten the original reason. As this is on shutdown the delay should be unneeded.

One of the original sources for this code is probably the pygtk faq: http://faq.pygtk.org/index.py?req=all

deluge/core/torrentmanager.py
cas--: I think without a concrete idea of return value, it's best left without so it's relatively obvious

deluge/ui/gtkui/gtkui.py
bendikro: The comment says `# Process any pending gtk events since the mainloop has been quit`, but what events would this be? When performing a long running task the GUI is unable to update making it seem to freeze. In which case this code would be useful.

Maybe it was added to try to fix issues with the GTKUI not properly closing in standalone mode when closing with CTRL-C?
If that is the case, it's not needed, as it does not fix that problem, and the issue is fixed by this PR anyways.

cas--: The changes were made in 790a10e3, d0fb657adaa, a3636ccd and 848b14605eb3b but there are few details of the whys...

deluge/ui/gtkui/gtkui.py
```diff
@@ -238,36 +239,52 @@ def nsapp_open_file(osxapp, filename):
...
-    def pre_shutdown(self, *args, **kwargs):
-        """Modal dialogs can prevent the application exiting so destroy mainwindow"""
         # Ensure columns state is saved
         self.torrentview.save_state()
+        # Shut down components
+        yield self.shutdown()
+
+        # Modal dialogs can prevent the application exiting so destroy mainwindow
+        # Must do this here to avoid hang when closing with SIGINT (CTRL-C)
         self.mainwindow.window.destroy()
```
cas--: This has to be done before shutdown as the comment describes and is tied to `self.torrentview.save_state()`

deluge/ui/gtkui/gtkui.py
```diff
@@ -238,36 +239,52 @@ def nsapp_open_file(osxapp, filename):
...
+        reactor.addSystemEventTrigger("before", "gtkui_close", self.close)
+
         reactor.callWhenRunning(self._on_reactor_start)
+        self.closing = False
+
+        def gtkui_sigint_handler(num, frame):
+            log.debug("SIGINT signal caught - firing event: 'gtkui_close'")
+            reactor.callLater(0, reactor.fireSystemEvent, 'gtkui_close')
+
+        signal.signal(signal.SIGINT, gtkui_sigint_handler)
```
cas--: To me this seems like the exact same as if twisted catches the sigint before shutdown.

Why not simply `reactor.addSystemEventTrigger("before", "shutdown", self.close)`?

deluge/ui/gtkui/gtkui.py
bendikro: They do explain the purpose of the changes:

- `Fix #535 fix saving column widths`
- `Fix exception when quitting in classic mode`
- `Fix saving torrent state on fresh configs`

So if all these work now it should be ok to remove the code. I haven't tested if the column widths are saved. State should be properly saved, but ideally there would be unit tests for this.

When testing with print statements the save state function did run properly, because the whole component shutdown process is now ensured to be performed before the reactor is shutting down.

I have not tested on Windows, which probably would be a good idea.

deluge/ui/gtkui/gtkui.py
bendikro: It is done before reactor shutdown. Before the last rebase, pre_shutdown was actually run after the close function, but now the same code is run before.

deluge/ui/gtkui/gtkui.py
bendikro: Well, it doesn't work. GTKUI ends up hanging.

bendikro: Umm...after rebase on latest develop SIGINT gives new stacktrace.

cas--: I was actually going to test this myself after lunch but thanks for checking.

deluge/ui/gtkui/gtkui.py
cas--: Considering for a while Windows was excluded from running this code I think it'll be safe to remove :D

bendikro: Good point. Removed the code.

bendikro:
Are there any unresolved issue left?

cas--:
Merged in develop: 47f14845ca

PR #69, opened 2016-04-18
[Core] Handle error when adding torrents to session at startup

bendikro:


cas--:
Merged: 38e0bc125737b7d7c3

PR #70, opened 2016-04-18
[Web] Fix error in WebApi in standalone mode

bendikro:
In GTKUI standalone mode, WebApi.enable would try to connect to
daemon if web.conf had the 'default_daemon' option set, causing
the client calls to break.

### DIFF
--- a/deluge/ui/web/json_api.py
+++ b/deluge/ui/web/json_api.py
@@ -383,18 +383,21 @@ class WebApi(JSONComponent):
             client.set_disconnect_callback(None)
 
     def enable(self):
-        if not client.is_classicmode():
-            client.set_disconnect_callback(self._on_client_disconnect)
         client.register_event_handler("PluginEnabledEvent", self._json.get_remote_methods)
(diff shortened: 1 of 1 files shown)


deluge/ui/web/json_api.py
```diff
@@ -383,18 +383,22 @@ def disable(self):
...
-        if component.get("DelugeWeb").config["default_daemon"]:
-            # Sort out getting the default daemon here
-            default = component.get("DelugeWeb").config["default_daemon"]
-            host = component.get("Web")._get_host(default)
-            if host:
-                return self._connect_daemon(*host[1:])
-            else:
-                return self._connect_daemon()
+
+        if not client.is_classicmode() or True:
```
cas--: left a True hanging around...

bendikro: Ops

bendikro: Which ensured it didn't fix anything at all. Getting tired I think

cas--:
Merged: e370d7d

PR #71, opened 2016-04-18
[Core] Fix adding magnets failing

bendikro:


cas--:
Merged: 5c69b56cd548caf6

PR #72, opened 2016-04-18
[GTKUI] Fix gtk warning on shutdown

bendikro:


deluge/ui/gtkui/listview.py
```diff
@@ -427,9 +427,10 @@ def remove_column(self, header):
         # Store a copy of this columns state in case it's re-added
         state = self.create_column_state(self.columns[header].column)
         self.removed_columns_state.append(state)
-
-        # Start by removing this column from the treeview
-        self.treeview.remove_column(self.columns[header].column)
+        # Only remove column if column is associated with the treeview. This avoids
+        # warning on shutdown when GTKUI is closed before plugins try to remove columns
+        if self.columns[header].column.get_tree_view() is not None:
```
cas--: Do we need the `is not None`, I can't see how it would return `False`?

bendikro: Well, it's definitely the less prone to buggy behavior. Without `is not None` it will depend on the object: http://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none

cas--: Fair enough, I did vaguely recall there something along those lines.

cas--:
Merged: f3bfe177ce3

PR #77, opened 2016-04-22
[Web] With --base option set, serve locally on new base path

bendikro:
When specifying the --base option to work with reverse proxy
the WebUI is no longer accessible locally since it listens
on the server root, but serves on the path specified to work
for the reverse proxy.
Change this to also handle local requests to the base path
such that the WebUI will be available both for the reverse proxy
as well as locally on the interface/port which the twisted sever
listens on.


cas--:
Thanks! merged in develop: 64ac5fdf73f2fecc

PR #78, opened 2016-04-23
[#1959] [WebUI] Allow user selectable GUI language

bendikro:
Implement [#1959](http://dev.deluge-torrent.org/ticket/1959) Allow user selectable GUI language


deluge/core/daemon_entry.py
```diff
@@ -17,6 +17,7 @@
 import deluge.configmanager
 import deluge.error
 from deluge.ui.baseargparser import BaseArgParser
+from deluge.ui.util import lang
```
cas--: I was hoping to not have a dependency on `ui` in core...

deluge/ui/util/lang.py
```diff
@@ -0,0 +1,114 @@
...
+
+import deluge.common
+
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())  # Silence: No handlers could be found for logger "deluge.util.lang"
+
+
+def set_dummy_trans():
+    import __builtin__
+    __builtin__.__dict__["_"] = lambda x: x
```
cas--: should add a log warning that translation is unavailable but strings have been marked for translation

cas--:
A few notes:
- English is not an option in the drop-down.
- It seems to require restarting the web-server to change language, is this unavoidable?
- Need to hit `apply` then `ok` but this is probably a pre-existing issue
- Change `height: 22` to `height: 17` to gain some height in Interface.js

cas--:
With the extra height I would like `language` to be a `xtype: fieldset`

deluge/ui/web/server.py
```diff
@@ -7,6 +7,8 @@
 # See LICENSE for more details.
 #
 
+from __future__ import print_function
```
bendikro: Oh, yeah

deluge/core/daemon_entry.py
bendikro: It only needs this:
```python
def set_dummy_trans():                                                                                                                                                                      
    import __builtin__                                                                                                                                                                      
    __builtin__.__dict__["_"] = lambda x: x                                                                                                                                                 
```

cas--: I can't see any other use for it in `lang.py` so perhaps move it here is a good idea

deluge/ui/util/lang.py
bendikro: Log warning where? in `set_dummy_trans()` or when `_` is called?

bendikro:
Haven't looked at the apply issue, but the Language option should now be working 100%.

deluge/ui/util/lang.py
cas--: When `_` is called

deluge/core/daemon_entry.py
bendikro: It's called from setup_translations as well

cas--:
Should we perhaps either prompt 'refresh browser' or actually refresh the page after clicking apply?

deluge/ui/util/lang.py
bendikro: Done

bendikro:
Probably a good idea

deluge/core/daemon_entry.py
```diff
@@ -55,7 +56,7 @@ def start_daemon(skip_start=False):
         deluge.core.daemon.Daemon: A new daemon object
 
     """
-    deluge.common.setup_translations()
+    lang.set_dummy_trans()
```
cas--: Was meant to set `warn_msg=True`?

bendikro: Ah, yeah

cas--:
You might be right that auto-refresh could be intrusive so popup box perhaps?

bendikro:
Yeah, if that's a simple fix?

cas--:
It should be with: https://docs.sencha.com/extjs/3.4.0/#!/api/Ext.MessageBox

cas--:
I think adding this to onApply should suffice:
```
            if ('language' in changed) {
                Ext.Msg.alert(_('Interface'), _('Please refresh browser...'));
            }
```

deluge/ui/web/js/deluge-all/preferences/InterfacePage.js
```diff
@@ -13,16 +13,22 @@ Ext.namespace('Deluge.preferences');
...
+            width: 320
+        }, config);
+        Deluge.preferences.Interface.superclass.constructor.call(this, config);
+    },
 
     initComponent: function() {
         Deluge.preferences.Interface.superclass.initComponent.call(this);
 
-        var om = this.optionsManager = new Deluge.OptionsManager();
+        var om = this.optionsManager = deluge.preferences.getOptionsManager();
```
bendikro: This causes an error when trying to save when not connected to a daemon. It should be possible to save the Interface page when not connected.

cas--: This change causes an error?

cas--: I shall revert it

bendikro: Reverted this change and added the refresh popup

PR #79, opened 2016-04-24
[Core] Save fastresume file on separate thread

bendikro:
To avoid blocking twisted main thread, defer file saving task to
separate thread with deferToThread.

This could alleviate the issue reported here: http://dev.deluge-torrent.org/ticket/2814

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -106,6 +106,7 @@ class TorrentManager(component.Component):
         self.torrents = {}
         self.queued_torrents = set()
         self.is_saving_state = False
+        self.save_resume_data_file_lock = defer.DeferredLock()
         self.torrents_loading = {}
 
         # This is a map of torrent_ids to Deferreds used to track needed resume data.
@@ -759,7 +760,7 @@ class TorrentManager(component.Component):
             # Use flush_disk_cache as a marker for shutdown so fastresume is
             # saved even if torrents are waiting.
             if not self.waiting_on_resume_data or flush_disk_cache:
-                return self.save_resume_data_file()
+                return self.save_resume_data_file(queue_task=flush_disk_cache)
 
         return DeferredList(deferreds).addBoth(on_all_resume_data_finished)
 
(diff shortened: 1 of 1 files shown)


cas--:
Was this a planned change? Is it possible that it's disk i/o blocking that's causing #2814?

aresch:
I think it's reasonable to assume it's the blocking calls on the main thread causing the issue. If disk i/o blocking in general was the problem, I think that you would be seeing more frequent pauses and much lower throughput in the normal case (non resume saving time).

bendikro:
Well, I have not tested that disk I/O is the problem in [#2814](http://dev.deluge-torrent.org/ticket/2814), so I'm simply assuming that based on the ticket reporter's description only.

If it's correct that it has to do with saving fastresume data, it is either due to libtorrent workload + handling alterts in Deluge, or it's the disk I/O. Libtorrent workload is threaded anyways, so the only thing we can address is the disk I/O.

deluge/core/torrentmanager.py
```diff
@@ -759,6 +760,7 @@ def on_all_resume_data_finished(dummy_result):
             # Use flush_disk_cache as a marker for shutdown so fastresume is
             # saved even if torrents are waiting.
             if not self.waiting_on_resume_data or flush_disk_cache:
+                self.save_resume_data_file()
```
cas--: Should this have been added?

bendikro: Wops, no

bendikro: Fixed

deluge/core/torrentmanager.py
```diff
@@ -795,6 +796,22 @@ def load_resume_data_file(self):
             return resume_data
 
     def save_resume_data_file(self):
+        """
+        Save resume data to file in a separate thread to avoid blocking main thread.
+
+        If a save file task is already running, this call is queued.
```
cas--: What does "call is queued" mean?

cas--: I knew this seemed familiar, could we not use the same implementation as we did for [`save_state`](https://github.com/bendikro/deluge/blob/develop-save-fastresume-separate-thread/deluge/core/torrentmanager.py#L662)

bendikro: The reasoning for stacking the calls is that the resume data is dependent on callbacks from libtorrent. In theory, a `_save_resume_data_file()` call may be running when a shutdown is initiated. If all `save_resume_data alerts` triggered by the call to `save_resume_data()` in `torrentmanager.stop()` are finished before the `_save_resume_data_file()` call (already running when shutting down) has finished, the new call to `_save_resume_data_file()` initiated by the shutdown would be ignored.

With that said, the problem with stacking calls in general did cross my mind, so I think the best solution is to only stack calls when `flush_disk_cache` is true, i.e. on shutdown.

bendikro:
Now only queue the last call on shutdown

deluge/core/torrentmanager.py
```diff
@@ -795,18 +795,29 @@ def load_resume_data_file(self):
...
+            return defer.succeed(None)
+
         def on_lock_aquired():
             d = threads.deferToThread(self._save_resume_data_file)
 
             def on_resume_data_file_saved(arg):
-                if self.save_state_timer.running:
-                    self.save_state_timer.reset()
+                if self.save_resume_data_timer.running:
+                    self.save_resume_data_timer.reset()
```
cas--: Good catch!

PR #80, opened 2016-04-24
[UI] [Daemon] Re-add --fork option

bendikro:


cas--:
The daemon never had this option so shouldn't be included

bendikro:
Removed from daemon

cas--:
Merged: dea43da4d21

PR #82, opened 2016-04-29
[UI] Add tests for ui_entry

bendikro:


deluge/ui/gtkui/gtkui.py
```diff
@@ -28,7 +28,7 @@
     reactor = gtk2reactor.install()
 except ReactorAlreadyInstalledError as ex:
     # Running unit tests so trial already installed a rector
```
cas--: Update comment to reflect change?

deluge/ui/web/server.py
```diff
@@ -567,7 +576,7 @@ def __init__(self, options=None):
         self.web_utils = WebUtils()
 
         self.auth = Auth(self.config)
-        self.standalone = True
+        self.standalone = standalone
```
cas--: Just wonder if this should be 'shared_reactor' instead of 'standalone'. Not keen on the standalone in code :)

PR #83, opened 2016-04-30
Fix Auto-add plugin to accept torrent files with uppercase extension

TannerMoore:
The auto-add feature completely misses the all-caps version of the .torrent extension. Some torrent sites share trackers with the extension as ".TORRENT" instead of ".torrent"

### DIFF
--- a/deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
+++ b/deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
@@ -252,12 +252,13 @@ class Core(CorePluginBase):
                 # Skip directories
                 continue
             else:
-                ext = os.path.splitext(filename)[1]
+                ext = os.path.splitext(filename)[1].lower()
                 if ext == ".torrent":
                     magnet = False
                 elif ext == ".magnet":
                     magnet = True
                 else:
+                    log.debug("File checked for auto-loading is invalid: %s", filename)
                     continue
                 try:
                     filedump = self.load_torrent(filepath, magnet)


deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
```diff
@@ -253,7 +253,7 @@ def update_watchdir(self, watchdir_id):
                 continue
             else:
                 ext = os.path.splitext(filename)[1]
-                if ext == ".torrent":
+                if ext == ".torrent" or ".TORRENT":
```
bendikro: This doesn't work as you may expect.

TannerMoore: How does it work?

bendikro: `ext == ".torrent" or ".TORRENT"` will resolve to True when ext contains ".torrent". If it does not contain ".torrent", it will resolve to ".TORRENT". This means the if test expression will always be true, and `magnet` set to `False`

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
```diff
@@ -253,7 +253,7 @@ def update_watchdir(self, watchdir_id):
                 continue
             else:
                 ext = os.path.splitext(filename)[1]
```
aresch: Why not just make this string lowercase? Simply add .lower() to the end.

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
TannerMoore: Ah good point. I do not code in python, I thought this was shorthand. I am going to add what aresch suggested and just make the extension lowercase.

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
TannerMoore: Done

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
```diff
@@ -252,12 +252,13 @@ def update_watchdir(self, watchdir_id):
                 # Skip directories
                 continue
             else:
-                ext = os.path.splitext(filename)[1]
+                ext = os.path.splitext(filename)[1].lower()
```
cas--: To be picky, I wouldn't modify `ext` here, as it's not strictly the filename extension. It is better to do it temporarily in the test below `if exe.lower() ==`.

cas--: Nevermind previous comment

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
```diff
@@ -252,12 +252,13 @@ def update_watchdir(self, watchdir_id):
...
                 if ext == ".torrent":
                     magnet = False
                 elif ext == ".magnet":
                     magnet = True
                 else:
+                    log.warning("File checked for auto-loading is invalid: %s", filename)
```
cas--: I'm not sure this is needed. If a user uses a directory that contains other files in it then a lot of log messages will be generated.

cas--:
Note to self: this fix should be backported to 1.3

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
bendikro: Yeah, that is a good solution. Your code could also be written as
`if ext  in [".torrent", ".TORRENT"]`

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
TannerMoore: The reason I want to add some sore of logging here is because I spent at least an hour trying to figure out why the auto-add setting wasn't paying any attention to my specified folder. in reality, it was paying close attention to that folder, but silently failing. I see your concern, though.

An alternative that I can think of right now would be to add a debug message to the extension check to just say something like "evaluating file for auto-add." 

I still stand behind adding some form of increased visibility.

cas--: Fair enough, log.debug would be fine

TannerMoore: Debug on the current line in this comment chain (already done above), or debug on the initial check of files in the watch folder?

cas--:
Could you please rebase with latest `develop` branch and fixup into a single commit. The commit message convention is here: http://dev.deluge-torrent.org/wiki/Contributing/CodingStyle#GitCommitMessages

TannerMoore:
I looked at that link you sent me, but it doesn't talk about commit messages at all.

cas--:
That's because I moved the text yesterday to [GitTips](http://dev.deluge-torrent.org/wiki/GitTips#CommitMessages)

TannerMoore:
Alright. Commit rebased and commit message cleaned up. Let me know if everything looks good.

cas--:
I'm afraid the commit title is not correct, it shouldn't start with a dash `-`. It also doesn't follow the same convention of existing [develop commits](https://github.com/deluge-torrent/deluge/commits/develop). I did leave a hint when I renamed the title of this PR as to the wording so something like this would be more appropriate:

`[AutoAdd] Fix watch directory accepting torrent files with uppercase extension`

As mentioned in GitTips, you don't need to mention the log line addition as it's self-evident.

TannerMoore:
Reworded again.

cas--:
Merged

PR #84, opened 2016-05-01
[#2813][Partial Revert][GTKUI] Remove old twisted and gnome code

doadin:
Fixes #2813 where the connection dialog does not update/show the proper
daemon state

### DIFF
--- a/deluge/ui/gtkui/gtkui.py
+++ b/deluge/ui/gtkui/gtkui.py
@@ -453,6 +453,8 @@ class GtkUI(object):
                     break
 
         if self.config["show_connection_manager_on_start"]:
+            # XXX: We need to call a simulate() here, but this could be a bug in twisted
+            reactor._simulate()
             self.connectionmanager.show()
 
     def __on_disconnect(self):


deluge/ui/gtkui/gtkui.py
```diff
@@ -453,6 +453,12 @@ def do_connect(try_counter, host, port, user, passwd):
                     break
 
         if self.config["show_connection_manager_on_start"]:
+            # XXX: We need to call a simulate() here, but this could be a bug in twisted
+            try:
+                reactor._simulate()
+            except AttributeError:
```
cas--: Since we are not supporting twisted < 12 the `try...except` can be removed

cas--:
The problem I have with reverting that change is we don't have a "why" it is needed and I cannot replicate on Ubuntu. Also is was renamed to a `private` method suggests that we should not be using it.

doadin:
Well im not 100% but it seems like connect is timing out when called the first time why exactly i don't know, and why it doesn't connect after deluge is like fully loaded since it shouldn't have a reason to time out i don't know.

And that brings me to the next part yes it is happening on windows(you dont test on windows?) so maybe its a windows only bug? So unless we can figure out a better fix or get twisted to fix it i believe its out of our control and at least for now have to revert it.

doadin:
also did you read the log i added to the ticket?

doadin:
Just to like super verify this i just remade an entire system os and all and still got this issue. Feel free to do the same and test windows if you like.

deluge/ui/gtkui/gtkui.py
doadin: related side note might wana throw in an update to depends file at some point

cas--:
None of the current devs have Windows machines, I certainly would only test it closer to release in virtualbox.

Thanks for the PR but I need to write a slightly modified and descriptive commit: 21789e0692d36

doadin:
All good, i kinda figured since there was this and that one other commit tat broke windows, dont remember which but any who. Im thinking of adding a jenkins job to my build pc to keep a closer eye on commits so i dont have to rever and try 10 different commits or w.e. when a issue comes up.

PR #85, opened 2016-05-01
[common] Cleanup and improve windows support for is_ip

doadin:


cas--:
If we are going to rely an external module then `ipaddress` backport (`python-ipaddress`) from Py3 would be a better option.

cas--:
I perhaps should have mentioned that with `ipaddress` we can simply do away with `common.is_ip` and use it directly in code. I don't think this change is that urgent so can be done at the same time as the move to python 3 support. A ticket with milestone 3.0 will suffice for now.

PR #87, opened 2016-05-06
Fix #2829 : Cannot disable Scheduler plugin

5haffl:
Please refer to http://dev.deluge-torrent.org/ticket/2829 for detailed information about this bug.


cas--:
This change certainly should not be applied to develop branch. It has been already fixed in 1.3-stable.

5haffl:
I'm very sorry for this mistake, my biggest concern was running into duplicate bug reports, in fact I did carefully search for similar issues in the ticketing sysem, but I found nothing.
The strange thing is that I still experience this bug with version 1.3.12-0 (latest), on Debian Wheezy/armhf. I installed the package provided by the repository suggested in Deluge's wiki: http://ppa.launchpad.net/deluge-team/ppa/ubuntu
I suspect the fix might not have been applied to (at least) this package in the repository.

PR #88, opened 2016-05-07
Enable more useful pylint messages

cas--:
One issue needs fixing:

```deluge/plugins/init.py:22: [E1102(not-callable), PluginInitBase.__init__] self._plugin_cls is not callable```


cas--:
- [ ] Fix plugins pylint issues
- [ ] Add plugins path the scrutinizer web config

PR #90, opened 2016-05-10
Develop pass ui client args

bendikro:


cas--:
Just revert/delete my commit removing of unicode_argv until we can test it on Windows and be sure it's working without it. Can be a python 3 milestone...

I think a few extra details in the commit message just to quicky explain the changes would be handy. Also detailing the original issue i.e. `~/delugeDev/deluge -u web -a "-h"` not displaying help.

deluge/ui/ui_entry.py
```diff
@@ -28,69 +29,136 @@
...
+    del config
 
-    selected_ui = options.ui if options.ui else config["default_ui"]
+    # We have parsed and got the config dir needed to get the default UI
+    # Now create a parser for choosing the UI
+    parser = BaseArgParser(common_help=True)
+    group = parser.add_argument_group(_("UI Options"))
+    # We need this option for parsing the succeed, but result is not used
+    group.add_argument("-s", "--set-default-ui", dest="default_ui", choices=ui_entrypoints.keys(),
+                       help=_("Sets the default UI to be run when no UI is specified"), action="store")
```
cas--: action='store' is default so can be omitted

deluge/ui/ui_entry.py
```diff
@@ -28,69 +29,136 @@
...
 
-    config.save()
-    del config
+    # Create a subparser for each registred UI
+    subparsers = parser.add_subparsers(help="The UI to launch", dest='selected_ui')
+    for ui, obj in ui_entrypoints.iteritems():
+        parser_ui = subparsers.add_parser(ui, common_help=False, help=getattr(obj, "cmd_description", ""))
+        parser_ui.add_argument("args", action="store", nargs=argparse.REMAINDER)
```
cas--: action='store' is default so can be omitted

bendikro: Fixed

deluge/ui/ui_entry.py
bendikro: Fixed

bendikro:
Current output of `deluge -h`
```
usage: deluge [-h] [-V] [-c <config>] [-l <logfile>] [-L <level>]
              [--logrotate [<max-size>]] [-q] [--profile [<profile-file>]]
              [-s {console,gtk,web}]
              {console,gtk,web} [UI arguments] ...

Common Options:
  -h, --help                     Print this help message
  -V, --version                  Print version information
  -c, --config <config>          Set the config directory path
  -l, --logfile <logfile>        Output to specified logfile instead of stdout
  -L, --loglevel <level>         Set the log level (none, error, warning, info, debug)
  --logrotate [<max-size>]       Enable logfile rotation, with optional maximum logfile
                                   size, default: 2M (Logfile rotation count is 5)
  -q, --quiet                    Quieten logging output (Same as `--loglevel none`)
  --profile [<profile-file>]     Profile deluge with cProfile. Outputs to stdout unless a
                                   filename is specified

UI Options:
  -s, --set-default-ui {console,gtk,web}
                                 Sets the default UI to be run when no UI is specified

  {console,gtk,web} [UI arguments]
                                 The UI to launch (default: *)
      console                    A console or command-line interface
      gtk                        A GTK-based graphical user interface
    * web                        A web-based interface (http://localhost:8112)
```

cas--:
Ok so I had added a few text tweaks and rebased them and was about to merge when I realised that the ui_entry tests should not be passing...

bendikro:
I think I got a working solution now that correctly handles the following key cases:

1. console command with no extra argument will connect in interactive mode: `./deluge console`
1. console help: `./deluge console -h`
1. console sub command: `./deluge console info`
1. console sub command help: `./deluge console info -h`

I've added tests to verify this to work, but more tests cases could probably be added.

`--ui` will now give `unrecognized arguments: --ui`. Is this not desirable?

deluge/ui/console/commander.py
```diff
@@ -86,14 +93,24 @@ def print_help(f=None):
...
-                import traceback
-                self.write("%s" % traceback.format_exc())
-                return defer.succeed(True)
-            else:
-                return ret
+        if getattr(parser, "_exit", False):
+            return
+        return options
+
+    def exec_command(self, options, *args):
```
cas--: At first glance I'm not sure what the differences are between `do_command`, `parse_command` and `exec_command` are. I think updates to the docstrings might help.

deluge/ui/gtkui/__init__.py
```diff
@@ -20,15 +20,16 @@ class Gtk(UI):
     cmd_description = """GTK-based graphical user interface"""
 
     def __init__(self, *args, **kwargs):
+        self.ui_args = kwargs.pop("ui_args", None)
```
cas--: I don't see further usage of `self.ui_args`?

deluge/ui/web/web.py
```diff
@@ -24,6 +24,7 @@ class Web(UI):
     cmd_description = """Web-based user interface (http://localhost:8112)"""
 
     def __init__(self, *args, **kwargs):
+        self.ui_args = kwargs.pop("ui_args", None)
```
cas--: I don't see further usage of `self.ui_args` here either?

deluge/ui/baseargparser.py
```diff
@@ -117,9 +182,24 @@ def __init__(self, *args, **kwargs):
...
+    def parse_args(self, args=None):
+        """Parse UI arguments and handle common and process group options.
+
+        Unknown arguments return an error and resulting usage text.
+        """
+        options = super(BaseArgParser, self).parse_args(args=args)
+        return self._parse_args(options)
+
+    def parse_known_ui_args(self, args=None):
+        """Parse UI arguments and handle common and process group options.
```
cas--: Needs a better docstring to differentiate from above one

deluge/ui/baseargparser.py
```diff
@@ -20,6 +20,69 @@
...
+def set_default_subparser(self, name, args=None, before_subcmd=None):
+    """Sets the default argparse subparser.
+
+    Args:
+        name (str): The name of the default subparser.
+        args (list, optional): The argument list passed to parse_args().
+        before_subcmd (list): The arguments to test for that come before the subcommand.
+
+    Returns:
+        bool: True if help option is not found after a subcommand, else False.
```
cas--: Is this text about the help option correct. It doesn't seem to have any relevance here?

deluge/ui/baseargparser.py
bendikro: The difference is explained tho...

cas--: Perhaps `Parse UI arguments and handle common and process group options without error.` will do

bendikro: Fixed

deluge/ui/ui_entry.py
```diff
@@ -32,65 +34,92 @@ def start_ui():
...
 
-    # reconstruct arguments to hand off to child client
-    client_args = []
-    if options.args:
-        import shlex
-        client_args.extend(shlex.split(options.args))
+    # We have parsed and got the config dir needed to get the default UI
+    # Now create a parser for choosing the UI
+    parser = BaseArgParser(common_help=True)
+    group = parser.add_argument_group(_("UI Options"))
```
cas--: I think you should replace these 3 lines with:

```
# Reuse UI Options group to actually display the group text to the user. The result from these args is not used.
parser._action_groups.append(group)
```

I know it's using a private method but the duplication is hurting my eyes :P

deluge/ui/ui_entry.py
```diff
@@ -32,65 +34,92 @@ def start_ui():
...
-    cmd_help = [_("The UI that you wish to launch. The UI choices are:")]
-    cmd_help.extend(["%s -- %s" % (k, getattr(v, "cmd_description", "")) for k, v in ui_entrypoints.iteritems()])
-
-    group.add_argument("-u", "--ui", action="store",
-                       choices=ui_entrypoints.keys(), help="\n  * ".join(cmd_help))
-    group.add_argument("-a", "--args", action="store",
-                       help=_('Arguments to pass to the UI. Multiple args must be quoted, e.g. -a "--option args"'))
-    group.add_argument("-s", "--set-default-ui", dest="default_ui", choices=ui_entrypoints.keys(),
-                       help=_("Sets the default UI to be run when no UI is specified"), action="store")
+    # Setup the initial argument parser for the UI selection
```
cas--: This comment should perhaps mention that the arg results are processed here but the text will not be shown and reusing the group later in code is required for displaying to the user.

deluge/ui/console/commander.py
bendikro: Updated docs

deluge/ui/ui_entry.py
bendikro: Updated comments

deluge/ui/gtkui/__init__.py
bendikro: No, can just do with the pop

deluge/ui/web/web.py
bendikro: Same. Only used in console

deluge/ui/ui_entry.py
```diff
@@ -32,65 +34,93 @@ def start_ui():
...
+                                          help=getattr(ui_entrypoints[ui], "cmd_description", ""))
+        parser_ui.add_argument("ui_args", nargs=argparse.REMAINDER)
+        # Change the command name in help to indicate default UI
+        subactions = subparsers._get_subactions()
+        prefix = "*" if ui == default_ui else " "
+        subactions[-1].dest = "%s %s" % (prefix, ui)
+
+    # Handle help which was ignored earlier. If -h/--help option is before the
+    # UI argument, we can now show the help message with a list of possible UIs
+    arg_before_subcmd = parser.set_default_subparser(default_ui, abort_opts=dubious_args)
```
cas--: This code seems to have no effect when removed?

bendikro: You are right. The issue this code handled was actually fixed by the changes here: https://github.com/deluge-torrent/deluge/pull/90/files#diff-1250fcb4c9343ad6037f48af416b41acL83

PR #92, opened 2016-05-11
Develop autodoc fix

bendikro:
Removed sphinx modules doc files and instead use apidoc to autogenerate the rst files.


cas--:
Could the autogen apidoc be added to setup.py docs command

tox.ini
```diff
@@ -184,15 +184,18 @@ deps =
...
 [testenv:docs]
 sitepackages = {[docsbase]sitepackages}
 changedir = {[docsbase]changedir}
 install_command = {[docsbase]install_command}
 deps = {[docsbase]deps}
 whitelist_externals = {[docsbase]whitelist_externals}
-commands = sphinx-build -v -E -T -b html -d build/doctrees source build/html
+commands =
+    sphinx-apidoc --force -o source/modules/ ../deluge ../deluge/plugins
+    sphinx-build -v -E -T -b html -d build/doctrees source build/html
```
cas--: Should we replace this with the `setup.py build_docs` command?

bendikro: I'd prefer it like this as it's much easier to see the exact commands that are executed as well as copy pasting for testing manually.

cas--: I guessed that might have been the reason but it means there is a potential disparity between `setup.py` and `tox`. How about verbose command to print out the command in `setup.py`?

bendikro: Sure, I guess that is possible, but I'm not sure how to access the build command executed by `sphinx.setup_command.BuildDoc`

cas--: I can't actually get that build command to work locally...

My other reason for asking is so that tox is testing the setup.py BuildDocs command. We can get the config from `BuildDoc.dump_options(self)` but it's not quite the same...

bendikro: Could add another tox target

cas--: That seems a waste...

How about print out those two lines and call `BuildDocs`?

bendikro: That works I guess

PR #93, opened 2016-05-11
Add external ip from lt to UI statusbars

cas--:
- [x] Add to console statusbar
- [x] Fix isort layout

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -92,6 +92,10 @@ class Core(component.Component):
         # New release check information
         self.new_release = None
 
+        # External IP Address from libtorrent
+        self.external_ip = None
+        self.eventmanager.register_event_handler("ExternalIPEvent", self._on_external_ip_event)
+
         # Get the core config
         self.config = ConfigManager("core.conf")
         self.config.save()
@@ -919,6 +923,16 @@ class Core(component.Component):
             return deluge.common.free_space(path)
         except InvalidPathError:
             return -1
+
+    def _on_external_ip_event(self, external_ip):
+        self.external_ip = external_ip
+
+    @export
+    def get_external_ip(self):
+        """
+        Returns the external ip address recieved from libtorrent.
+        """
+        return self.external_ip
 
     @export
     def get_libtorrent_version(self):
--- a/deluge/ui/gtkui/statusbar.py
+++ b/deluge/ui/gtkui/statusbar.py
@@ -179,6 +179,9 @@ class StatusBar(component.Component):
             text=_("No Incoming Connections!"),
             callback=self._on_health_icon_clicked, pack_start=True)
 
+        self.external_ip_item = self.add_item(
+            tooltip=_("External IP Address"), pack_start=True)
+
         self.health = False
 
         def update_config_values(configs):
@@ -201,6 +204,7 @@ class StatusBar(component.Component):
             self.remove_item(self.health_item)
             self.remove_item(self.traffic_item)
             self.remove_item(self.diskspace_item)
+            self.remove_item(self.external_ip_item)
         except Exception as ex:
             log.debug("Unable to remove StatusBar item: %s", ex)
         self.show_not_connected()
@@ -270,6 +274,7 @@ class StatusBar(component.Component):
 
         client.core.get_session_status(keys).addCallback(self._on_get_session_status)
         client.core.get_free_space().addCallback(self._on_get_free_space)
+        client.core.get_external_ip().addCallback(self._on_get_external_ip)
 
     def on_configvaluechanged_event(self, key, value):
         """
@@ -325,6 +330,10 @@ class StatusBar(component.Component):
     def _on_max_upload_speed(self, max_upload_speed):
         self.max_upload_speed = max_upload_speed
         self.update_upload_label()
+
+    def _on_get_external_ip(self, external_ip):
+        ip = external_ip if external_ip else _("n/a")
+        self.external_ip_item.set_markup(_("<b>IP</b> %s") % ip)
 
     def update_connections_label(self):
         # Set the max connections label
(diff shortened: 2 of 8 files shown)


deluge/ui/gtkui/statusbar.py
```diff
@@ -326,6 +331,12 @@ def _on_max_upload_speed(self, max_upload_speed):
         self.max_upload_speed = max_upload_speed
         self.update_upload_label()
 
+    def _on_get_external_ip(self, external_ip):
+        if external_ip:
```
aresch: Maybe this instead to reduce repetition?
```
e = external_ip if external_ip else "n/a"
self.external_ip_item.set_markup(_("<b>IP</b> %s") % e)
```

Also, should we be concatenating the strings to ensure translation can happen?
```
_("<b>IP</b> " + e)
```

deluge/ui/gtkui/statusbar.py
```diff
@@ -326,6 +331,12 @@ def _on_max_upload_speed(self, max_upload_speed):
...
+            self.external_ip_item.set_markup(_("<b>IP</b> %s") % external_ip)
+        else:
+            self.external_ip_item.set_text(_("<b>IP</b> n/a"))
```
aresch: This should be set_markup() if you don't change the function to remove this line.

deluge/core/core.py
```diff
@@ -920,6 +924,15 @@ def get_free_space(self, path=None):
         except InvalidPathError:
             return -1
 
+    def _on_external_ip_event(self, external_ip):
+        self.external_ip = external_ip
+
+    @export
+    def get_external_ip(self):
+        """
```
aresch: Need a docstring here.

deluge/ui/web/json_api.py
```diff
@@ -551,6 +554,9 @@ def on_complete(result):
         d4 = client.core.get_free_space(self.core_config.get("download_location"))
         d4.addCallback(got_free_space)
 
+        d5 = client.core.get_external_ip()
+        d5.addCallback(got_external_ip)
+
         dl = DeferredList([d1, d2, d3, d4], consumeErrors=True)
```
bendikro: Shouldn't d5 be added to the DeferredList?

deluge/ui/web/js/deluge-all/Statusbar.js
```diff
@@ -83,27 +83,27 @@ Deluge.Statusbar = Ext.extend(Ext.ux.StatusBar, {
             menu: new Deluge.StatusbarMenu({
                 items: [{
                     value: '5',
-                    text: '5 KiB/s',
+                    text: _('5 KiB/s'),
```
bendikro: Would translating `5 KiB/s` ever be needed?

cas--: This was to keep consistency, the unit is translated.

deluge/ui/gtkui/statusbar.py
cas--: Good point, at the time I thought it could be neater...

I have been making the mistake of thinking I should avoid `%s` in the translation string but found since then that [concat is not correct](http://read-the-docs.readthedocs.io/en/latest/i18n.html#interpolation). The string should use substitution (or format) so that translators can change the position of `%s` if the word order is different. With this new knowledge I may have to go back through the code and make the correct alterations!

deluge/ui/web/json_api.py
cas--: oh hah, it worked thought :D

deluge/ui/web/js/deluge-all/Statusbar.js
```diff
@@ -289,5 +294,11 @@ Deluge.Statusbar = Ext.extend(Ext.ux.StatusBar, {
 
         this.items.get('statusbar-dht').setText(stats.dht_nodes);
         this.items.get('statusbar-freespace').setText(stats.free_space >= 0 ? fsize(stats.free_space): _('Error'));
+        if (stats.external_ip) {
+            this.items.get('statusbar-externalip').setText(
+                String.format(_('<b>IP</b> {0}'), stats.external_ip));
```
cas--: Should make the same changes as gtkui

deluge/ui/web/js/deluge-all/AddConnectionWindow.js
```diff
@@ -83,7 +83,7 @@ Deluge.AddConnectionWindow = Ext.extend(Ext.Window, {
                 if (!result[0]) {
                     Ext.MessageBox.show({
                         title: _('Error'),
-                        msg: 'Unable to add host: ' + result[1],
+                        msg: _('Unable to add host: ') + result[1],
```
cas--: Use substitution

cas--:
@aresch Perhaps I should simply update core attribute in tm.on_alert_external_ip: `component.get("Core").external_ip = external_ip` rather than register event in core...?

aresch:
I think that you're currently doing it the right way.  Directly updating
the core attribute would feel like it breaks encapsulation.

doadin:
would it not be wise to maybe add a prefference for showing this? What if someone has other plugins that use status bar and now they have to make the window bigger to see them or default ones they do want get pushed away. Just a thought.

aresch:
At one point I thought that including a Preferences page for the Statusbar
would be a good idea.  The user could toggle on/off the various
indicators/items.  Maybe that is what we need?

cas--:
There is certainly an issue with the statusbar items setting the minimum width of the main window so a toggle for statusbar items or something similar would be nice. There should be an existing ticket that covers this issue.

cas--:
This is the existing ticket: http://dev.deluge-torrent.org/ticket/1326

doadin:
ah ok. yea i think like a prefferences page for status bar to add/remove items would be nice. maybe something simple like if checks to the gtk config around the statsbar create item or something more complex like status bar items registering like plugins. idk. glade to know its being thought about, though that ticket has been open for quite some time now lol.

PR #94, opened 2016-05-15
Fixes for issue found with webui and plugins

cas--:
- [x] How to handle disabling WebUI plugin from within browser now that it actually appears enabled in the plugins list.

### DIFF
--- a/deluge/tests/test_web_api.py
+++ b/deluge/tests/test_web_api.py
@@ -139,13 +139,13 @@ class WebAPITestCase(BaseTestCase, DaemonBase):
         self.assertEquals(status, tuple(status))
 
     def test_get_host(self):
-        self.assertEquals(self.deluge_web.web_api._get_host("invalid_id"), None)
+        self.assertFalse(self.deluge_web.web_api._get_host("invalid_id"))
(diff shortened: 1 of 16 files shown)


deluge/tests/test_web_api.py
```diff
@@ -139,13 +139,13 @@ def get_host_status(self):
         self.assertEquals(status, tuple(status))
 
     def test_get_host(self):
-        self.assertEquals(self.deluge_web.web_api._get_host("invalid_id"), None)
+        self.assertFalse(self.deluge_web.web_api._get_host("invalid_id"))
```
bendikro: Is using `self.assertFalse` really a good idea when the only correct value is a list?

bendikro:
Looks good

deluge/tests/test_web_api.py
cas--: Well I did originally have it equal to `[]` but I thought since we test for `False` in code whenever there is an empty list it would be a more equivalent test...

cas--:
I have added a commit that simply disables changing the status of the WebUi plugin in the browser. It is the simplest solution for now as it has rather unpredictable results, i.e. web server continues to run even when disabled...

PR #97, opened 2016-05-17
#2807 WebUI needs minified before setup install 

cas--:
WiP for [#2807](http://dev.deluge-torrent.org/ticket/2807)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,46 @@ class PyTest(_test):
         sys.exit(errcode)
 
 
+class BuildWebUI(cmd.Command):
+    description = 'Minify WebUI files'
+
+    user_options = [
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -60,6 +60,47 @@ def run_tests(self):
...
+        self.develop = False
+
+    def finalize_options(self):
+        self.set_undefined_options('build', ('build_lib', 'build_lib'))
+
+    def run(self):
+        js_dir = os.path.join(os.path.dirname(__file__), 'deluge/ui/web/js/')
+
+        if self.develop:
+            basedir = js_dir
```
bendikro: basedir isn't used. Should it be?

cas--: Not sure what this is for, need to rattle my brain...

PR #99, opened 2016-05-21
Develop misc fixes tests component core

bendikro:
Various minor fixes.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -346,10 +346,13 @@ class Core(component.Component):
 
         Args:
             torrent_ids (list): The torrent IDs to remove.
-            remove_data (bool, optional): If True, also remove the downloaded data.
+            remove_data (bool): If True, also remove the downloaded data.
 
         Returns:
-            list: a list containing all the errors, empty list if no errors occured
+            list of tuples: A list tuples containing all the errors, empty list if no errors occured
+               tuple format:
+                  0: str: torrent-id
+                  1: str: Error message
 
         """
         log.info("Removing %d torrents from core.", len(torrent_ids))
@@ -360,9 +363,11 @@ class Core(component.Component):
                 try:
                     self.torrentmanager.remove(torrent_id, remove_data=remove_data, save_state=False)
                 except InvalidTorrentError as ex:
-                    errors.append((torrent_id, ex))
+                    errors.append((torrent_id, str(ex)))
             # Save the session state
             self.torrentmanager.save_state()
+            if errors:
+                log.warn("Failed to remove %d of %d torrents.", len(errors), len(torrent_ids))
             return errors
         return task.deferLater(reactor, 0, do_remove_torrents)
 
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -349,10 +349,10 @@ class Core(component.Component):
             remove_data (bool): If True, also remove the downloaded data.
 
         Returns:
-            list of tuples: A list tuples containing all the errors, empty list if no errors occured
-               tuple format:
-                  0: str: torrent-id
-                  1: str: Error message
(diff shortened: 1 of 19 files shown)


cas--:
Just a few points on docstrings: 
* Every line should be finished with a full-stop as they are sentences and makes it clear there is no continuation of the text.
* The first line summary of a method should be on the same line as the triple-quote e.g. `"""This is a module that...`
* I was discussing this before that unfortunately sphinx complains about indented lines in the `Returns` sections. I don't know if we should comply with it though...
So the following is invalid according to sphinx:
```
Returns:
    tuple: some long text...
        continuing text....
```
and sphinx wants it to be:
 ```
Returns:
    tuple: some long text...
    continuing text....
```
To me it seems like a bug that needs reporting but may never be fixed ;)

File: deluge/tests/common.py [hunk 1]
```diff
@@ -150,6 +168,26 @@ def errReceived(self, data):  # NOQA
 
 def start_core(listen_port=58846, logfile=None, timeout=10, timeout_msg=None,
                custom_script="", print_stderr=True, extra_callbacks=None):
+    """
+    Start the deluge core as a daemon
+
+    Kwargs:
```
cas--: Should be Args

deluge/core/core.py
```diff
@@ -359,8 +359,8 @@ def do_remove_torrents():
             for torrent_id in torrent_ids:
                 try:
                     self.torrentmanager.remove(torrent_id, remove_data=remove_data, save_state=False)
-                except InvalidTorrentError as ex:
-                    errors.append((torrent_id, ex))
+                except InvalidTorrentError:
+                    errors.append(torrent_id)
```
cas--: Could you elaborate in the commit message why this fix was needed

deluge/tests/common.py
bendikro: I was looking here which uses Kwargs as well as Args: https://pythonhosted.org/an_example_pypi_project/sphinx.html#full-code-example

But maybe that's not necessary.

cas--: Ah... I don't really see the need and the convention so far is without

cas--:
So based on that sphinx example you shared, I realised that for the `Returns` section issue we can use `::` to enable the use of indents.

deluge/core/core.py
bendikro: The problem is simply that exceptions can't be serialized by rencode. The tests pass because it's run locally without serializing. I would prefer a solution where the reason could be returned as part of the return value. We could return the exceptions getMessage value?

bendikro: Updated with a fixup that reverts the change and uses a string value instead of the exception. What do you think?

deluge/core/core.py
```diff
@@ -359,10 +362,12 @@ def do_remove_torrents():
...
                 try:
                     self.torrentmanager.remove(torrent_id, remove_data=remove_data, save_state=False)
-                except InvalidTorrentError:
-                    errors.append(torrent_id)
+                except InvalidTorrentError as ex:
+                    errors.append((torrent_id, str(ex)))
             # Save the session state
             self.torrentmanager.save_state()
+            if errors:
+                log.error("Failed to remove %d of %d torrents.", len(errors), len(torrent_ids))
```
cas--: Do we care enough about removing torrents that don't exist to log it? :)

deluge/core/core.py
cas--: That seems like a better fix

deluge/core/core.py
bendikro: Well, the main reason is that it shouldn't really be happening, so at least it informs us that something isn't right

cas--: Ok perhaps warning or debug then?

bendikro: Sure

PR #101, opened 2016-05-22
[Console] Fix to console argument parsing

bendikro:
When starting console with './deluge-console', providing
loggin level '-L info' would fail to parse as it identified
'info' as a subcommand.

### DIFF
--- a/deluge/ui/baseargparser.py
+++ b/deluge/ui/baseargparser.py
@@ -19,25 +19,27 @@ from deluge import common
 from deluge.configmanager import get_config_dir, set_config_dir
 
 
-def find_subcommand(self, args=None):
+def find_subcommand(self, args=None, sys_argv=True):
     """Find if a subcommand has been supplied.
 
     Args:
-        args (list, optional): The argument list handed to parse_args().
+        args (list, optional): The argument list to search through.
+        sys_argv (bool): Use sys.argv[1:] if args is None.
 
     Returns:
         int: Index of the subcommand or '-1' if none found.
 
     """
     subcommand_found = -1
-    test_args = args if args is not None else sys.argv[1:]
+    if args is None:
+        args = sys.argv[1:] if sys_argv is None else []
 
     for x in self._subparsers._actions:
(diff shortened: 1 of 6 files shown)


deluge/ui/baseargparser.py
```diff
@@ -19,25 +19,26 @@
 from deluge.configmanager import get_config_dir, set_config_dir
 
 
-def find_subcommand(self, args=None):
+def find_subcommand(self, args=None, sys_argv=True):
     """Find if a subcommand has been supplied.
 
     Args:
-        args (list, optional): The argument list handed to parse_args().
+        args (list): The argument list to search through.
```
cas--: `args` is still optional

bendikro: Fixed

cas--:
Made some minor changes and updated docscrings

deluge/ui/baseargparser.py
```diff
@@ -184,20 +186,47 @@ def __init__(self, *args, **kwargs):
...
+        if withhold:
+            args = [a for a in args if a not in withhold]
         options, remaining = super(BaseArgParser, self).parse_known_args(args=args)
         options.remaining = remaining
-        return self._parse_args(options)
+        # Hanlde common and process group options
+        return self._handle_ui_options(options)
+
+    def _handle_ui_options(self, options):
+        """Parse UI arguments and handle common and process group options without error.
```
cas--: Should this be `Handle UI args...`, I don't think a third method with almost identical summary is a good idea :)

bendikro:
Close to perfection :-)

deluge/ui/baseargparser.py
bendikro: Well, the idea is that arguments are the cmd line arguments, and the options is the result of the parsed arguments.

cas--: Sorry I meant `handle UI options`

bendikro: Oh, you mean the docs...Yeah of course

PR #103, opened 2016-06-09
Develop ui entry tests

bendikro:


deluge/log.py
```diff
@@ -107,7 +107,7 @@ def findCaller(self):  # NOQA
 }
 
 
-def setup_logger(level="error", filename=None, filemode="w", logrotate=None):
+def setup_logger(level="error", filename=None, filemode="w", logrotate=None, twisted_observer=True):
```
cas--: Can you add a git commit message about this change.

deluge/ui/console/main.py
```diff
@@ -238,15 +239,18 @@ def __init__(self, options=None, cmds=None):
         # Set the interactive flag to indicate where we should print the output
         self.interactive = True
         self._commands = cmds
-        if options.parsed_cmds:
+        self.coreconfig = CoreConfig()
+
+    def start_ui(self):
+        ret = None
```
cas--: docstring for return value

deluge/ui/console/main.py
```diff
@@ -263,6 +267,8 @@ def __init__(self, options=None, cmds=None):
         else:
             reactor.run()
 
+        return ret
```
cas--: Think this be moved under `reactor.run()`

deluge/log.py
bendikro: Added comment on commit message:
"Added parameter log.setup_logger to prevent output noise in unit tests"

Is that good?

deluge/ui/console/main.py
bendikro: Removed the return value as it isn't needed.

deluge/ui/console/main.py
bendikro: Removed.

deluge/ui/console/main.py
bendikro: Crap, that was apparently just wrong.

bendikro:
Pushed working fixes

deluge/ui/console/main.py
```diff
@@ -238,15 +239,28 @@ def __init__(self, options=None, cmds=None):
...
+              replaced by a mock object, leaving the return deferred object
+              necessary for the tests to run properly.
+
+        Returns: Deferred that fires when all commands are executed.
+        """
+        deferred = None
+        if self.options.parsed_cmds:
             self.interactive = False
-            if not cmds:
+            if not self._commands:
```
cas--: Think that this can be refactored to bring the reactor.run and return into this statement.

bendikro: I just left the logic as it already was, but I wasn't too happy about it :-)
I'll rewrite it...

bendikro: Pushed a change to this function.

deluge/ui/console/main.py
```diff
@@ -250,33 +250,33 @@ def start_ui(self):
               replaced by a mock object, leaving the return deferred object
               necessary for the tests to run properly.
 
-        Returns: Deferred that fires when all commands are executed.
+        Returns: If valid commands are provived, a deferred that fires when
+                 all commands are executed. Else None is returned.
```
cas--: Should start indented under `Returns` with `Deferred: ...`

bendikro: Fixed.

cas--:
Applied: c8a3fd72d411

PR #104, opened 2016-06-29
[#2852] Set maximum supported libtorrent version to 1.0.x

cas--:


deluge/_libtorrent.py
```diff
@@ -46,11 +46,17 @@
 """
 
 REQUIRED_VERSION = "0.14.9.0"
+# The first major version (and all subsequent ones) to be unsupported.
+UNSUPPORTED_VERSION = "0.1.1.0"
```
minus7: Shouldn't it be 1.1.0.0?

cas--:
Committed: 852b51f224ef47

PR #105, opened 2016-07-10
Libtorrent 1.1 compatibility

minus7:
Fixed a few settings and the call to `post_torrent_updates`. Latter has a default value in the C++ API (`0xffffffff`) but not in the Python API. It would support the same flags as `torrent_handle::status(int flags)` but I didn't see an immediate way to do something useful with that.

I tried with lt 1.1.0 and 1.0.9 and it seems to work for me so far.

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -1352,7 +1352,9 @@ class TorrentManager(component.Component):
             # Ask libtorrent for status update
             self.torrents_status_requests.insert(0, (d, torrent_ids, keys, diff))
             if lt_1_1_compat:
-                self.session.post_torrent_updates(0)
+                # libtorrent 1.1 takes a combination of status_flags_t`s here
+                # like torrent_handle::status()
+                self.session.post_torrent_updates(0xffffffff)
             else:
                 self.session.post_torrent_updates()
         return d
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -23,7 +23,7 @@ from twisted.internet.task import LoopingCall
 
 import deluge.component as component
 from deluge._libtorrent import lt
-from deluge.common import decode_string, get_magnet_info, utf8_encoded
+from deluge.common import decode_string, get_magnet_info, utf8_encoded, VersionSplit
 from deluge.configmanager import ConfigManager, get_config_dir
 from deluge.core.authmanager import AUTH_LEVEL_ADMIN
 from deluge.core.torrent import Torrent, TorrentOptions, sanitize_filepath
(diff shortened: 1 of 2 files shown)


bendikro:
Since libtorrent v1.1 is backwards incompatible, we should aim for support without any deprecated functions in develop. Have you tried compiling without deprecated functions?

deluge/core/torrentmanager.py
```diff
@@ -1349,5 +1351,10 @@ def torrents_status_update(self, torrent_ids, keys, diff=False):
         else:
             # Ask libtorrent for status update
             self.torrents_status_requests.insert(0, (d, torrent_ids, keys, diff))
-            self.session.post_torrent_updates()
+            if lt_1_1_compat:
+                # libtorrent 1.1 takes a combination of status_flags_t`s here
+                # like torrent_handle::status()
+                self.session.post_torrent_updates(0xffffffff)
```
cas--: If this is the default flag then there should be no need for this change

minus7: You need to pass one parameter though (although it has a default value in C++). lt bug?

minus7:
I didn't compile lt without backwards compatibility, I just used the version provided on ArchLinux. Doing away with deprecated APIs sounds like a good idea though.

deluge/core/torrentmanager.py
cas--: Yes looks like a bug in the [binding](https://github.com/arvidn/libtorrent/blob/RC_1_1/bindings/python/src/session.cpp#L775)

This should fix it:
```
-    .def("post_torrent_updates", allow_threads(&lt::session::post_torrent_updates))
+   .def("post_torrent_updates", allow_threads(&lt::session::post_torrent_updates), arg("flags") = 0xffffffff)
```

bendikro: Should probably create an issue for this?

cas--: Done

minus7:
I'd suggest to merge this, so that lt 1.1 at least works with develop since lt with disabled deprecated APIs seems like a bigger change

cas--:
After a discussion on IRC lt 1.1.1 release should fix these compatibility issues, if not raise a libtorrent issue.

minus7:
Apparently there's some more incomatibilities with lt 1.1.1:

```
  File "/usr/lib/python2.7/site-packages/deluge/core/preferencesmanager.py", line 448, in _on_set_geoip_db_location
    self.session.load_country_db(str(geoip_db))
exceptions.AttributeError: 'session' object has no attribute 'load_country_db'
```
and
```
  File "/usr/lib/python2.7/site-packages/deluge/core/torrentmanager.py", line 1042, in on_alert_tracker_reply
    torrent.scrape_tracker()
  File "/usr/lib/python2.7/site-packages/deluge/core/torrent.py", line 1244, in scrape_tracker
    self.handle.scrape_tracker()
Boost.Python.ArgumentError: Python argument types in
    torrent_handle.scrape_tracker(torrent_handle)
did not match C++ signature:
    scrape_tracker(libtorrent::torrent_handle {lvalue}, int)
```

cas--:
Please report to libtorrent issue tracker

PR #107, opened 2016-09-06
[#2885] [CORE] Fix Unable to scrape tracker

derhoch:
This fixes the following issue:

Python argument types in
  torrent_handle.scrape_tracker(torrent_handle)
did not match C++ signature:
  scrape_tracker(libtorrent::torrent_handle {lvalue}, int)


cas--:
This bug needs to actually be fixed in libtorrent not Deluge.

PR #108, opened 2016-09-27
Check if libtorrent supports load_country_db before usage

minus7:
Fixes an uncaught exception with lt 1.1.1:
```
  File "/usr/lib/python2.7/site-packages/deluge/core/preferencesmanager.py", line 448, in _on_set_geoip_db_location
    self.session.load_country_db(str(geoip_db))
exceptions.AttributeError: 'session' object has no attribute 'load_country_db'
```

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -443,7 +443,7 @@ class PreferencesManager(component.Component):
         else:
             log.warning("Unable to find GeoIP database file!")
 
-        if geoip_db:
+        if geoip_db and hasattr(self.session, "load_country_db"):
             try:
                 self.session.load_country_db(str(geoip_db))
             except RuntimeError as ex:


cas--:
I'm sorry this has to be fixed in libtorrent as a deprecated method. By all means provide to package managers as interim workaround.

minus7:
I'm not sure how this should be solved then. To my understanding, the method was removed at your request to be able to distinguish of the feature is available or not: https://github.com/arvidn/libtorrent/issues/927

aresch:
This seems like a reasonable fix to allow Deluge to work with newer versions of libtorrent.  Why don't we just include it? It's harmless if the function is re-added to future libtorrent versions.

cas--:
I forgot about that change... So the actual fix is for the try block to also catch the AttributeError, at the time I think I assumed a RuntimeError would be raised.

@aresch I am trying to avoid any changes to 1.3 codebase so either 1.1 is backward compatible or it is not.

cas--:
Looking at the [1.3 code](https://github.com/deluge-torrent/deluge/blob/1.3-stable/deluge/core/preferencesmanager.py#L501) it has a general exception catch so that is fine.

I shall just commit the changes to develop code.

Thanks minus for investigating! :)

cas--:
Fix details in [#2882](http://dev.deluge-torrent.org/ticket/2882)

PR #109, opened 2016-10-03
[UI] Fix ZeroDivisionError at pieces bar update

elemongw:
I was getting the following traceback before. I was unable to pinpoint the exact circumstances under which this bug occurs.

```
  File "c:\users\Giorgos\downloads\portable\deluge\deluge\ui\gtkui\piecesbar.py", line 79, in do_expose_event
    self.__draw_pieces_completed()
  File "c:\users\Giorgos\downloads\portable\deluge\deluge\ui\gtkui\piecesbar.py", line 152, in __draw_pieces_completed
    piece_width = self.__width * 1.0 / self.__num_pieces
ZeroDivisionError: float division by zero
```

### DIFF
--- a/deluge/ui/gtkui/piecesbar.py
+++ b/deluge/ui/gtkui/piecesbar.py
@@ -73,12 +73,12 @@ class PiecesBar(gtk.DrawingArea):
         # Restrict Cairo to the exposed area; avoid extra work
         self.__roundcorners_clipping()
 
-        if not self.__pieces and self.__num_pieces is not None:
+        if self.__pieces:
+            self.__draw_pieces()
+        elif self.__num_pieces:
             # Special case. Completed torrents do not send any pieces in their
             # status.
             self.__draw_pieces_completed()
-        elif self.__pieces:
-            self.__draw_pieces()
 
         self.__draw_progress_overlay()
         self.__write_text()


cas--:
The most likely reason is magnets that have not yet retrieved metadata: https://github.com/deluge-torrent/deluge/blob/develop/deluge/core/torrent.py#L1035

So this is not the correct fix and instead the code at [L76](https://github.com/deluge-torrent/deluge/blob/develop/deluge/ui/gtkui/piecesbar.py#L76) should be altered to accommodate for this use-case.

elemongw:
It does not happen every time I download with a magnet link through. Any ideas why that may be? I am gonna do some more testing tomorrow in case I have better luck reproducing it. Also, should I remove the comment following line 76 or update it to describe this case too?

elemongw:
Nevermind you were right. How does it look now?

deluge/ui/gtkui/piecesbar.py
```diff
@@ -73,7 +73,7 @@ def do_expose_event(self, event):
         # Restrict Cairo to the exposed area; avoid extra work
         self.__roundcorners_clipping()
 
-        if not self.__pieces and self.__num_pieces is not None:
+        if not self.__pieces and self.__num_pieces > 0:
```
cas--: You shouldn't use numeric comparison where `None` is a known possible value. This should be a simple boolean comparison of self.__num_pieces as `0` and `None` return `False`.

What is the UI outcome if neither `__draw_pieces_completed()` or `__draw_pieces()` methods are called?

elemongw:
When neither is called an empty bar is drawn

deluge/ui/gtkui/piecesbar.py
elemongw: When neither is called an empty bar is drawn

deluge/ui/gtkui/piecesbar.py
```diff
@@ -73,7 +73,7 @@ def do_expose_event(self, event):
         # Restrict Cairo to the exposed area; avoid extra work
         self.__roundcorners_clipping()
 
-        if not self.__pieces and self.__num_pieces is not None:
+        if not self.__pieces and self.__num_pieces is not None and self.__num_pieces != 0:
```
elemongw: Yeah I thought you wanted me to write it explicitly, I am gonna fix it soon.

elemongw:
Okay this should be as you asked. However, as you brought up refactoring maybe we should also look at __draw_pieces() and __draw_pieces_completed() as they could probably be merged into one function.

cas--:
Cheers. If you are happy to try refactoring it, go ahead, it does look rather copy-pasted.

cas--:
I've merged this change to fix the issue but feel free to open a new PR with refactored code. 6b630c9

PR #111, opened 2016-10-11
[#2889] Fix missing file closure statement

cas--:
* Convert all open file code to use 'with' statement.


cas--:
- [ ] Put tracker icon tempfiles in subfolder

PR #112, opened 2016-10-11
[#2900] Fix Error loading torrent: invalid bencoded value

bendikro:
Proposed fix is to revert upstream fix in bencode.py

Issues:
- [ ] The included torrent does not produce the bencode error

### DIFF
--- a/deluge/bencode.py
+++ b/deluge/bencode.py
@@ -15,6 +15,10 @@
 from types import DictType, IntType, ListType, LongType, StringType, TupleType
 
 
+class BTFailure(Exception):
+    pass
+
+
 def decode_int(x, f):
     f += 1
     newf = x.index('e', f)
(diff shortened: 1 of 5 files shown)


bendikro:
Ah, yeah, I forgot to label the last commit with the ticket #2875.
But I don't see what the web test should be a separate commit? It is added to verify that the bug is fixed.
I do see that it may look cleaner, but imo, tests and implementation really goes together.

cas--:
To me it looks like Web test refactoring that's not directly related to the bug?

bendikro:
Well, `deluge/tests/test_webserver.py ` contains the new test, but there were some refactoring to reduce code reuse.

cas--:
It'd be best to have an initial 'refactor web test code' then 'fix bencode' followed by 'fix filehash (with tests)'. Currently the commits are not distinct with filehash tests in bencode commit...

Keeping issue commits clean is important for historical viewing, blame and merging especially where there is a clear defined bug and fix.

bendikro:
Ahh, I see the misunderstanding. The test together with `[#2900] Fix Error loading torrent: invalid bencoded value` was not supposed to test the filehash field, but the bencode issue, but I think I forgot that somehow, and tested the filehash twice. After creating the new torrent with the filehash field, the bencode issue cannot be reproduced on that torrent, so the test is pointless atm.

I'll move the bencode issue into a separate PR.

cas--:
Thanks for effort on this! Merged: 41fed16d

PR #115, opened 2016-10-16
Use libtorrent.__version__ since libtorrent.version is deprecated

aresch:


cas--:
I think until we have made all the other changes that require 1.1 we should wait to do this

aresch:
I guess we could do it in a way that is backwards compatible, ie, check for __version__ and if not available use version.

cas--:
It's not worth it as it's only deprecated currently. It can be included in a wider change for develop to support 1.1 with no deprecated methods.

aresch:
Are we not targeting >lt-1.1 for develop now?

The reason I want this patch in, is because I run with a non-deprecated functions libtorrent build and without this the thin client will not start on my desktop.

cas--:
Yes but 1.1 is still backward compatible so we don't need to do anything and this PR would force a change for no real reason, meaning everyone else requires updating libtorrent. It makes sense for this change to be part of a larger branch that fixes all the deprecated methods.

cas--:
Ok I think I understand now that you mean it's a client side issue?

cas--:
Should we be looking at modifying this code: https://github.com/deluge-torrent/deluge/blob/develop/deluge/ui/baseargparser.py#L87

cas--:
Is it an AttributeError so we can do this?

    try:
        lt_version = libtorrent.version
    except AttributeError:
        lt_version = libtorrent.__version__

cas--:
This is a result of the overhaul of UI and I think better fixed in baseargparser: https://github.com/deluge-torrent/deluge/blob/develop/deluge/ui/baseargparser.py#L87

    ...
    except (ImportError, AttributeError):
        pass
    ...

aresch:
It needs to be fixed in 2 places.  The other place being _libtorrent.py where we read the libtorrent version to ensure we meet the minimum required version.

aresch:
I just installed libtorrent 1.0.10 to see if __version__ exists there and it does.  It looks like this change should work fine with older versions of libtorrent.

```
>>> libtorrent.__version__
'1.0.10.0'
>>> libtorrent.__version__
'1.0.10.0'
```

cas--:
Ah it was added in 1.0.7 so the minimum version will need bumped to that.

cas--:
It will also need changing in these places:
```
git grep lt.version
.travis.yml:43:  - python -c "import libtorrent as lt; print lt.version"
deluge/core/core.py:60:        log.info("Starting libtorrent %s (%s, %s) session...", lt.version, client_id, client_version)
deluge/core/core.py:68:        settings["user_agent"] = "Deluge/%(deluge_version)s libtorrent/%(lt_version)s" % {
deluge/core/core.py:70:            'lt_version': self.get_libtorrent_version().rpartition(".")[0]
deluge/core/core.py:951:        return lt.version
deluge/ui/baseargparser.py:91:        version_str += "libtorrent: %s\n" % lt.version
deluge/ui/gtkui/aboutdialog.py:264:            def on_lt_version(result):
deluge/ui/gtkui/aboutdialog.py:273:                client.core.get_libtorrent_version().addCallback(on_lt_version)
deluge/ui/gtkui/aboutdialog.py:278:                client.core.get_libtorrent_version().addCallback(on_lt_version)
setup.py:234:            print('Found libtorrent version: %s' % lt.version)
tox.ini:50:    python -c "import libtorrent as lt; print lt.version"
```

cas--:
I decided to tidy up _libtorrent.py so have opened a new PR #116 with all the relevant changes.

PR #117, opened 2016-10-18
Python 3 fixes

aresch:
This is a series of changes to fix the code for Python 3 compatibility.  Most of these changes shouldn't cause any regressions and should be relatively safe.  This PR does not provide Python 3 support for Deluge, but was simply an attempt to modernize the code as best I can before we do more invasive changes.  The next steps will be more involved and will require more thorough work.

There are a lot of changes here and I'd really like some additional eyes to review.  Thanks :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ virtualenv:
 # We use xvfb for the GTKUI tests
 before_script:
   - export PYTHONPATH=$PYTHONPATH:$PWD
-  - python -c "import libtorrent as lt; print lt.version"
+  - python -c "import libtorrent as lt; print lt.__version__"
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16"
   - echo '2.0.0.dev0' > RELEASE-VERSION
 
--- a/DEPENDS
+++ b/DEPENDS
@@ -1,5 +1,5 @@
 === Core ===
- * libtorrent (rasterbar) >= 0.16.7
+ * libtorrent (rasterbar) >= 1.0.7
  * python >= 2.6
  * setuptools
  * twisted >= 11.1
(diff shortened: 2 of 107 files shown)


cas--:
That is helpful, it's odd that I missed some of these from my last sweep :)

I think the iteritems changes should be done with six or similar as it could have a memory performance hit on python 2. However if there are areas where is make no obvious difference then then less we rely on py2to3 shims the better.

Oh there are several pylint issues needing attention.

aresch:
Yes, there could be some performance changes due to removing iteritems, etc.. but I don't think it's going to be that big of a deal since most of our data structures don't contain that many items.  I really wanted this change to not introduce any outside dependencies, plus it will give us more incentive to get the rest of  things Python 3 ready :)

I'll try to fix the lint errors.

aresch:
I have no idea why this isort run fails.  On my machine I get no diffs..

[aresch@squid deluge]$ isort -q --diff --recursive deluge docs packaging *.py
[aresch@squid deluge]$ echo $?
0

bendikro:
What if you run `tox -e isort`?

aresch:
Yea, if I run tox2 -e isort I get the same output from travis.  I'll just make a change based on that diff.

cas--:
Those changes to `new` and `HTMLParser` imports are definitely incorrect, what version of isort is installed via pip? I have 4.2.5 and it gives the same output as tox.

cas--:
Just noted that the `new` import should be changed for Py3 as that module has been [removed](https://docs.python.org/2/library/new.html). Which funnily enough if you also take into account that [HTMLParser](https://docs.python.org/2/library/htmlparser.html) is renamed in Py3 explains your isort issue... :P

aresch:
I think this looks OK to merge now?  There are plenty more fixes necessary for Python 3, but we should get this merged now so we have a better base to work from for the future fixes.

cas--:
I've a new pull request with the branch rebased and some extra changes #124

 * Rebasing the develop branch rather than merging it in is better as it creates quite a commit mess otherwise.
 * Using `git commit --fixup/--squash` or modifiyng the commits that need fixed is much easier to rebase into a short set of commits.
 * For pylint we use the full name rather than the shortcode and place the disable comment inline with the offending code.

PR #119, opened 2016-10-19
[Docs] Clean module sources dir before generating docs

bendikro:
If old and outdated sources are present in docs/sources/modules,
python setup.py build_docs will fail to generate docs.


cas--:
Merged: 75714b60ca

PR #120, opened 2016-10-19
[#2861] [Core] Add support for python-geoip

cas--:
* libtorrent >= 1.1 dropped support for GeoIP so this add support
   using MaxMind GeoIP Legacy Python Extension API.
 * It is known by the following package names:
       * Maxmind: geoip-api-python
       * Linux: python-geoip
       * PyPi: GeoIP

### DIFF
--- a/DEPENDS
+++ b/DEPENDS
@@ -1,5 +1,5 @@
 === Core ===
- * libtorrent (rasterbar) >= 0.16.7
+ * libtorrent (rasterbar) >= 1.0.7
  * python >= 2.6
  * setuptools
  * twisted >= 11.1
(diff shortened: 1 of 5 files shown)


aresch:
I feel like we shouldn't have an instance to the geoip instance in the PreferencesManager.  Maybe we should store it in the Core?  We could even create a new component for it, but that might be a bit overkill. What do you think?

cas--:
Storing it in core is perhaps a better idea, it is the one detail that I wasn't sure about.

I have made that change and realised that for develop code we can just switch over to GeoIP now so it's a much simpler changeset and user simply needs python-geoip installed.

PR #122, opened 2016-10-20
[GtkUI] Avoid duplicate code in piecesbar's drawing logic

elemongw:


deluge/ui/gtkui/piecesbar.py
```diff
@@ -124,11 +119,10 @@ def __draw_pieces(self):
...
             ctx = cairo.Context(self.__pieces_overlay)
             start_pos = 0
-            num_pieces = self.__num_pieces and self.__num_pieces or len(self.__pieces)
-            piece_width = self.__width * 1.0 / num_pieces
+            piece_width = self.__width * 1.0 / (self.__num_pieces or len(self.__pieces))
 
-            for state in self.__pieces:
-                color = self.gtkui_config["pieces_color_%s" % COLOR_STATES[state]]
+            def fill():
+                global start_pos
```
cas--: Really should avoid using global.. In this situation a good alternative would be to enumerate for loop and pass as an arg to fill function. 

However if you implement the suggestion in my other comment below you can remove this function and put the code back into the for loop.

elemongw:
I wanted to avoid doing it this way so as to not introduce any extra potential slowness.

deluge/ui/gtkui/piecesbar.py
```diff
@@ -124,10 +114,14 @@ def __draw_pieces(self):
...
-            piece_width = self.__width * 1.0 / num_pieces
-
-            for state in self.__pieces:
+            piece_width = self.__width * 1.0 / (self.__num_pieces or len(self.__pieces))
+
+            if self.__pieces:
+                pieces = self.__pieces
+            elif self.__num_pieces:
+                # Special case. Completed torrents do not send any pieces in their status.
+                pieces = (COLOR_STATES[3] for _ in xrange(self__num_pieces))
```
cas--: This doesn't look right... in my example it was: `[3] * self.__num_pieces`

cas--:
Which bit of code would be slow?

cas--:
Thanks for these changes, I made a few tweaks and is now merged: c4282f29abbd

PR #123, opened 2016-10-21
[#2133] [WebUI] Modify UI to display single incoming port

cas--:
Convert to a single port in WebUI preferences.

Issues:
- [x] The option manager gets and sets the value from core so needs translating from an int to a list of identical ints for `listen_ports`.


sjnewbury:
Why was this change made?  Is it no longer possible to supply a list of ports?

PR #124, opened 2016-10-24
Feature/python3 support

cas--:
A rebase and continuation of the work in PR #117

Issue needing reviewed:
 - [x] In [` deluge/ui/console/commands/config.py`](https://github.com/deluge-torrent/deluge/pull/124/files#diff-50c098ec747e111ab711aee1a88205aeL61) is the change of `src.next` to `src` correct?
 - [x] Can use `from future_builtins import zip` to retain use of izip in Py2.

------ 

To verify code at this stage `python-modernize` can be used without `six` dependency. (Ignore long to int conversions in rencode and test_core):
`python-modernize --no-six -wn -j 2 --no-diffs -x libmodernize.fixes.fix_import deluge`
or
`futurize -1 -w -n --no-diffs -x absolute_import deluge`

So far from futurize Stage2 confirmed only these pass:

libfuturize.fixes.fix_division_safe 
libfuturize.fixes.fix_execfile
lib2to3.fixes.fix_long

- [ ] Use newstyle classes to fix: `libpasteurize.fixes.fix_newstyle`

This may help decide which direction we want to go with the codebase for Py3: 

http://python-future.org/faq.html#what-is-the-relationship-between-future-and-six

- [ ] Note must add these details to ticket or wiki
 - http://dev.deluge-torrent.org/ticket/1328
 - http://dev.deluge-torrent.org/wiki/Plan/Python3

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -10,6 +10,7 @@
 """Common functions for various parts of Deluge to use."""
 
 import base64
+import functools
 import locale
 import logging
 import numbers
@@ -818,6 +819,7 @@ def utf8_encoded(s, encoding="utf8"):
     return s
 
 
+@functools.total_ordering
 class VersionSplit(object):
     """
     Used for comparing version numbers.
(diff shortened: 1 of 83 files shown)


aresch:
The changes to the commands/config.py look correct to me.  The next() function will call the correct function on the object to get the next iterable, .next() for Python2 or .__next__() for Python3.  Instead of passing the 'next function' to the atom function, we just pass it the object and let it do next(obj) instead of calling .next() or .__next__() on the object directly.

aresch:
The changes to import the zip from future_builtins will require a new dependency for Deluge.  This is probably OK, but we need to make sure it ends up in DEPENDS at a minimum.  As an aside, have we considered creating a proper requirements.txt file instead of the DEPENDS file? This would make it easier for people to do a simple pip install -r requirements.txt to fetch (most) of what's necessary.

aresch:
The fixes to division look good.  I was a bit worried about doing this and was going to do it in a separate PR, but I think what you've done makes sense.  The other option would have been to import in an old_division function from futurize but I wasn't terribly keen on that.

cas--:
What new dependency for [future_builtins](https://docs.python.org/2/library/future_builtins.html), it's included in Python 2. 

I have been looking at requirements.txt for pip but I'm not sure how much it will help on Linux as all the requirements are packaged and still need gtk and libtorrent.

aresch:
So it is.  I must have tried importing it with python3.

This PR looks good to me.

cas--:
@andar I have had to revert the iteritems changes as it has too many issues needing resolved. Namely `exceptions.RuntimeError: dictionary changed size during iteration`. Either we go through each change and decide what exact operation is needed or use `future` to apply the changes with the py2 shim.

aresch:
Yup, that sound good to me.  We can fix it or import something from the futurize package like you said.

PR #127, opened 2016-10-30
Feature/fixup lt1.1 deprecated

cas--:
This is a refactor that works with both 1.0 and 1.1 (with/without deprecated) libtorrent branches. It is designed to try non-deprecated and fallback to deprecated with the aim to make it trivial to remove the code in future.

I am keeping compatibility for now to ease testing (I still use 1.0.9 for dev...) but likely will remove support for 1.0 and 1.1 deprecated methods shortly before release.

Still to be done:
- [x] ~~Need to merge i2p proxy into proxy settings in UI~~ See [#2919](http://dev.deluge-torrent.org/ticket/2919)

- [x] ~~Session stats still need converting~~ See [#2920](dev.deluge-torrent.org/ticket/2920)

- [x] ~~Could move and combine all the `_on_set_*` methods for session settings into `do_config_set_func` testing if key in settings_pack.  `preferencesmanager.session_set_setting` would become redundant.~~ See [#2921](http://dev.deluge-torrent.org/ticket/2921)

- [x] `core.apply_session_setting` perhaps should be plural and accept `dict` instead of `key, value`

- [x] Remove unneeded `preferencesmanager.session_set_settings`

- [x] ~~lt `listen_interfaces` and `outgoing_interfaces` now supports adapter names.~~ (advised against use by arvid in 1.1, plus no Windows support atm)

- [x] ~~Add support for 'new' `outgoing_interfaces` setting?~~ Leave it for now.

- [x] `tracker_error_alert`: `msg` has been replace by `error_message`

- [x] `filerenamed_alert`: `name` is now `new_name`

- [x] Use of `torrent.handle` `save_path` in alert should use alert path attribute.

- [x] torrent `status.error` is split into `errc` (error code) and `error_file`.

- [x] ~~`states.queued_for_checking` has been removed.~~ Should be ok to be left as is.

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -158,9 +158,19 @@ class PreferencesManager(component.Component):
             on_set_func(key, value)
 
     def session_set_setting(self, key, value):
-        settings = self.session.get_settings()
-        settings[key] = value
-        self.session.set_settings(settings)
+        try:
+            self.session.apply_settings({key: value})
+        except AttributeError:
+            # Deprecated in libtorrent 1.1
+            if key in ("enable_lsd", "enable_upnp", "enable_natpmp", "enable_dht"):
+                start_stop = key.replace("enable", "start") if value else key.replace("enable", "stop")
+                getattr(self.session, start_stop)()
+            elif key == "dht_bootstrap_nodes":
+                self.session.add_dht_router("router.bittorrent.com", 6881)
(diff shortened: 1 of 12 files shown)


deluge/core/preferencesmanager.py
```diff
@@ -185,6 +200,7 @@ def _on_set_random_port(self, key, value):
         self.__set_listen_on()
 
     def __set_listen_on(self):
+
```
cas--: remove newline

PR #128, opened 2016-10-30
[Tests] Increase file descriptor limit

bendikro:


cas--:
I think `openbt.png` can be deleted, I removed the test but forgot about the icon.

bendikro:
Removed

deluge/tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Increase open file descriptor limit to allow tests to run
+# without getting error: what():  epoll: Too many open files
+import resource
+
+try:
+    resource.setrlimit(resource.RLIMIT_NOFILE, (65536, 65536))
```
cas--: Does Windows fail with ValueError for this?

deluge/tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Increase open file descriptor limit to allow tests to run
+# without getting error: what():  epoll: Too many open files
+import resource
```
cas--: This will `ImportError` on Windows

bendikro:
Does this look better?

deluge/tests/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+
+try:
+    resource.setrlimit(resource.RLIMIT_NOFILE, (65536, 65536))
+except ValueError as ex:
+    print "Failed to raise file descriptor limit:", ex
```
aresch: Please use the print function from __future__: from __future__ import print_function

bendikro: Sure

cas--:
Merged, thanks!

PR #129, opened 2016-10-31
[Tests] Fix for isort config package handling

bendikro:


cas--:
Tidied up the config as it should not require specifying third-party modules. (fingers crossed passes travis...)

PR #130, opened 2016-10-31
[Lint] Fix pylint warnings

bendikro:
TODO:
* [x] Enable extra warnings for console and fix those ([#2797](http://dev.deluge-torrent.org/ticket/2797))

### DIFF
--- a/deluge/ui/Win32IconImagePlugin.py
+++ b/deluge/ui/Win32IconImagePlugin.py
@@ -141,7 +141,7 @@ class Win32IcoFile(object):
             # figure out where AND mask image starts
             mode = a[0]
             bpp = 8
-            for k in PIL.BmpImagePlugin.BIT2MODE.keys():
+            for k in PIL.BmpImagePlugin.BIT2MODE:
                 if mode == PIL.BmpImagePlugin.BIT2MODE[k][1]:
                     bpp = k
                     break
@@ -214,7 +214,7 @@ class Win32IcoFile(object):
 # end Win32IcoFile
 
 
-class Win32IconImageFile(PIL.ImageFile.ImageFile):
+class Win32IconImageFile(PIL.ImageFile.ImageFile):  # pylint: disable=abstract-method
     """
     PIL read-only image support for Microsoft .ico files.
 
--- a/deluge/ui/console/widgets/inputpane.py
+++ b/deluge/ui/console/widgets/inputpane.py
@@ -31,6 +31,7 @@ class BaseInputPane(InputKeyHandler):
     def __init__(self, mode, allow_rearrange=False, immediate_action=False, set_first_input_active=True,
                  border_off_west=0, border_off_north=0, border_off_east=0, border_off_south=0,
                  active_wrap=False, **kwargs):
+        InputKeyHandler.__init__(self)
         self.inputs = []
         self.mode = mode
         self.active_input = 0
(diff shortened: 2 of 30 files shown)


cas--:
* I fixed the pylint rcfile as broad-except was enabled as comments must start at beginning of line.

* The `connectionDone` listed in #2797 was a clipboard pasting mistake. It should have actually been `invalid-name` as I now recall that console code was consistently using single letter variable names. I think I have come to the conclusion that fixing the codebase for this convention is highly disruptive and unlikely to be enabled. 

* Outstanding issues needing fixed:
  - [ ] `deluge/ui/Win32IconImagePlugin.py:144: consider-iterating-dictionary`
  - [ ] `deluge/ui/Win32IconImagePlugin.py:217: abstract-method`
  - [ ] `deluge/ui/console/__init__.py:12: wrong-import-position`

* Docs test failing with the following error:
  - [ ] `Double requirement given: pillow (already in pillow, name='pillow')`

* Docs test showing deprecation warning:
  - [ ] `--download-cache has been deprecated and will be removed in the future. Pip now automatically uses and configures its cache.`

* It would be good if these could be fixed in this PR:
  - [ ] `arguments-differ`
  - [ ] `global-statement`

cas--:
Below is the output of pylint with the entire Refactor category enabled and count for each message found. What I propose is to fix the messages that have only a handful of counts.

```
pylint --msg-template='{symbol}' deluge > pylint_count.txt
pylint --msg-template='{symbol}' *.py deluge/scripts/*.py >> pylint_count.txt
pylint --msg-template='{symbol}' deluge/plugins/*/deluge/ >> pylint_count.txt
sort pylint_count.txt | uniq -c | grep -v "*"
```


```
      1 consider-iterating-dictionary
    186 no-self-use
     14 redefined-variable-type
      2 simplifiable-if-statement
     71 too-few-public-methods
      1 too-many-ancestors
     32 too-many-arguments
     58 too-many-branches
     53 too-many-instance-attributes
     36 too-many-locals
      6 too-many-nested-blocks
     14 too-many-public-methods
      5 too-many-return-statements
     25 too-many-statements
      1 using-constant-test
      1 wrong-import-position
```

deluge/ui/console/widgets/inputpane.py
```diff
@@ -31,6 +31,7 @@ class BaseInputPane(InputKeyHandler):
     def __init__(self, mode, allow_rearrange=False, immediate_action=False, set_first_input_active=True,
                  border_off_west=0, border_off_north=0, border_off_east=0, border_off_south=0,
                  active_wrap=False, **kwargs):
+        super(BaseInputPane, self).__init__()
```
cas--: This change has broken deluge-console

bendikrof: Fixed

bendikro: Fixed

deluge/ui/Win32IconImagePlugin.py
```diff
@@ -215,6 +215,7 @@ def __repr__(self):
 
 
 class Win32IconImageFile(PIL.ImageFile.ImageFile):
+    # pylint: disable=W0223
```
cas--: I'll fix this but should be using symbol name `abstract-method` and where possible appended to the line that raises the message.

PR #133, opened 2016-11-02
[WebUI] Log correct http address if listening on IPv6

bendikro:


deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
     """
     A simple test to see if 'ip' is valid
```
cas--: Docstrings should start on the first line as opening quotes: `"""A simple test...`

deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
...
-    :type ip: string
-    :returns: True or False
-    :rtype: bool
+    Args:
+        ip (str): The IP to test
+
+    Returns:
+        bool: Whether IP is valid is not
 
     ** Usage **
```
cas--: Replace usage with `Examples:` and indent the code below

deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
...
+
+def is_ipv6(ip):
+    import socket
+    if not socket.has_ipv6:
+        return False
     try:
         if windows_check():
             log.warning("ipv6 check unavailable on windows")
-            return True
+            return None
```
cas--: I don't think this should be `None` and it can't be `False` as we don't know so kind of leaves us back at `True`?

deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
...
                 return True
     except socket.error:
-        if not socket.has_ipv6:
-            return False
-    # now test ipv6
+        return False
+
+
+def is_ipv6(ip):
+    import socket
```
cas--: Needs docstring

deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
...
 
     >>> is_ip("127.0.0.1")
     True
 
     """
+    return is_ipv4(ip) or is_ipv6(ip)
+
+
+def is_ipv4(ip):
     import socket
```
cas--: Needs docstring

deluge/common.py
```diff
@@ -730,34 +730,41 @@ def is_ip(ip):
...
 
-    :param ip: the ip to check
-    :type ip: string
-    :returns: True or False
-    :rtype: bool
+    Args:
+        ip (str): The IP to test
+
+    Returns:
+        bool: Whether IP is valid is not
```
cas--: All docstring sentences should end with a fullstop `.`.

cas--:
Fixed docstrings and added extra tests using proper documentation IPs: [rfc5737](https://tools.ietf.org/html/rfc5737), [rfc3849](https://tools.ietf.org/html/rfc3849)

bendikro:
Excellent!

cas--:
Still need a decision on the return type...

cas--:
I have rebased and pushed a modified is_ipv6

deluge/ui/web/server.py
```diff
@@ -649,7 +652,9 @@ def start_ssl(self):
         options.getContext().set_options(SSL.OP_NO_SSLv2 | SSL.OP_NO_SSLv3)
 
         self.socket = reactor.listenSSL(self.port, self.site, options, interface=self.interface)
-        log.info('Serving at https://%s:%s%s', self.interface, self.port, self.base)
+        ip = self.socket.getHost().host
+        ip = '[%s]' % ip if is_ipv6(ip) else ip
```
cas--: I'm not sure of the purpose of this?

cas--:
Need to add python-interface as optional dependency to DEPENDS and perhaps add to pip install for tests

PR #138, opened 2016-12-04
host:port paste support

matiasw:
added a feature so that when host:port form data is pasted into proxy host field, it gets split into host and port fields

### DIFF
--- a/deluge/ui/gtkui/connectionmanager.py
+++ b/deluge/ui/gtkui/connectionmanager.py
@@ -12,6 +12,7 @@ import logging
 import os
 import time
 from socket import gethostbyname
+from urlparse import urlparse
 
 import gtk
 from twisted.internet import reactor
@@ -232,6 +233,21 @@ class ConnectionManager(component.Component):
 
         # Update the status of the hosts
         self.__update_list()
+
+    def on_entry_host_paste_clipboard(self, widget):
+        text = gtk.clipboard_get().wait_for_text().strip()
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', text)
+        text = text if '//' in text else '//' + text
+        parsed = urlparse(text)
+        if parsed.hostname:
+            widget.set_text(parsed.hostname)
+            widget.emit_stop_by_name('paste-clipboard')
+        if parsed.port:
+            self.builder.get_object('spinbutton_port').set_value(parsed.port)
+        if parsed.username:
+            self.builder.get_object('entry_username').set_text(parsed.username)
+        if parsed.password:
+            self.builder.get_object('entry_password').set_text(parsed.password)
 
     # Private methods
     def __save_hostlist(self):
--- a/deluge/ui/gtkui/preferences.py
+++ b/deluge/ui/gtkui/preferences.py
@@ -11,6 +11,7 @@
 import logging
 import os
 from hashlib import sha1 as sha
+from urlparse import urlparse
 
 import gtk
 from gtk.gdk import Color
@@ -134,6 +135,7 @@ class Preferences(component.Component):
             'on_button_cache_refresh_clicked': self._on_button_cache_refresh_clicked,
             'on_combo_encryption_changed': self._on_combo_encryption_changed,
             'on_combo_proxy_type_changed': self._on_combo_proxy_type_changed,
+            'on_entry_proxy_host_paste_clipboard': self._on_entry_proxy_host_paste_clipboard,
             'on_button_associate_magnet_clicked': self._on_button_associate_magnet_clicked,
             'on_accounts_add_clicked': self._on_accounts_add_clicked,
             'on_accounts_delete_clicked': self._on_accounts_delete_clicked,
@@ -977,6 +979,21 @@ class Preferences(component.Component):
                 self.builder.get_object(entry).show()
             else:
                 self.builder.get_object(entry).hide()
+
+    def _on_entry_proxy_host_paste_clipboard(self, widget):
+        text = gtk.clipboard_get().wait_for_text().strip()
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', text)
+        text = text if '//' in text else '//' + text
+        parsed = urlparse(text)
+        if parsed.hostname:
+            widget.set_text(parsed.hostname)
+            widget.emit_stop_by_name('paste-clipboard')
+        if parsed.port:
+            self.builder.get_object('spin_proxy_port').set_value(parsed.port)
+        if parsed.username:
+            self.builder.get_object('entry_proxy_user').set_text(parsed.username)
+        if parsed.password:
+            self.builder.get_object('entry_proxy_pass').set_text(parsed.password)
 
     def _on_button_associate_magnet_clicked(self, widget):
         associate_magnet_links(True)
--- a/deluge/ui/gtkui/glade/connection_manager.addhost.ui
+++ b/deluge/ui/gtkui/glade/connection_manager.addhost.ui
@@ -114,6 +114,7 @@
                     <property name="secondary_icon_activatable">False</property>
                     <property name="primary_icon_sensitive">True</property>
                     <property name="secondary_icon_sensitive">True</property>
+                    <signal name="paste-clipboard" handler="on_entry_host_paste_clipboard" />
                   </object>
                 </child>
               </object>
--- a/deluge/ui/gtkui/glade/preferences_dialog.ui
+++ b/deluge/ui/gtkui/glade/preferences_dialog.ui
@@ -3524,6 +3524,7 @@ used sparingly.</property>
                                             <property name="secondary_icon_activatable">False</property>
                                             <property name="primary_icon_sensitive">True</property>
                                             <property name="secondary_icon_sensitive">True</property>
+                                            <signal name="paste-clipboard" handler="on_entry_proxy_host_paste_clipboard" />
                                           </object>
                                           <packing>
                                             <property name="left_attach">1</property>


cas--:
Can this handle pasting an ipv6 address?

matiasw:
Not really, the last part of the address (after the right-most ":") would get interpreted as a port number.

cas--:
So it is an issue if you paste an ipv6 address without a port...

matiasw:
Ok, it now correctly handles ipv6 addresses, and parses the port from the [2001:0:3238:DFE1:63::FEFB]:3188 style.

cas--:
Ok can you fix the errors and rebase into a single commit

matiasw:
Done. Just so you know, the pydef check of Travis-CI failed twice with different results, after I had fixed my errors, which is why I needed to re-open this pull request. So, I think there's something wonky with it.

matiasw:
Ok, I've done that now. 
I noticed that there is also a host, port duo in the connection manager window > edit. Would you like to have this implemented there, as well?

matiasw:
Done.
I noticed that there is also a host, port duo in Connection Manager > Edit. Would you like this to be implemented there, as well? If so, should I make a separate commit about it? And where would you like the newly-created parse_host_port utility, in that case?

cas--:
Have you rebased from latest develop because there is no separate `i2p` anymore

deluge/ui/gtkui/preferences.py
```diff
@@ -975,6 +977,45 @@ def _on_combo_proxy_type_changed(self, widget):
...
+
+    def _on_entry_proxy_host_paste_clipboard(self, widget):
+        paste = gtk.clipboard_get().wait_for_text()
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', paste)
+        host, port = self.parse_host_port(paste)
+        if host is not None:
+            widget.set_text(host)
+            widget.emit_stop_by_name('paste-clipboard')
+            try:
+                self.builder.get_object('spin_proxy_port').set_value(int(port))
```
cas--: Checking if the port is a valid int should be done in `parse_host_port`

deluge/ui/gtkui/preferences.py
```diff
@@ -975,6 +977,45 @@ def _on_combo_proxy_type_changed(self, widget):
         for show_entry in shows:
             self.builder.get_object(show_entry).show()
 
+    def parse_host_port(self, paste):
```
cas--: a better descriptive name would be `split_ip_port` also it needs a docstring.

deluge/ui/gtkui/preferences.py
```diff
@@ -975,6 +977,45 @@ def _on_combo_proxy_type_changed(self, widget):
...
+                    ip, port = paste[paste.find(']'):].rsplit(':', 1)
+                    ip = paste[:paste.find(']') + 1]
+                except ValueError:   # not an [address]:port formatted ipv6
+                    return
+            else:
+                ip, port = paste.rsplit(':', 1)
+            log.debug('parsed host and port: %s, %s', ip, port)
+            return (ip, port)
+        else:
+            return None
```
cas--: Should return an empty tuple

deluge/ui/gtkui/preferences.py
```diff
@@ -975,6 +977,45 @@ def _on_combo_proxy_type_changed(self, widget):
...
+
+    def _on_entry_proxy_host_paste_clipboard(self, widget):
+        paste = gtk.clipboard_get().wait_for_text()
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', paste)
+        host, port = self.parse_host_port(paste)
```
cas--: What happens if the return from method is None or empty tuple?

matiasw:
I added the functionality to the  connection manager window, as well.

cas--:
Ok looking better. 

The `split_host_port` function needs improvement and simplified. 

- Use `host` not `ip` as variable name in the function to keep consistency. 
- Start by setting the two variable to `None` and then the last step of the function is `return host, port`
- The tuple is implicit for returning multiple values so can drop the brackets.
- If you can't extract a port then at least pass back the `host` value with `port` set `None`.
- Use int not float for port.
- Docstring should be google style.
- Use `str.partition` to get ipv6 host and port.
- A better descriptive name than `paste` for the string parameter.

gazpachoking:
Maybe using the stdlib would be more clear for that split function?
```python
def split_host_port(paste):
    parse_result = urlparse('//' + paste)
    return parse_result.host, parse_result.port
```

matiasw:
Yeah I changed it to use urlparse. Thanks for the tip! It now supports pasting username and password, as well. 
It would also be possible to have `https://' type scheme identifiers change the proxy type automatically, what do you think?
I believe this version should work, but I'm suddenly unable to test the preferences, because the window won't open, and I get  

> _libtorrent.py:24: RuntimeWarning: to-Python converter for boost::shared_ptr<libtorrent::alert> already registered; second conversion method ignored.
>   import libtorrent as lt
> 18:08 $ Traceback (most recent call last):
>   File "install/deluged", line 11, in <module>
>     load_entry_point('deluge', 'console_scripts', 'deluged')()
>   File "/home/appas/code/deluge/deluge/core/daemon_entry.py", line 55, in start_daemon
>     from deluge.core.daemon import is_daemon_running
>   File "/home/appas/code/deluge/deluge/core/daemon.py", line 21, in <module>
>     from deluge.core.core import Core
>   File "/home/appas/code/deluge/deluge/core/core.py", line 27, in <module>
>     from deluge._libtorrent import lt
>   File "/home/appas/code/deluge/deluge/_libtorrent.py", line 31, in <module>
>     'Ubuntu users can add the develop PPA: ppa:deluge-team/develop' % get_version())
> ImportError: Deluge 2.0.0.dev898 requires latest github code from libtorrent RC_1_1 branch.
> Ubuntu users can add the develop PPA: ppa:deluge-team/develop

Now, this seems to be because hasattr(lt, 'generate_fingerprint') is False. Why? I'm on the latest develop HEAD and libtorrent 1.1.1.0.

cas--:
> I'm on the latest develop HEAD and libtorrent 1.1.1.0.

```
Deluge 2.0.0.dev898 requires latest github code from libtorrent RC_1_1 branch.
Ubuntu users can add the develop PPA: ppa:deluge-team/develop
```

> where am I supposed to place these imports?

Run `isort` on the file and it will update the changes required automatically.

deluge/ui/gtkui/connectionmanager.py
```diff
@@ -233,6 +234,23 @@ def add_host(self, host, port, username='', password=''):
...
+    def on_entry_host_paste_clipboard(self, widget):
+        paste = gtk.clipboard_get().wait_for_text()
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', paste)
+        if '//' not in paste:
+            parsed = urlparse('//' + paste)
+        else:
+            parsed = urlparse(paste)
+        if parsed.hostname is not None:
+            widget.set_text(parsed.hostname)
+            widget.emit_stop_by_name('paste-clipboard')
```
cas--: Is this required? If so it likely should be at the start or end of the method.

deluge/ui/gtkui/connectionmanager.py
```diff
@@ -233,6 +234,23 @@ def add_host(self, host, port, username='', password=''):
...
+        if parsed.port is not None:
+            self.builder.get_object('spinbutton_port').set_value(parsed.port)
+        if parsed.username is not None:
+            self.builder.get_object('entry_username').set_text(parsed.username)
+        if parsed.password is not None:
+            self.builder.get_object('entry_password').set_text(parsed.password)
```
cas--: Since you are looking for text that is either not an empty string or None, all the `is not None` should be removed.

deluge/ui/gtkui/connectionmanager.py
```diff
@@ -233,6 +234,23 @@ def add_host(self, host, port, username='', password=''):
         # Update the status of the hosts
         self.__update_list()
 
+    def on_entry_host_paste_clipboard(self, widget):
+        paste = gtk.clipboard_get().wait_for_text()
```
cas--: Can you rename paste to clipboard so it's a bit clearer what the var is.

cas--:
Just thought you should strip whitespace from the clipboard text.

I also wonder if it should automatically paste those values from the clipboard. A good example is for urls in add_torrent_dialog.

deluge/ui/gtkui/connectionmanager.py
matiasw: Well, it is the pasted text and not the clipboard returned by clipboard_get(), but ok.

deluge/ui/gtkui/connectionmanager.py
```diff
@@ -233,6 +234,23 @@ def add_host(self, host, port, username='', password=''):
...
+        log.debug('on_entry_proxy_host_paste-clipboard: got paste: %s', paste)
+        if '//' not in paste:
+            parsed = urlparse('//' + paste)
```
matiasw: If there is no scheme (like `http://`), the address needs an empty scheme (or "prefixing with slashes") so that eg. ?in `username:password@host` style addresses, `username` does not get confused with a scheme identifier.

deluge/ui/gtkui/connectionmanager.py
matiasw: If you mean the `emit_stop_by_name`, it is required strictly when we modify the entry field manually (to stop the clipboard contents from getting pasted at the end of the text we've set).

cas--: Ok I see

deluge/ui/gtkui/connectionmanager.py
cas--: you can just use `text`

deluge/ui/gtkui/connectionmanager.py
cas--: Ok so just do:

```
# [comment with explanation]
text = text if '//' in text else '//' + text
parsed_txt = urlparse(text)
```

deluge/ui/gtkui/connectionmanager.py
matiasw: I don't understand what you mean by this? Would you rather have `TypeError: Gtk.Entry.set_text() argument 1 must be string, not None` each time one of these is None?

cas--: `if parsed.username:` is false for None

matiasw:
Whitespace is now stripped.
Do you mean like, if, on inspection, the clipboard is found to contain suitable values, these would be pasted automatically? I think that's a separate feature that we could look into implementing.
Meanwhile, what should I do to get deluged working again? I got this 
> ImportError: Deluge 2.0.0.dev897 requires lastest github code from libtorrent RC_1_1 branch.
completely out of the blue.

cas--:
That is because the latest develop code requires the latest libtorrent git branch code. So you would need to checkout RC_1_1 and build that code or use the ubuntu ppa if you can.

deluge/ui/gtkui/preferences.py
```diff
@@ -978,6 +980,21 @@ def _on_combo_proxy_type_changed(self, widget):
             else:
                 self.builder.get_object(entry).hide()
 
+    def _on_entry_proxy_host_paste_clipboard(self, widget):
+        text = ''.join(gtk.clipboard_get().wait_for_text().split())
```
matiasw: `strip()` only gets whitespace from the beginning and end of a string.

matiasw:
Yeah, I have the libtorrent code from github. I've tried with the latest master, as well as the libtorrent-1_1_1 release tag. Or do you mean some other version?

deluge/ui/gtkui/preferences.py
cas--: We are primarily concerned with trailing newlines causing issues so start and end is fine.

cas--:
I have repeated many times:  git branch `RC_1_1` not master and not tag 1.1.1.

    git checkout RC_1_1

cas--:
Thanks d75afc007dcc5b

PR #139, opened 2017-01-07
[WebUI] Added X-Deluge-Session header support.

N3X15:
This gives external WebGUI API consumers the option to send the login session ID as an HTTP header (`X-Deluge-Session`), rather than as a cookie.

# Use Case
Firefox and Chrome extensions using the WebExtension API make use of XmlHttpRequest and fetch(), which are not allowed to get or set cookies.  While building an extension to send .torrents and magnet links to deluge, I discovered this limitation. In addition, the most obvious work around is with webRequests, which appear to not have been "hooked up" to XHRs and fetch() in Firefox, for unknown reasons.

Since it's not likely Google and Mozilla will be able to straighten out the standard fast enough, I made this quick patch to work around it, as HTTP headers (other than cookie-related ones) are modifiable with XHR and fetch().

# API changes
* auth.login returns `{success: bool(), session_token: string()}` rather than `true` on success.  
 * This is to permit aforementioned scripts to discover their session ID after login, since reading cookies is also blocked.
 * Does not affect webUI scripts, since the login logic just checks for null.
* Consumers may send `X-Deluge-Session` rather than the `_session_id` cookie, if desired.

# BONUS!
* Fixes a typo in setup.py.


cas--:
What was the reason for closing this?

N3X15:
Well, since it's been tried before and the behaviour described is discouraged, I decided it wasn't worth trying to make a duplicate PR and step on established security norms.

cas--:
Yeah fair enough, is there a way to workaround the issue though?

N3X15:
Running a proxy application, as someone suggested to me a while ago.

Something tells me a browser extension running a script or executable just to modify HTTP requests is a bit scary to some folk, though.

cas--:
You could create a simple plugin to override the methods. I'm just not sure about adding this to the code.

N3X15:
The problem is the session itself:  This messes with the core authentication methods to ensure that A. We're told what our session ID is (previously, we had to get it via cookies, which is hard if you don't have access to them) and B. That we can tell the webUI provider what that session is via means other than cookies.  

Sure, we could hack around the session login part, but what if I want to upload a torrent?  I'd have to reimplement that to include a session setter.  What if I want to upload a magnet? Same thing.  It'd basically involve reimplementing the entire webUI.  Then it all comes back to, "well, the WebUI is doing pretty much the same thing, why not just add in a tweak to the authentication system and call it a day?"

Another suggestion was to implement the deluge RPC protocol, but I don't see a decent rencoding library for Javascript, and I'm not smart enough to do it myself.

cas--:
Did you try using `with_credentials`, that is the issue you are facing?

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials
http://stackoverflow.com/a/37955635/175584

PR #143, opened 2017-02-16
Feature/unicode literal

cas--:


deluge/tests/test_client.py
```diff
@@ -119,7 +119,6 @@ def on_failure(failure):
                 error.BadLoginError
             )
             self.assertEquals(failure.value.message, 'Password does not match')
-            self.addCleanup(client.disconnect)
```
cas--: @bendikro was the disconnect from copy-paste of login ok tests?

cas--:
The `msgmft` module is currently only py 2.7.12 compatible so likely will require the following dependency: http://python-future.org/stdlib_incompatibilities.html#array-array

PR #144, opened 2017-03-08
added slimit to setup_requires

krichter722:
Added `slimit` to `setup_requires` in order to avoid `Import error: Requires slimit package for minifying WebUI files.`.

The [test failures](https://travis-ci.org/krichter722/deluge/jobs/209017439) seem to indicate a further problem which occurs after adding. It'd be good to get some feedback because the changes proposed here ease the local build with `python setup.py build && sudo python setup.py install` on Ubuntu 16.10.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -487,6 +487,7 @@ setup(
     url='http://deluge-torrent.org',
     license='GPLv3',
     cmdclass=cmdclass,
+    setup_requires=['slimit'],
     tests_require=['pytest'],
     data_files=_data_files,
     package_data=_package_data,


cas--:
I don't actually want slimit to be a dependency for setup. When shipping the release source code I intend to include minified js files so that it is not required. It is listed under DEPENDS for development and perhaps need to make that clearer.

cas--:
I think the better solution would be that if slimit is missing that non-minified js is copied over.

http://dev.deluge-torrent.org/ticket/3053#ticket

PR #148, opened 2017-03-17
Issue/plugin resources path

cas--:
This branch is attempting to deal with the non-ascii issue in [#2341](http://dev.deluge-torrent.org/ticket/2341) and unify the access to resources by plugins.

 - I've ensured pkg_resources.resource_filename filepath is decoded.
 - However I realised it seems unnecessary to use a `pkg_resources` lookup when the plugin `data` dir is a relative path and a call to `abspath` would resolve an absolute path.

- [x] Perhaps need to decode abspath to string...

There is another option to get the entire file as a buffer string using [`pkgutil.get_data`](https://docs.python.org/2/library/pkgutil.html#pkgutil.get_data) which can be used with `gtk.glade.xml_new_from_buffer` (instead of `gtk.glade.XML(filename)`)

### DIFF
--- a/deluge/plugins/Blocklist/deluge/plugins/blocklist/common.py
+++ b/deluge/plugins/Blocklist/deluge/plugins/blocklist/common.py
@@ -1,6 +1,11 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2009 Andrew Resch <andrewresch@gmail.com>
+# Basic plugin template created by:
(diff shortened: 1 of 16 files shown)


deluge/plugins/Blocklist/deluge/plugins/blocklist/common.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
```
cas--: Oops must fix this overwrite mistake

PR #151, opened 2017-03-30
Refactor the UI Connection Mangers

cas--:
Runaway refactoring snowball...

- [ ] Re-test all connection managers before merge
- [ ] Check for missing docstrings
- [x] Tested GTKUI connection manager
- [x] Fix tests
- [x] Add edit to webui (reuse add dialog)
- [x] Ensure console connection manager works
- [x] Ensure webui connection manager works
- [x] Consolidate the duplicated connecting and status code

### DIFF
--- a/deluge/tests/common_web.py
+++ b/deluge/tests/common_web.py
@@ -9,8 +9,6 @@
 
 from __future__ import unicode_literals
 
-from twisted.internet import reactor
-
 import deluge.common
 import deluge.component as component
 import deluge.ui.web.auth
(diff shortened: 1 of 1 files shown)


deluge/ui/console/modes/connectionmanager.py
```diff
@@ -174,28 +170,21 @@ def on_resize(self, rows, cols):
     def read_input(self):
         c = self.stdscr.getch()
 
-        if c > 31 and c < 256:
-            if chr(c) == 'q' and self.inlist:
-                return
+        if is_printable_chr(c):
```
bendikro: `is_printable_chr` tests for value `c >= 32 and c <= 126`, so this doesn't test for the same as the original code. I'm not sure what input values are > 126 though.

bendikro: Considering it only tests for four different values inside the if, this isn't an issue. Too early in the morning :-)

cas--: Yeah I don't think we will be using symbols as key shortcuts ;)

For reference this is the 256 ascii table: http://www.flexcomm.com/library/ASCII256.htm

PR #153, opened 2017-05-01
Update gitignore

nagracks:
Ignore .cache

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 build
+.cache
 dist
 docs/source/modules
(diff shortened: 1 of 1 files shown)


cas--:
What is `.cache` from?

nagracks:
When I tried to run tests with `py.test`, it generated that .cache/ dir. I thought it should be in .gitignore.

cas--:
Ah ok, I always use trial so didn't realise that :)

cas--:
Merged thanks

PR #155, opened 2017-05-11
Trac#3011 Handle WM_ENDSESSION on Windows to shutdown cleanly

cas--:
This is not ready for merge however it is working code and will catch the wm_queryendsession messages. The next step is to handle this correctly so Deluge is given enough time to save data to disc as the default 5 secs is probably not enough. 

See trac ticket for more details.

### DIFF
--- a/deluge/ui/gtkui/mainwindow.py
+++ b/deluge/ui/gtkui/mainwindow.py
@@ -109,6 +109,24 @@ class MainWindow(component.Component):
             while gtk.events_pending():
                 gtk.main_iteration(False)
             self.vpaned.set_position(self.initial_vpaned_position)
+        elif deluge.common.windows_check():
+            # Realize window to create a handle for wndproc.
+            self.window.realize()
+
+        if deluge.common.windows_check():
+            # Handle win32 WM_QUERYENDSESSION message so GUI can be shutdown cleanly.
+            from win32gui import CallWindowProc, SetWindowLong
+            from win32con import GWL_WNDPROC, WM_QUERYENDSESSION
+
+            def on_wndproc(hwnd, msg, wparam, lparam):
+                # log.debug("%s, %s, %s", msg, wParam, lParam)
+                if msg == WM_QUERYENDSESSION:
+                    reactor.stop()
+                # Pass all messages on to the original WndProc.
+                return CallWindowProc(old_wndproc, hwnd, msg, wparam, lparam)
+
+            # Set WndProc to above function and store old value.
+            old_wndproc = SetWindowLong(self.window.window.handle, GWL_WNDPROC, on_wndproc)
 
     def show(self):
         try:
--- a/deluge/ui/gtkui/mainwindow.py
+++ b/deluge/ui/gtkui/mainwindow.py
@@ -84,6 +84,10 @@ class MainWindow(component.Component):
         # UI when it is minimized.
         self.is_minimized = False
 
+        # Where to store the original WNDPROC value
+        self.win32_prev_wndproc = None
+        self.win32_shutdown_block = None
+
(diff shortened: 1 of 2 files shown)


DjLegolas:
I wrote a handler in ef6ca96.
On [`WM_QUERYENDSESSION`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376890(v=vs.85).aspx) message, Deluge will register itself as a Shutdown-Blocker.
Next, when receiving [`WM_ENDSESSION`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376889(v=vs.85).aspx) message, Deluge should have 30 seconds to save all the data - depending on whether critical shutdown or not.

But the lack of a proper tool for testing this makes me feel like the is something wrong with my work.

deluge/ui/gtkui/mainwindow.py
```diff
@@ -109,6 +109,24 @@ def first_show(self):
             while gtk.events_pending():
                 gtk.main_iteration(False)
             self.vpaned.set_position(self.initial_vpaned_position)
+        elif deluge.common.windows_check():
+            # Realize window to create a handle for wndproc.
+            self.window.realize()
+
+        if deluge.common.windows_check():
+            # Handle win32 WM_QUERYENDSESSION message so GUI can be shutdown cleanly.
```
cas--: Lets try and drop pywin32 module usage, so I'm thinking:

```python
from ctypes.wintypes import HWND, WPARAM, LPARAM, UINT, WINFUNCTYPE
from ctypes import windll, c_wchar_p
CallWindowProc = windll.user32.CallWindowProcW
SetWindowLong = windll.user32.SetWindowLongW

LRESULT = LPARAM
WNDPROC = WINFUNCTYPE(LRESULT, HWND, UINT, WPARAM, LPARAM)

WM_QUERYENDSESSION = 17
GWL_WNDPROC = -4
```

References:

- https://github.com/KellyZ/USB-Clipboard/blob/master/clipboard.py
- https://github.com/rene-aguirre/pywinusb/blob/master/pywinusb/hid/wnd_hook_mixin.py

deluge/ui/gtkui/mainwindow.py
```diff
@@ -109,6 +109,24 @@ def first_show(self):
...
+
+            def on_wndproc(hwnd, msg, wparam, lparam):
+                # log.debug("%s, %s, %s", msg, wParam, lParam)
+                if msg == WM_QUERYENDSESSION:
+                    reactor.stop()
+                # Pass all messages on to the original WndProc.
+                return CallWindowProc(old_wndproc, hwnd, msg, wparam, lparam)
+
+            # Set WndProc to above function and store old value.
+            old_wndproc = SetWindowLong(self.window.window.handle, GWL_WNDPROC, on_wndproc)
```
cas--: Need to fix `old_wndproc` scoping.

Additionally if dropping pywin32 need to wrap `on_wndproc` with above `WNDPROC` func.

DjLegolas: In what way to fix `old_wndproc` scoping?

DjLegolas:
So, I was trying to implement what you suggested, and now the GTK crashes...
The only errors I found related are 0xc0000005 (Access Violation) followed by 0xc000041d (An unhandled exception was encountered during a user callback)... Still not sure why.

cas--:
Oh hmm, probably some issue with setting up SetWindowLong. Don't worry about the pywin32 deprecation, it can wait, I posted it here as a reminder until everything is working.

deluge/ui/gtkui/mainwindow.py
cas--: Needs to be a class variable when you have a callback so i'd call it `self.win32_prev_wndproc`

Also it is best to define a class variable in __init__ and assign it None.

PR #156, opened 2017-05-24
Update Deluge icons with a cleaner look

cas--:
selection_053

The 16px icons are not as good as they could and will need revisiting to tweak the svg for the lower resolution.

- [x] Need to squash commits
- [x] Rename script to make it's purpose clearer
- [x] Update About logo
- [x] Update webui icons
- [ ] Monochrome deluge icon for ubuntu themes (22px):
    * ubuntu-mono-dark
    * ubuntu-mono-light


aresch:
You may want to remove the dashed line from the downloading/seeding arrows.. it will probably make the smaller versions look a bit cleaner.  I'd also suggest that since the icons are being changed, maybe we should consider using a colour palette that would work better for colour-blind people.  Not sure how that would look, but it may be appreciated by some of the population :)

cas--:
> remove the dashed line from the downloading/seeding arrows..

I do intend to make the smaller version cleaner either with fewer dashes or a solid line.

> consider using a colour palette that would work better for colour-blind people.

I am not sure if we need to go down that route as there is text alongside but I shall look into it :)

cas--:
I am merging this with the current changes as it's taking more time than I intended... 

Created a ticket to revisit issues: http://dev.deluge-torrent.org/ticket/3052

PR #157, opened 2017-06-05
lossless compression of images saved 457KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 457KB docent sound like much but why not


cas--:
Thanks for this but how can this be scripted? 

Also this is failing the tests because you compressed a comparison png.

Fish2:
I used AdvPNG and PNGOUT so you could probably use that but would have to add it to the build requirements which would then make the build bigger.

I have added a fix for the comparison test

cas--:
It is a dev tool not for inclusion with build. I see that `advpng` is available for Ubuntu with `apt install advancecomp` so we can use that. The current scripts under `deluge/scripts` can be extended to run over the pngs and compress them.

Fish2:
that sounds good to me

diogocp:
advpng is not very good. pngcrush is also available in Ubuntu and will probably give you much better results. See https://css-ig.net/png-test-corpus for a comparison of various tools.

Converting the GIFs to PNG should also help.

cas--:
pngs have been compressed with zopflipng: fe42fb2c31d87b013

PR #158, opened 2017-06-11
Save torrent state only if state has changed. 

ratoaq2:
Running deluge in a PC or NAS server prevents HDD to spin down, even if there is no activity, no torrents or whatsoever. This happens because deluge writes the torrent state to a file every time.

This PR will make sure that this file is only written if the torrent state has actually changed.

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -111,9 +111,21 @@ class TorrentState:
         self.move_completed = move_completed
         self.move_completed_path = move_completed_path
 
+    def __eq__(self, other):
+        return isinstance(other, TorrentState) and self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
+        return not self == other
+
 class TorrentManagerState:
     def __init__(self):
         self.torrents = []
+
+    def __eq__(self, other):
+        return isinstance(other, TorrentManagerState) and self.torrents == other.torrents
+
+    def __ne__(self, other):
+        return not self == other
 
 class TorrentManager(component.Component):
     """
@@ -156,6 +168,9 @@ class TorrentManager(component.Component):
 
         # Workaround to determine if TorrentAddedEvent is from state file
         self.session_started = False
+
+        # Keep the previous saved state
+        self.prev_saved_state = None
 
         # Register set functions
         self.config.register_set_function("max_connections_per_torrent",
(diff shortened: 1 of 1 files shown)


deluge/core/torrentmanager.py
```diff
@@ -157,6 +173,9 @@ def __init__(self):
         # Workaround to determine if TorrentAddedEvent is from state file
         self.session_started = False
 
+        # Keep the last saved state
+        self.last_saved_state = None
```
cas--: I'd prefer it called `prev_saved_state` as a loose consistency ;)

deluge/core/torrentmanager.py
```diff
@@ -111,10 +111,26 @@ def __init__(self,
         self.move_completed = move_completed
         self.move_completed_path = move_completed_path
 
+    def __eq__(self, other):
+        if not isinstance(other, TorrentState):
```
cas--: Would prefer a `try..except` here. It is more likely to be an instance of TorrentState than not, and bit more pythonic.

deluge/core/torrentmanager.py
```diff
@@ -111,10 +111,26 @@ def __init__(self,
...
+
+    def __ne__(self, other):
+        return not self == other
+
 class TorrentManagerState:
     def __init__(self):
         self.torrents = []
 
+    def __eq__(self, other):
+        if not isinstance(other, TorrentManagerState):
```
cas--: Same use `try..except` as above

deluge/core/torrentmanager.py
ratoaq2: I didn't fully get you here:

```
    def __eq__(self, other):
        try:
            return self.__dict__ == other.__dict__
        except:
            return False
```

In theory the following could still happen:

```
class A:
    def __init__(self):
        a = 'a'
    
class B:
    def __init__(self):
        a = 'a'

A().__dict__ == B().__dict__
True
```

cas--: Fair point. I do prefer a positive isinstance statement and you could do it in one line:

```python    
def __eq__(self, other):
    return isinstance(other, TorrentState) and self.__dict__ == other.__dict__
```

ratoaq2:
Done

cas--:
That's great thanks. One final thing, could you update the commit message with [Core] and move the second sentence out of the summary.

http://dev.deluge-torrent.org/wiki/GitTips#CommitMessages

ratoaq2:
Done

PR #159, opened 2017-06-11
Save torrent state only if state has changed.

ratoaq2:
Same as #158, but against develop branch.


cas--:
Thank you, merged: 60c764ac338

PR #160, opened 2017-06-14
[Core] Ensure state is written to disk on shutdown

cas--:


cas--:
Not sure really needed and may slow shutdown

PR #168, opened 2017-07-26
[GKTUI] Fix high priority files tab trigger

bendikro:


cas--:
Ah ignore appveyor, it's not setup yet but forgot to disable builds

cas--:
Merged

PR #169, opened 2017-08-10
Refactor/gtk builder plugins

DjLegolas:
Ticket  [#3002](http://dev.deluge-torrent.org/ticket/3002) - Convert Deluge Plugins to GtkBuilder

### DIFF
--- a/deluge/plugins/AutoAdd/deluge/plugins/autoadd/data/autoadd_options.ui
+++ b/deluge/plugins/AutoAdd/deluge/plugins/autoadd/data/autoadd_options.ui
@@ -1,33 +1,63 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<glade-interface>
-  <!-- interface-requires gtk+ 2.16 -->
+<interface>
+  <requires lib="gtk+" version="2.24"/>
   <!-- interface-naming-policy toplevel-contextual -->
-  <widget class="GtkDialog" id="options_dialog">
+  <object class="GtkAdjustment" id="adjustment1">
+    <property name="lower">-1</property>
+    <property name="upper">10000</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="adjustment2">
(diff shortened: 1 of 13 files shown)


cas--:
Thanks for this, I'll need to look at why the tests are failing.

DjLegolas:
I tested it on my system (configured as close as possible the one Travis's initializes) and all tests passed.

In addition, I started testing the AutoAdd changes, but there is an issue with the Label plugin in which the label selected is not being added to the torrent.

cas--:
So I have fixed and merged these changes. Thanks again :+1: 

Quick note that the change from ComboBoxEntry to ComboBox also needed a change to the method that was being used to set the item from `set_text_column` to `set_entry_text_column`.

The was another unrelated autoadd-label issue to do with changes to the core api, which I have fixed: 26720ca4c2

doadin:
Not 100% sure it was after this change but, after updating from ~ 7b87a93862bae5bcb53d4134a111a308ec36f9e8 to 26720ca4c2e96b1c4c2df145a3642e57d870652b the notifications plugin enables but has no tab label in the plugins list section in the Preferences window. After clicking Prefferences > Plugins and enabling notifications a tab is made but is blank. Clicking on the blank tab does bring up the options for notifications though.

DjLegolas:
I looked in the code and the reason Notifications behaves like this is due to the way it inserts itself to the list.
Every other plugin just calls to the [`add_page`](https://github.com/deluge-torrent/deluge/blob/develop/deluge/ui/gtkui/preferences.py#L194) function.
But Notifications doesn't. It just [adds itself](https://github.com/deluge-torrent/deluge/blob/develop/deluge/plugins/Notifications/deluge/plugins/notifications/gtkui.py#L266) to the list of items directly and does it partially. And, it's easy to fix this. Just need to replace lines 266 to 271 with the line:
```python
component.get('Preferences').add_page(_('Notifications'), self.prefs)
```
Regarding Blocklist... My bad, didn't see this line.
@cas-- Should I fix the error in Bloaklist in a new PR or in this one?

PR #170, opened 2017-09-14
Store state and fastresume in per-torrent files.

AllSeeingEyeTolledEweSew:
I was trying to load large nubmers of torrents in deluge (several hundred), and I noticed deluge was often unresponsive.

I noticed that when any fastresume data or state needs to change in deluge, it must serialize all state for all torrents (for fastresume data it must deserialize all the existing data first). It seems this regularly takes tens of seconds with several hundred torrents. The performance impact of this is especially drastic since the event loop is blocked while doing this single huge step.

I wrote a quick patch to split state and fastresume into multiple files, and I saw a huge increase in performance. I was able to load 10,000 torrents into deluge with not too many problems. I had to do all my work over RPC though, since the web UI fell into some unresponsive state I haven't investigated yet around 1,800 torrents.

I saw that deluge actually originally used multiple fastresume and state files, and was converted to single file mode in d28e5998b9f814933d5805daefc516b879c2ce98. I couldn't find any discussion about the motivations behind the change however. Multiple state files seems better for performance (given deluge's single threaded event loop model) and robustness against file corruption. I would love to know more about the history of this change though.

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -157,20 +157,11 @@ class TorrentManager(component.Component):
         # and that their resume data has been written.
         self.shutdown_torrent_pause_list = []
 
-        # self.num_resume_data used to save resume_data in bulk
-        self.num_resume_data = 0
-
         # Keep track of torrents finished but moving storage
         self.waiting_on_finish_moving = []
 
-        # Keeps track of resume data that needs to be saved to disk
-        self.resume_data = {}
-
         # Workaround to determine if TorrentAddedEvent is from state file
         self.session_started = False
-
-        # Keep the previous saved state
-        self.prev_saved_state = None
 
         # Register set functions
         self.config.register_set_function("max_connections_per_torrent",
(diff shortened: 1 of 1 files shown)


cas--:
Yeah I have no idea why it was switched to a single file either and have been mulling this change to mitigate data loss and few reads/writes. I would not want to put this change into `1.3-stable` but more than welcome for it to go into `develop` for 2.0.

aresch:
Sorry, I meant to reply to this when I first saw it.

This was a long time ago, but I think the reason to make it a single file was the same as you have now splitting it up -- performance.  I think writing a lot of small files (especially with very large sessions) could cause some slow downs, but it could have just been the way it was done at the time.  I could see a slow disk causing some issues.  For what it's worth, my original implementation was single files and my intention was to allow updating individual fast resume data without having to write out a large file each time.  *shrug*

I'm ok with changing it back to separate files, but I think cas is right that we should probably stage that for 2.0.

AllSeeingEyeTolledEweSew:
I'm a little concerned that 2.0 is ... very far away. I found this bug of *7 years* slated to be fixed in 2.0: http://dev.deluge-torrent.org/ticket/1466

Is there any way I could modify this change so it's more acceptable for 1.3?

cas--:
I am working on this currently so will be in a 2.0 release

cas--:
There is nothing new going in to 1.3-stable only critical bugfixes. Help with getting 2.0 to release status is always appreciated, the problem has been that too many features were added to develop and squashing all the resulting issues takes time.

AllSeeingEyeTolledEweSew:
Closing then.

PR #175, opened 2018-01-09
Call the blocking wait_for_alert in a thread.

AllSeeingEyeTolledEweSew:
This spawns a thread in alertmanager to call `wait_for_alert` in a thread. This reduces latency to deluge responding to events.

To quantify performance, I created a plugin to expose access to `read_piece` and `read_piece_alert` as an event, then measured the delta between the time a client received a response to the RPC, and the time the client received the event. I saw this time fall from ~140ms to ~50ms with  this change.

This change was recommended by Arvid Norberg, maintainer of libtorrent. See here: https://github.com/arvidn/libtorrent/pull/1318#issuecomment-260192822

### DIFF
--- a/deluge/core/alertmanager.py
+++ b/deluge/core/alertmanager.py
@@ -41,7 +41,10 @@ This should typically only be used by the Core.  Plugins should utilize the
 
 """
 
-from twisted.internet import reactor
+import threading
+import time
+
+from twisted.internet import reactor, threads
 
 import deluge.component as component
 from deluge._libtorrent import lt
@@ -51,7 +54,7 @@ from deluge.log import LOG as log
 class AlertManager(component.Component):
     def __init__(self):
         log.debug("AlertManager initialized..")
-        component.Component.__init__(self, "AlertManager", interval=0.3)
+        component.Component.__init__(self, "AlertManager")
         self.session = component.get("Core").session
 
         self.session.set_alert_mask(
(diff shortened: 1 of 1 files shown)


deluge/core/alertmanager.py
```diff
@@ -72,11 +102,17 @@ def update(self):
         self.delayed_calls = [dc for dc in self.delayed_calls if dc.active()]
         self.handle_alerts()
```
aresch: You'll want to remove this line.  Currently the update() method is called every ~0.3 seconds (see line 84) by the Component system.

deluge/core/alertmanager.py
```diff
@@ -41,13 +41,43 @@
...
 
+class AlertPoller(object):
+
+    def __init__(self, alert_manager):
+        self.alert_manager = alert_manager
+        self.running = False
+
+    def maybeUpdate(self):
+        if self.running:
+            self.alert_manager.update()
```
aresch: Change this to call self.alert_manager.handle_alerts() directly.

deluge/core/alertmanager.py
```diff
@@ -41,13 +41,43 @@
...
+
+    def run(self):
+        while self.running:
+            try:
+                alert = self.alert_manager.session.wait_for_alert(1000)
+            except:
```
aresch: You should probably at least log the exception here since it should rarely (if ever) occur and it would be good for debugging issues.

deluge/core/alertmanager.py
```diff
@@ -41,13 +41,43 @@
...
+import threading
+
+from twisted.internet import reactor, threads
 
 import deluge.component as component
 from deluge._libtorrent import lt
 
 from deluge.log import LOG as log
 
+class AlertPoller(object):
```
aresch: Typically you would subclass threading.Thread and override run() for something like this.  However, I think that you could probably even forego the class entirely and just run a thread of a method in AlertManager itself.  You could probably use the state of the Component class in lieu of self.running. I would just write a new method called 'wait_for_alerts' (essentially your run() method) and run that as a thread.  It will have access to all the member variables of AlertManager and you don't need to worry about synchronization since Python has the GIL.

Here is an example of just running a method as a thread:

```python
import threading


class AlertManager:
    def __init__(self):
        self.foo = 'bar'

    def thread(self):
        if self.foo == 'bar':
            self.foo = 'foobar'

    def start(self):
        print(self.foo)
        t = threading.Thread(target=self.thread)
        t.start()
        t.join()
        print(self.foo)

a = AlertManager()
a.start()
```

AllSeeingEyeTolledEweSew: Done.

deluge/core/alertmanager.py
AllSeeingEyeTolledEweSew: Done.

deluge/core/alertmanager.py
AllSeeingEyeTolledEweSew: I think I already do this.

deluge/core/alertmanager.py
AllSeeingEyeTolledEweSew: I'm a little paranoid about this, since I'm not sure I've caught all the cases of the component lifecycle.

AllSeeingEyeTolledEweSew:
I thought of a bad case that happens when AlertManager is paused. See the comments.

If there's a better way to handle this in Twisted, please let me know.

cas--:
Please use the develop code, 1.3-stable will not see any new features.

AllSeeingEyeTolledEweSew:

cas--:
Closing in favour of #221

PR #176, opened 2018-01-09
Discussion: some enhancements to the RPC protocol.

AllSeeingEyeTolledEweSew:
I don't expect this PR to be accepted as a single change, but I introduce it to start some discussion.

There are two major changes:
1. Do zlib/rencode work for outgoing messages in a thread
2. Introduce explicit framing around messages in the protocol, rather than relying on zlib for implicit framing. Introduce simple protocol version negotiation to accomplish this.

Number 1 is pretty straightforward, and is very useful for large messages, such as status on a large number of torrents. I've also been working with a plugin giving RPC access to `read_piece` and emitting `read_piece_alert` as an event. These can be quite large, and compression can block the event loop for seconds. I've taken care to ensure messages are compressed via a queue, so they are sent in the order they were generated. In particular it's important to ensure events are emitted in a consistent order.

Number 2 requires some justification. Today the Deluge RPC relies on zlib decompression to detect a message boundary, but I have found correctness and performance problems with this. I have found there are messages for which `zlib.decompress(buf) == zlib.decompress(buf[:-1])`. That is, sometimes a compressed buffer may contain an extraneous byte which may be omitted. So if a client gets `buf[:-1]` via a socket, it will successfully decompress, and the extraneous byte will corrupt the stream. This only happens when exchanging a very large amount of data. I haven't nailed down the exact circumstances in which this happens, but it happens enough that I had to create a client which tries ignoring bytes when trying to read a message (see https://github.com/AllSeeingEyeTolledEweSew/deluge_client_sync/blob/master/deluge_client_sync.py#L247) and I see it successfully log doing this at least a few times a day.

More than that, the fact remains that we must try to decompress a growing buffer after every `recv()`. So the time to decompress a message is `O(n^2)` in the length of a message, and this ends up being tens of milliseconds for messages of a few megabytes, or hundreds of milliseconds with the corruption detection described above.

Introducing an explicit length around each message solves both these problems. I chose a 4-byte unsigned int length encoded in little endian.

To negotiate which protocol to use, I added a `protocol_version` parameter to `daemon.login`. I defined the protocol such that the stream will operate in v1 until the response to `daemon.login` is received, and thereafter use the negotiated protocol in both directions. (One reason for including two RPC-related changes in this PR is to demonstrate how they interact here; the response to `daemon.login` must be sent synchronously before the protocol changes). A client passing `protocol_version` should detect a `TypeError` response as a server which only knows v1. It was necessary to piggyback on an existing RPC method, since Deluge yields no response to unknown methods; if we introduce a new method we'd have to wait for a timeout when trying it on an old server.

### DIFF
--- a/deluge/core/rpcserver.py
+++ b/deluge/core/rpcserver.py
@@ -41,10 +41,11 @@ import sys
 import zlib
 import os
 import stat
+import struct
 import traceback
 
 from twisted.internet.protocol import Factory, Protocol
-from twisted.internet import ssl, reactor, defer
+from twisted.internet import ssl, reactor, defer, threads
 
 from OpenSSL import crypto, SSL
 from types import FunctionType
@@ -63,6 +64,9 @@ from deluge.core.authmanager import AUTH_LEVEL_NONE, AUTH_LEVEL_DEFAULT
 RPC_RESPONSE = 1
 RPC_ERROR = 2
 RPC_EVENT = 3
+
+DEFAULT_PROTOCOL_VERSION = 1
+PROTOCOL_VERSION = 2
 
 def export(auth_level=AUTH_LEVEL_DEFAULT):
     """
@@ -124,6 +128,9 @@ class DelugeError(Exception):
 class NotAuthorizedError(DelugeError):
     pass
 
+class InvalidProtocolVersionError(DelugeError):
+    pass
+
 class ServerContextFactory(object):
     def getContext(self):
         """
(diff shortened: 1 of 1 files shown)


cas--:
Have you looked at the develop or master code? 1.3 will not see any changes.

aresch:
Yes, please make sure you look at the develop branch.  You'll notice that we already have your second suggestion implemented. If you want to re-sync your first suggestion against develop we can discuss using it but this PR as it stands will be closed.

AllSeeingEyeTolledEweSew:

PR #177, opened 2018-01-20
Save copies of torrent files from magnet links.

cykerway:
This patch is meant for 1.3-stable (1.3.15 as of the commit date).

The bug is: when you add a new magnet link to download, even if you tick
the option *Copy of .torrent files to:*, deluge still fails to save a
copy of torrent. This patch adds extra code to do so.

Bug discovered and patch tested on a Ubuntu 16.04 machine with deluge downloaded from official PPA.


cas--:
Seems I forgot to merge this! Now merged: e050905b291

PR #178, opened 2018-01-31
french typo

flip-bs:
Hi,
The right term in french for `encrypt` is `chiffrer` 
https://chiffrer.info/


cas--:
As per deluge docs we have use launchpad: https://translations.launchpad.net/deluge/1.3/+pots/deluge/fr/+translate

However I have made the changes and will be in the next release.

PR #180, opened 2018-05-02
#3171 Add Option To Specify Outgoing Connection Interface

doadin:
Currently if you specify -i only the libtorrent listen_interfaces is set.

I believe one of two options should be used:

-i sets listen_interfaces and outgoing_interfaces.

or

-i sets listen_interface and a new -o sets outgoing_interfaces.

Im sure for some reason or another someone would want to use two interfaces so the second option is probably best.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -101,7 +101,7 @@ DELUGE_VER = deluge.common.get_version()
 
 
 class Core(component.Component):
-    def __init__(self, listen_interface=None, read_only_config_keys=None):
+    def __init__(self, listen_interface=None, outgoing_interface=None, read_only_config_keys=None):
         component.Component.__init__(self, 'Core')
 
         # Start the libtorrent session.
@@ -163,6 +163,13 @@ class Core(component.Component):
                 self.config['listen_interface'] = listen_interface
             else:
                 log.error('Invalid listen interface (must be IP Address): %s', listen_interface)
+        self.__old_outgoing_interface = None
+        if outgoing_interface:
+            if deluge.common.is_ip(outgoing_interface):
+                self.__old_outgoing_interface = self.config['outgoing_interface']
+                self.config['outgoing_interface'] = outgoing_interface
+            else:
+                log.error('Invalid outgoing interface (must be IP Address): %s', outgoing_interface)
 
         # New release check information
         self.__new_release = None
@@ -196,6 +203,9 @@ class Core(component.Component):
         # We stored a copy of the old interface value
         if self.__old_interface:
             self.config['listen_interface'] = self.__old_interface
+
+        if self.__old_outgoing_interface:
+            self.config['outgoing_interface'] = self.__old_outgoing_interface
 
         # Make sure the config file has been saved
         self.config.save()
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -47,6 +47,7 @@ DEFAULT_PREFS = {
     'download_location': deluge.common.get_default_download_dir(),
     'listen_ports': [6881, 6891],
     'listen_interface': '',
+    'outgoing_interface': '',
     'random_port': True,
     'listen_random_port': None,
     'listen_use_sys_port': False,
@@ -187,6 +188,9 @@ class PreferencesManager(component.Component):
     def _on_set_listen_interface(self, key, value):
         self.__set_listen_on()
 
+    def _on_set_outgoing_interface(self, key, value):
+        self.__set_outgoing_on()
+
     def _on_set_random_port(self, key, value):
         self.__set_listen_on()
 
@@ -209,6 +213,13 @@ class PreferencesManager(component.Component):
         self.core.apply_session_settings(
             {'listen_system_port_fallback': self.config['listen_use_sys_port'],
              'listen_interfaces': ''.join(interfaces)})
+
+    def __set_outgoing_on(self):
+        """ Set the interface address for outgoing BitTorrent connections."""
+        outinterface = self.config['outgoing_interface'].strip()
+        outinterface = outinterface if outinterface else '0.0.0.0'
+        self.core.apply_session_settings(
+            {'outgoing_interfaces': outinterface})
 
     def _on_set_outgoing_ports(self, key, value):
         self.__set_outgoing_ports()
--- a/deluge/ui/console/modes/preferences/preference_panes.py
+++ b/deluge/ui/console/modes/preferences/preference_panes.py
@@ -83,6 +83,10 @@ class BasePreferencePane(BaseInputPane, BaseWindow, PopupsHandler):
                     interface = ipt.get_value().strip()
                     if is_ip(interface) or not interface:
                         conf_dict['listen_interface'] = interface
+                elif ipt.name == 'outgoing_interface':
+                    outinterface = ipt.get_value().strip()
+                    if is_ip(outinterface) or not outinterface:
+                        conf_dict['outgoing_interface'] = outinterface
                 elif ipt.name.startswith('proxy_'):
                     if ipt.name == 'proxy_type':
                         conf_dict.setdefault('proxy', {})['type'] = ipt.get_value()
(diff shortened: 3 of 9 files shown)


deluge/core/core.py
```diff
@@ -163,6 +163,12 @@ def __init__(self, listen_interface=None, read_only_config_keys=None):
                 self.config['listen_interface'] = listen_interface
             else:
                 log.error('Invalid listen interface (must be IP Address): %s', listen_interface)
+        if outgoing_interface:
+            if deluge.common.is_ip(outgoing_interface):
+                self.__old_interface = self.config['outgoing_interface']
```
aresch: This overrides the listen_interface, so it either needs to be renamed (and logic added below to restore the config value) or removed.

deluge/core/preferencesmanager.py
```diff
@@ -203,6 +204,9 @@ def __set_listen_on(self):
         interface = str(self.config['listen_interface'].strip())
         interface = interface if interface else '0.0.0.0'
 
+        outinterface = str(self.config['outgoing_interface'].strip())
```
doadin: This should maybe be put in its own def.

doadin:
This should be ready for merge now. Unless theres something missing im not aware of it runs and sets the proper setting in GTKUI and WebUI.

deluge/core/core.py
```diff
@@ -101,7 +101,7 @@
 
 
 class Core(component.Component):
-    def __init__(self, listen_interface=None, read_only_config_keys=None):
+    def __init__(self, listen_interface=None, outgoing_interfaces=None, read_only_config_keys=None):
```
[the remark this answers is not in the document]
doadin: I don't know how long it has been like this but it is multiple just like listen_interfaces/listen_interface thats kinda why i was a little confused as I was seeing a lot of "listen_interface" and not "listen_interfaces" when I was looking how that was implemented. From the online documents for libtorrent "a comma-separated list of IP port-pairs.". However at least more recently I believe this can be a interface eg. eth0,wlan0... On linux or unix type OSs. However in deluge you have it set to verify against a single ip and no interface so a rework would be needed to support those.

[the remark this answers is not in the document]
doadin: so something like https://github.com/deluge-torrent/deluge/pull/180/commits/438b2fa40b34bc3e357f6e3cf94b866c257b24d9 is what I was thinking however this still doesn't cover network interfaces but does fix checking ips that may or may not have more than one or an ipv6 address with "[]" which we also have to separate out to get the ip separate from the port. If that made sense...lol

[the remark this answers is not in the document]
cas--: Can we concentrate on just a single outgoing interface for now? We can add code to support multiple interfaces and interface name in future PRs. 

Keeping a PR simple will help it get merged sooner. 

Also can you add a changelog entry

deluge/ui/console/modes/preferences/preference_panes.py
```diff
@@ -251,7 +251,7 @@ def create_pane(self, core_conf, console_config):
 
         self.add_header(_('Outgoing Interface'), space_above=True)
         self.add_text_input('outgoing_interfaces', '%s:' % _('IP address of the interface to make connections out on '
-                                                          '(leave empty for default)'),
+                                                             '(leave empty for default)'),
```
cas--: I'd prefer this change:

```diff
-        self.add_text_input('listen_interface', '%s:' % _('IP address of the interface to listen on '
-                                                          '(leave empty for default)'),
-                            core_conf['listen_interface'])
+        self.add_text_input(
+            'listen_interface',
+            _('IP address of the interface to listen on (leave empty for default):'),
+            core_conf['listen_interface'],
+        )
```

The translation string was incorrect since the colon could not be moved by the translator and there is no need for substitution as it is a fixed string.

deluge/ui/gtkui/filtertreeview.py
```diff
@@ -335,7 +335,7 @@ def set_menu_sensitivity(self):
             item.set_sensitive(sensitive)
 
     def select_all(self):
-        'for use in popup menu'
+        # for use in popup menu
```
cas--: This should be a docstring

doadin: Yup sorry meant to make same change as above...whoever added these bits of code didn't flake8 shoulda realized that but really was wanting passing tests to see when i broke things :P

deluge/core/core.py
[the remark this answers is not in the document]
doadin: sure. I was a bit confused before I was like how is this setting getting applied if the libtorrent setting is "listen_interfaces" but every where in deluge its 'listen_interface" but i see it now deluge\core\preferencesmanager.py>__set_list_on has apply_session_settings "listen_interfaces". *facepalm*

deluge/ui/console/modes/preferences/preference_panes.py
```diff
@@ -241,9 +245,18 @@ def create_pane(self, core_conf, console_config):
         self.outto.set_depend(outrand, inverse=True)
 
         self.add_header(_('Interface'), space_above=True)
```
doadin: should this really be listen interface maybe? since "interface" setting is for the daemon interface to listen on.

doadin:
I think I have addressed all the issues now?

deluge/ui/console/modes/preferences/preference_panes.py
cas--: The daemon setting is rarely something a user will specify. However, if you have added Outgoing Interface then yes changing this to Incoming Interface .

cas--:
Yeah I shall try to fix those failing tests and you can rebase when it's pushed

doadin:
Yea, so far I can see the changes so far does affect travis.ci test but should be some kinda of solution to make them more broadly compatible. GL.  :)

deluge/ui/console/modes/preferences/preference_panes.py
doadin: This would also make the heading match gtkui and webui. So it will all look more uniform.

doadin:
Is this ready to go now?

asavah:
Was this runtime tested? I don't think so ...

doadin:
It was runtime tested I use this build and it runs just fine. question look at the error...it says invalid interface what did you use for a interface? currently only a ip is supported. as deluge runs the is_ip check on the input.

asavah:
I ve tried no ip at all (blank field).
Tried local ip of the server.
Same result.
The error itself comes from libtorrent (I'm using 1.1.7).
the peer list is empty for all the torrents, reverting this PR makes everything work again.

Currently I'm trying to figure out what's wrong, will submit a fix if I find one.

And I'm not the only one https://forum.deluge-torrent.org/viewtopic.php?f=8&t=54970

doadin:
I just tested on windows and it runs just fine with no option selected or with a interface ip selected, maybge this is a linux issue?

asavah:
Not sure yet,
 yes, I'm running on linux (ubuntu 18.04 server), trying to dig through libtorrent api , c++ is not my favorite language ...

asavah:
https://libtorrent.org/single-page-ref.html


name | type | default
-- | -- | --
outgoing_interfaces | string | ""

sets the network interface this session will use when it opens outgoing connections. By default, it binds outgoing connections to INADDR_ANY and port 0 (i.e. let the OS decide). **Ths parameter must be a string containing one or more, comma separated, adapter names. Adapter names on unix systems are of the form "eth0", "eth1", "tun0", etc. When specifying multiple interfaces, they will be assigned in round-robin order. This may be useful for clients that are multi-homed**. Binding an outgoing connection to a local IP does not necessarily make the connection via the associated NIC/Adapter. Setting this to an empty string will disable binding of outgoing connections.

It should be interface name not ip address, may be windows resolves it somehow ...

asavah:
Dirty (not proper) fix:
https://github.com/deluge-torrent/deluge/pull/180/files#diff-fb7ff7c24ea09f1000011185039b4219R220
change '0.0.0.0' to ''

outgoing_interfaces is handled different to listen_interface in libtorrent itself.

PR #181, opened 2018-05-18
Initial Appveyor Support

doadin:
Goal was same tests as travis.ci. Outcome here is that while most tests run with this setup only flake8 runs 100% as it should(produces same output as travis).

However while the other tests run the tests need to be adjusted/modified to be windows compatible. For example in one test the ui.console gets tested which uses python curses... python curses doesn't exist so the test runs but fails. In "wintrial" (skips=1, failures=7, errors=41, successes=115).

Currently even on travis.ci "docs" is failing and the only addition fail comes from the curses issue which that "test" should probably just be ignored/skiped on windows since it will never work.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,128 @@
+environment:
+
+  matrix:
+    # See: http://www.appveyor.com/docs/installed-software#python
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: pydef
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: flake8
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: docs
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: trial
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: pygtkui
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      TOXENV: plugins
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
+
+pull_requests:
+  do_not_increment_build_number: true
+
+install:
+  # If there is a newer build queued for the same PR, cancel this one.
+  # The AppVeyor 'rollout builds' option is supposed to serve the same
+  # purpose but it is problematic because it tends to cancel builds pushed
+  # directly to master instead of just PR builds (or the converse).
+  # credits: JuliaLang developers.
+  - ps: if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER -ne ((Invoke-RestMethod `
+        https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
+        Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
+          throw "There are newer queued builds for this pull request, failing early." }
+
+  - "python -m pip install --upgrade pip"
+  - if NOT DEFINED TOXENV (
+      pip install bbfreeze pefile tox pywin32 slimit twisted[tls] chardet mako pyxdg pillow slimit pygame
+    )
+
+  - if DEFINED TOXENV (
+      pip install tox
+    )
+
+  - if not exist pygtk-all-in-one-2.24.2.win32-py2.7.msi (
+      ECHO "Downloading pygtk...."
+      & appveyor-retry appveyor DownloadFile "https://ftp.gnome.org/pub/GNOME/binaries/win32/pygtk/2.24/pygtk-all-in-one-2.24.2.win32-py2.7.msi"
+    )
+
+  - ECHO "Installing pygtk...."
+  - cmd: msiexec /i pygtk-all-in-one-2.24.2.win32-py2.7.msi /quiet /qn /norestart TARGETDIR=C:\Python27 INSTALLLEVEL=3
+  - if not exist openssl-1.1.0f-vs2008.7z (
+      ECHO "Downloading openssl...."
+      & appveyor-retry appveyor DownloadFile "https://www.npcglib.org/~stathis/downloads/openssl-1.1.0f-vs2008.7z"
+    )
+
+  - ECHO "Installing openssl...."
+  - "7z x -oc:\\ -aoa openssl-1.1.0f-vs2008.7z"
+  - "rename c:\\openssl-1.1.0f-vs2008 openssl-1.1"
+  - "copy c:\\openssl-1.1\\bin\\libsslMD.dll c:\\openssl-1.1\\bin\\libssl-1_1.dll"
+  - "copy c:\\openssl-1.1\\bin\\libcryptoMD.dll c:\\openssl-1.1\\bin\\libcrypto-1_1.dll"
+
+  - if not exist libtorrent.pyd (
+      ECHO "Downloading libtorrent...."
+      & appveyor-retry appveyor DownloadFile "https://github.com/doadin/libtorrent/releases/download/1.1.7.test/libtorrent.pyd"
+    )
+
+  - ECHO "Installing libtorrent...."
+  - "copy /Y libtorrent.pyd c:\\Python27\\Lib\\site-packages\\libtorrent.pyd"
+  - "SET PATH=%TOXENV%;%PYTHON%;%PYTHON%\\Scripts;c:\\openssl-1.1\\bin;C:\\Program Files (x86)\\NSIS;%PATH%"
+
+  - ECHO "Python Verison"
+  - "python --version"
+  - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
+  - ECHO "libtorrent Verison"
+  - "python -c \"import libtorrent; print(libtorrent.__version__)\""
+  - ECHO "OpenSSL Verison"
+  - openssl version -v
+
+cache:
+  - pygtk-all-in-one-2.24.2.win32-py2.7.msi
+  - openssl-1.1.0f-vs2008.7z
+  - '%LOCALAPPDATA%\pip\cache'
+  - libtorrent.pyd
+
+build: false
+
+test_script:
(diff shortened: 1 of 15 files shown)


doadin:
The existing environments don't work as is in windows. This is a large part due to the TOXENV not being set which I was having some issue with appveyor and I didn't try adding it to the existing environments because I have been busy trying to get the tests running and so forth and didnt want to break them for travis.ci as im using both to compare to see proper output. Now that its working I can clean it up a bit more and make some adjustments.

However in at least the one case if you look you have a instance of the environment variable PWD this variable is not available in windows so that one is not going to work even with a appveyor setup adjustment.

tox.ini
```diff
@@ -59,12 +73,29 @@ commands =
...
+[testenv:wintrial]
+changedir = {toxinidir}
+setenv =
+    TOXENV=wintrial
+commands =
+    python -c "import libtorrent as lt; print lt.__version__"
+    python -m twisted.trial deluge.tests
+
 [testenv:plugins]
 setenv = PYTHONPATH = {env:PWD}:{env:PWD}/deluge/plugins
```
doadin: here wont work in windows

cas--: See fix here: https://github.com/deluge-torrent/deluge/pull/182/commits/3cbbd6186ee0b9fd58c6590917e3353623298299

doadin: ok, ill re-run all the tests on windows and modify the commit as needed. ty.

tox.ini
```diff
@@ -18,7 +18,7 @@ minversion=1.8
 [testenv]
 install_command = pip install --ignore-installed {opts} {packages}
 passenv = DISPLAY PYTHONPATH
-setenv = PYTHONPATH = {toxinidir}:
+setenv = PYTHONPATH = {toxinidir}
```
doadin: apparently this breaks the deps install on appveyor. good news is it seems its not needed anyways since test still run just fine on travis.ci.

doadin:
I think we still need a few different environments for windows because windows has a few extra deps then linux does.

cas--:
For different env platforms see: https://tox.readthedocs.io/en/latest/example/platform.html

tox.ini
cas--: The colon is os specific: https://tox.readthedocs.io/en/latest/config.html#globally-available-substitutions

But I agree there is no need for it here.

tox.ini
```diff
@@ -57,7 +57,7 @@ commands = pytest -v --basetemp=_pytest_temp -s -m "todo" deluge/tests
 setenv = {[testenv]setenv}:{toxinidir}/deluge/tests
 commands =
     python -c "import libtorrent as lt; print lt.__version__"
-    python -m twisted.trial --reporter=deluge-reporter deluge.tests
+    python -m twisted.trial deluge.tests
```
cas--: Why is the deluge-reporter being removed?

deluge/tests/common.py
```diff
@@ -236,7 +236,10 @@ def start_core(listen_port=58846, logfile=None, timeout=10, timeout_msg=None,
 import sys
 import deluge.core.daemon_entry
 
-sys.argv.extend(['-d', '-c', '%s', '-L', 'info', '-p', '%d'])
+    if sys.platform == 'win32':
```
cas--: Please use `windows_check()` from deluge.common

deluge/tests/test_ui_entry.py
```diff
@@ -31,6 +28,13 @@
 from .basetest import BaseTestCase
 from .daemon_base import DaemonBase
 
+try:
```
cas--: Use a windows check to exclude importing instead of a try..except. We want the test to error if there are any issues importing these modules.

tox.ini
doadin: It stops the test from being run at all. The setup goes fine but then at the first command fails. says something about only passing report to something...I don't remember at the moment.

cas--:
@doadin Reason for closing?

doadin:
Its a ways away from being ready, unless you want to take just the appveyor config or take it + skipping ~40 tests.

cas--:
I am happy with skipping the tests is fine for now, config needs a bit of a polish but it seems to be working and getting windows tests and builds from appveyor would be great.

doadin:
Well it also needs some tox changes for multi os like we talked about before which for some reason I cant figure out I had it run on my computer pushed the change to github and it failed..not sure.

doadin:
As you may have noticed to I was using a libtorrent build from my own we need to find a good reliable source to install that form as the last python build from the libtorrent repo is not good.

doadin:
ok I tried to clean it up a bit more I will leave it be for now let me know any adjustments or w.e. you would like to make.

doadin:
ok so after that gotta clean up the sommits and stuff but all the tests run and pass we get PASSED (skips=69, successes=116) which is alot of skips of which about 40 actually need addressing at a later point the others are never gonna pass because of the console ui dependent on python curses. Other than cleaning the commits and stuff make the tox multi os compatible I think i pretty well cleaned the appveyor config but let me know.

tox.ini
```diff
@@ -59,20 +59,20 @@ commands = pytest -v --basetemp=_pytest_temp -s -m "gtkui" deluge/tests
 commands = pytest -v --basetemp=_pytest_temp -s -m "todo" deluge/tests
 
 [testenv:trial]
-setenv = {[testenv]setenv}:{toxinidir}/deluge/tests
+setenv = {[testenv]setenv};{toxinidir}/deluge/tests
```
gazpachoking: Should the separator be `{:}` to be cross platform compatible? https://tox.readthedocs.io/en/latest/config.html#globally-available-substitutions

I've barely followed this PR, so please ignore me if I'm off base.

doadin: Im not sure in general about the cross platform of the whole tox file still have to fix the depends too. Have not really tried a whole lot so I dont know much either though @cas-- mentioned about the separator earlier. Any thoughts on it much appreciated.

doadin:
I guess maybe we should use appveyor-retry? to try and download the deps in case for whatever reason it fails to download. They should be cached so it should be rare to even need to download but if it does.

appveyor.yml
```diff
@@ -0,0 +1,128 @@
+environment:
+
+  matrix:
+    # See: http://www.appveyor.com/docs/installed-software#python
+
```
cas--: This matrix section can be improved by removing all the duplication, see: https://packaging.python.org/guides/supporting-windows-using-appveyor/

cas--: I *think* you can set Python version outside matrix for all matrix jobs e.g. 

```
environment:
  PYTHON_VERSION: "2.7"
  matrix:
    TOXENV: pydef
...
```

deluge/tests/test_client.py
```diff
@@ -75,6 +75,9 @@ def __on_disconnect(self):
 
 class ClientTestCase(BaseTestCase, DaemonBase):
 
+    if windows_check:
+        skip = 'windows cant start_core not enough arguments for format string'
```
cas--: I just realised we should use the unittest `skipIf`: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

deluge/tests/test_common.py
```diff
@@ -77,6 +77,8 @@ def test_is_infohash(self):
         self.assertTrue(is_infohash('2dc5d0e71a66fe69649a640d39cb00a259704973'))
 
     def test_get_path_size(self):
+        if windows_check():
+            raise unittest.SkipTest('os devnull is different on windows')
         self.assertTrue(get_path_size(os.devnull) == 0)
```
cas--: So os.devnull is cross-platform so perhaps `get_path_size` needs fixed rather than skip this test

deluge/tests/test_files_tab.py
```diff
@@ -96,6 +97,8 @@ def test_files_tab(self):
         self.assertTrue(ret)
 
     def test_files_tab2(self):
+        if windows_check():
+            raise unittest.SkipTest('on windows \ != / for path names')
```
cas--: A possible issue with `files_list`!

However, inside Python Windows should handle this so perhaps the test needs tweaked.

deluge/tests/test_tracker_icons.py
```diff
@@ -24,6 +25,9 @@
 @pytest.mark.internet
 class TrackerIconsTestCase(BaseTestCase):
 
+    if windows_check():
+        skip = 'cannot use os.path.samefile to compair on windows(unix only)'
+
```
cas--: Since this is in the tests, a more generic version would do: https://stackoverflow.com/a/8894488/175584

tox.ini
```diff
@@ -31,6 +35,7 @@ deps =
     mock
     slimit
     pillow
+    pywin32
```
cas--: should this not be prefixed with `mywindows:`?

appveyor.yml
doadin: for now sure since we are only python 2 compatible but in the future adding 3.x python testing it would have to be non-global. changing now later it doesn't make a difference too much.

deluge/tests/common.py
```diff
@@ -236,7 +236,12 @@ def start_core(listen_port=58846, logfile=None, timeout=10, timeout_msg=None,
 import sys
 import deluge.core.daemon_entry
 
-sys.argv.extend(['-d', '-c', '%s', '-L', 'info', '-p', '%d'])
+from deluge.common import windows_check
+
+    if windows_check():
+        sys.argv.extend(['-c', '%s', '-L', 'info', '-p', '%d'])
```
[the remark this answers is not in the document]
doadin: yes -d isnt a option on windows so it says unrecognized option or something like that.

tox.ini
doadin: I had issues trying to get that to work in my testing. got various results including it saying something like invalid package pywin3 or :pywin. feel free to try but I couldn't get it right and with ONLY adding mywindows: pywin32 it didn't work if I remeber right.

doadin:
yea..I definitely messed up there hopefully I got it right this time?

deluge/tests/test_common.py
doadin: in linux if you do os.path.getsize(os.devnull) you get 0 in windows you get error parameter incorrect nul.

tox.ini
cas--: Well we cannot have this in tox.ini if it breaks on linux. 

Could it be related to this issue: https://github.com/mhammond/pywin32/issues/1177

doadin: not sure how thats related thats a issues, as stated in that issue, with pywin32's setup. I havn't even gotten it to go to that step. it errors trying to install a module not even in the list(pywin3, NOT pywin32) hence why its not valid and fails.

doadin:
Im not sure where ive gone wrong with this @ unittest change.

tox.ini
cas--: If you could provide the actual error that would be helpful.

cas--:
> Im not sure where ive gone wrong with this @ unittest change.

Apologies that is my bad, I forgot that twisted has it's own `unittest` so just revert to your original code.

tox.ini
cas--: I think there is a bug in tox when reusing deps, I was testing with `tox -e flake8` and this workaround worked:

```diff
--- tox.ini
+++ tox.ini
@@ -17,9 +17,9 @@ minversion = 2.0
 
 [testenv]
- platform =
-           mylinux: linux
-           mymacos: darwin
-           mywindows: win32
 install_command = pip install --ignore-installed {opts} {packages}
 passenv = DISPLAY PYTHONPATH
 setenv = PYTHONPATH = {toxinidir}
@@ -35,7 +35,6 @@ deps =
     mock
     slimit
     pillow
-    mywindows: pywin32
 whitelist_externals = pytest bash
 commands = {envpython} setup.py test
 
@@ -78,8 +77,13 @@ commands =
     pytest -v --basetemp=_pytest_temp -s deluge/plugins
 
 [testenv:py27]
+platform =
+    mylinux: linux
+    mymacos: darwin
+    mywindows: win32
 deps =
     {[testenv]deps}
+    mywindows: pywin32
     py2-ipaddress
 basepython = python2.7
 commands = {[testenv:pydef]commands}

```

All other combinations failed. So you need to specify pywin32 in each env

doadin: unless I did it wrong seems to work, besides a minor mistake I made, on travis(linux) but still a no go on appveyor(windows).

cas--:
Ok so I finally worked out that the platform support for tox relies on running `tox` without any env vars or using `tox -e` so adding `pywin32` to tox is not possible currently. I have created another PR #184 with the fixes and changes I wished to make.

PR #182, opened 2018-05-19
Fix/tox tests

cas--:


doadin:
Currently with this tox setup docs test fails to install twisted however the trial test seems to run fine which does install twisted as well. The only difference I see that would maybe make a difference is setenv in trial.

PR #187, opened 2018-06-04
Fix And Cleanup Interface Code

doadin:


deluge/core/core.py
```diff
@@ -159,20 +159,20 @@ def __init__(self, listen_interface=None, outgoing_interface=None, read_only_con
         # store the one in the config so we can restore it on shutdown
         self._old_listen_interface = None
         if listen_interface:
-            if deluge.common.is_ip(listen_interface):
+            if (windows_check() and deluge.common.is_ip(listen_interface)) or not windows_check():
```
gazpachoking: I'd probably avoid the duplicate windows_check, and just do:
```python
if deluge.common.is_ip(listen_interface) or not windows_check():
```

deluge/core/core.py
```diff
@@ -159,20 +159,20 @@ def __init__(self, listen_interface=None, outgoing_interface=None, read_only_con
...
                 self._old_listen_interface = self.config['listen_interface']
                 self.config['listen_interface'] = listen_interface
             else:
-                log.error('Invalid listen interface (must be IP Address): %s', listen_interface)
+                log.error('Invalid listen interface (must be adapter name or IP Address on unix ip on win): %s', listen_interface)
```
gazpachoking: Perhaps it's clearer to say:
`Invalid listen interface. (Must be IP Address or, on Linux, adapter name.)`? (That question mark is because I don't actually know if it's clearer.)

Alternatively, just add a comma, (and fix capitalization,) into your current message. i.e. `must be adapter name or IP address on Linux, or IP address on Windows`

doadin:
why should 'deliberate logic to not allow IP address for outgoing_interface (on non-windows) should not be removed!" NOT be removed? its a valid option on non windows its just not the BEST option.

doadin:
this ok now?

cas--:
I need to test what is and isn't allowed, I asked for clarification but no definitive answer: arvidn/libtorrent#3087

cas--:
After discussing with arvidn, I removed the checks and made the code changes: c415b097fe

Thanks for helping with this.

PR #189, opened 2018-06-10
[AutoAdd] Add WebUI interface

DjLegolas:
Adds a WebUi interface for AutoAdd plugin.

### DIFF
--- a/deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
+++ b/deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
@@ -90,6 +90,7 @@ class Core(CorePluginBase):
         self.config.save()
         self.watchdirs = self.config['watchdirs']
 
+        self.rpcserver = component.get('RPCServer')
         component.get('EventManager').register_event_handler(
             'PreTorrentRemovedEvent', self.__on_pre_torrent_removed,
         )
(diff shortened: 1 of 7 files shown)


DjLegolas:
ok, so i added everything but there is something strange.
It doesn't catch any errors returned.. for example look at the ```onAddClick``` in the file ```autoadd_options.js```.
Does the callbacks functions passed to the ajax are ```success``` and ```error```? as only the ```success``` callbacks are being called.

cas--:
Excellent, I shall have a look. I also noticed the multiple listings of autoadd plugin, I'm sure I encountered this before with other plugins but cannot remember the fix atm.

deluge/plugins/AutoAdd/deluge/plugins/autoadd/core.py
```diff
@@ -496,3 +496,13 @@ def __on_pre_torrent_removed(self, torrent_id):
                         'Failed to removed torrent file "%s" from "%s": %s',
                         torrent_fname, copy_torrent_path, ex,
                     )
+
+    @export
+    def is_admin_level(self):
+        rpcserver = component.get('RPCServer')
```
cas--: assign getting the component in the plugin __init__ to `self.rpcserver` and it can be reused in the entire class.

DjLegolas: Done.

DjLegolas:
> I also noticed the multiple listings of autoadd plugin, I'm sure I encountered this before with other plugins but cannot remember the fix atm.

Currently I set the ```prefsPage``` of the plugin as a static member of the ```Deluge.plugins.AutoAddPlugin``` class and check it's value before adding a page.

cas--:
I'm thinking that I'll merge this and deal with the duplicate listing later :)

cas--:
> Does the callbacks functions passed to the ajax are success and error? as only the success callbacks are being called.

Just checked code and it should be `success` and `failure`

DjLegolas:
Ok, now everything suppose to work, including showing errors.

cas--:
Thanks for this, merged!

PR #191, opened 2018-06-14
Use `gio open` instead of deprecated xdg-open on gnome 3.24

GianpaMX:
Solves [#3184](https://dev.deluge-torrent.org/ticket/3184)


GianpaMX:
Ups... bug is in xdg-utils and it's fix now. Sorry for the inconvenience

PR #192, opened 2018-06-15
[Core] Fix strip None in set_listen_on

DjLegolas:
Trying to strip None will not work so combined the check as a condition
for the strip.

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -210,8 +210,10 @@ class PreferencesManager(component.Component):
             self.config['listen_random_port'] = None
             listen_ports = self.config['listen_ports']
 
-        interface = str(self.config['listen_interface'].strip())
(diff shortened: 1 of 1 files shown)


deluge/core/preferencesmanager.py
```diff
@@ -210,8 +210,7 @@ def __set_listen_on(self):
             self.config['listen_random_port'] = None
             listen_ports = self.config['listen_ports']
 
-        interface = str(self.config['listen_interface'].strip())
-        interface = interface if interface else '0.0.0.0'
+        interface = str(self.config['listen_interface'].strip()) if self.config['listen_interface'] else '0.0.0.0'
```
cas--: The str() call can be removed and I would prefer a multi-line if statement here.

DjLegolas: Done

cas--:
thanks, merged

PR #193, opened 2018-06-16
[GTKUI] Fix GTK column sorting TypeError

DjLegolas:
Resolves [#3001](https://dev.deluge-torrent.org/ticket/3001) - GTK column sorting TypeError


cas--:
Thanks for PR but I decided to fix this properly with an working sync function and specified column: https://github.com/deluge-torrent/deluge/commit/bd78bd2643915319b236c2ae6fd33390bf3cc09a

PR #195, opened 2018-06-18
[#2867][WebUI] Fix Daemon connection problem

DjLegolas:
Trying to connect to daemon B while still connected to A will cause
the torrents from A to be shown after connecting to B.
Therefor, checking if connected to any daemon before connecting to B.

This will fix [#2867](https://dev.deluge-torrent.org/ticket/2867) - web-gui does not changing deluge instance properly.


cas--:
Thanks! Merged with develop

PR #196, opened 2018-06-27
WiP [#3065] Enhance TLS Security

cas--:


deluge/common.py
```diff
@@ -55,6 +55,25 @@
...
+# Sources for more information on TLS ciphers:
+# - https://wiki.mozilla.org/Security/Server_Side_TLS
+# - https://www.ssllabs.com/projects/best-practices/index.html
+# - https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
+#
+# The general intent is:
+# - prefer cipher suites that offer perfect forward secrecy (ECDHE),
+# - prefer AES-GCM over ChaCha20 because hardware-accelerated AES is common,
+# - disable NULL authentication, MD5 MACs and DSS for security reasons.
+TLS_CIPHERS = ':'.join(['ECDH+AESGCM',
```
cas--: Should avoid using hanging indents so drop the first item onto a new line and dedent.

Also, include a reference link to where this list originated from for aid future devs looking at this code.

deluge/core/rpcserver.py
```diff
@@ -100,10 +101,10 @@ def getContext(self):  # NOQA: N802
         SSL transport.
         """
         ssl_dir = deluge.configmanager.get_config_dir('ssl')
-        ctx = SSL.Context(SSL.SSLv23_METHOD)
-        ctx.set_options(SSL.OP_NO_SSLv2 | SSL.OP_NO_SSLv3)
+        ctx = SSL.Context(SSL.TLSv1_2_METHOD)
         ctx.use_certificate_file(os.path.join(ssl_dir, 'daemon.cert'))
         ctx.use_privatekey_file(os.path.join(ssl_dir, 'daemon.pkey'))
+        ctx.set_cipher_list(str(TLS_CIPHERS))
```
cas--: Should not need to use `str` here

deluge/ui/web/server.py
```diff
@@ -69,6 +69,17 @@
 )
 
 
+def remove_version_header(headers):
```
cas--: Instead of a major addition we could override the server version manually before creating the site:
```diff
+ server.version = 'TwistedWeb'
self.site = server.Site(self.top_level)
```
Since this is only set once when the server module is imported it should not be altered.

deluge/ui/web/server.py
```diff
@@ -668,9 +695,10 @@ def start_ssl(self):
             certificate = Certificate.loadPEM(cert.read()).original
         with open(configmanager.get_config_dir(self.pkey)) as pkey:
             private_key = KeyPair.load(pkey.read(), FILETYPE_PEM).original
-        options = CertificateOptions(privateKey=private_key, certificate=certificate, method=SSL.SSLv23_METHOD)
+        ciphers = AcceptableCiphers.fromOpenSSLCipherString(TLS_CIPHERS)
+        options = CertificateOptions(privateKey=private_key, certificate=certificate,
+                                     insecurelyLowerMinimumTo=TLSVersion.TLSv1_2, acceptableCiphers=ciphers)
```
cas--: I don't think we want `insecurelyLowerMinimumTo`, see https://twistedmatrix.com/documents/current/core/howto/ssl.html#tls-protocol-options

I have set TLS 1.2 for Python 3 changes using: `raiseMinimumTo=TLSVersion.TLSv1_2`

Do you think that the minimum that Twisted require will be an issue?

deluge/common.py
DjLegolas: Done.

deluge/core/rpcserver.py
DjLegolas: Done.

deluge/ui/web/server.py
DjLegolas: hmm, good idea.. Done.

deluge/ui/web/server.py
DjLegolas: No, I don't think so.
BTW, your change forced me to merge the two branches manually.

DjLegolas:
I'm trying to add tests, as you suggested, with no luck.
It seem as the server is up but when trying to connect to it with SSL\TLS, it does not respond to the protocol, and handle the negotiation process. And this cause the test process to halt and wait to results.
My initial try is in [this](https://pastebin.com/UtJ91YZx) pastebin
@cas-- Any ideas how to fix it?

cas--:
I didn't realise the tests were so slow. Perhaps leave it for now then. I did see https://github.com/tomato42/tlsfuzzer/

The issues you are facing with running a subprocess command in tests is likely related to the twisted reactor. I had success with the following instead: 

```
        from twisted.internet.utils import getProcessOutputAndValue
        d = getProcessOutputAndValue(
            'bash',
            [
            get_test_data_file('testssl.sh'),
            '-p',
            '127.0.0.1:8999',
        ])
        def on_result(result):
            #print(result)
            self.assertIn('OK', result[0])
            self.assertNotIn('Not OK', result[0])
        d.addCallback(on_result)
        return d
```

DjLegolas:
I think I didn't explain the problem correctly.
The script will take time to run, but it's not the problem we have.

When the test run, it starts the WebUI with `self.deluge_web.start()`. The expected behavior is the ability to connect to the server itself by connecting to it's IP and port - `127.0.0.1:8999` and being able to use it. Unfortunately, this is not the case.

When connecting to it with a web browser, firefox and chrome, there is no response from the WebUI at all. All you get is a message that the server is down.
Connecting with `openssl -s_client -connect IP:PORT` only shows `CONNECTED(*some-code*)` and the handshake process being stuck after the 1st phase, `CLIENT HELLO`. A good connection will show the important info about the secured connection, like public cert, protocol version and agreed cipher. 

BTW, this testing library looks cool..  I'll check it out.

DjLegolas:
Ok, I've implemented the tests using `testssl.sh`. This effects the duration of `pydef` and `trial` to be completed.

In addition, Not all tests currently passes as a further configuration is needed for `Twisted`:
* `Client-Initiated Renegotiation` needs to be canceled.
* There is no `server cipher order` for the daemon.

Finally, some of the tests are for `http\s` servers, so whey are being skipped.

deluge/common.py
```diff
@@ -56,6 +58,50 @@
 
 log = logging.getLogger(__name__)
 
+# A TLS ciphers list.
```
cas--: Perhaps instead of adding more code to common.py create a new crypto.py or crypto_utils.py file

DjLegolas: Done. Moved to `crypto_utils.py`

DjLegolas:
> Impressed with the additional tests but not sure about extra test time. Perhaps not run these by default

Yeah, I know. As a fix for `pytest`, I've added a new `toxenv` only for the security tests and set the main test run to ignore it.
The problem is going to be with `trial` run, as currently it tests everything with no option to split the tests... Please write if you know of any way to do it.

cas--:
You could set an env var and skip the tests if not set

```python
SECURITY_TESTS = bool(os.getenv('SECURITY_TESTS', False))

if not SECURITY_TESTS:
    skip = 'Skipping security tests'
```

DjLegolas:
> Can the testssl package be removed from the code and downloaded by tox. The tests would also need to be skipped if testssl.sh is missing.

It can be done, but this will extend the running time of the test as we need to download 2 files, `testssl.sh` and `etc.tar.gz`, and extract the later (or download everything from its [repo](https://github.com/drwetter/testssl.sh/releases) and then extract).

To do so, I'll need to add to `commands = ` the commands to download and extract the files, right?

cas--:
> To do so, I'll need to add to commands = the commands to download and extract the files, right?

Actually, I just realised that the code should be downloaded by travis instead in `install` section. Tox is for setting up a python test env so we should specify in the documentation that testssl.sh is a requirement.

Use whatever is the simplest way to install the files, using the releases seems like a good idea. 

To deal with the time penalty I intend to only run these security tests either manually or with commits on the `develop` or `master` branch. So not PRs or other branches. Another idea so you can run them in a PR is to include something like the word `SECURITY_TEST` in the commit message.

deluge/crypto_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    '!MD5'
+])
+
+# This value tells OpenSSL to disable all SSL/TLS renegotiation.
+SSL_OP_NO_RENEGOTIATION = 0x40000000
+
+
+def get_context_factory(cert_path, pkey_path):
+    """
+    Generates an OpenSSL context factory using Twisted's CertificateOptions class.
```
cas--: Can we use google style docstrings here please

cas--:
Thanks! Merged

PR #197, opened 2018-07-01
Migrate GTKUI to a GTK3UI as separate module

cas--:
To prevent confusion I think it will be best to create a new module for GTK3, it will make development a bit harder with duplicate code but should help with moving users to GTK3. Also, it can be independent of GTK2 and released as it becomes ready and hopefully make it easier for plugin authors to add support.

This branch does display the main windows with torrents (even under Python 3) but there are a lot of errors still needing to be fixed. 

 - Move from GtkBox to GtkGrid for future-proofing: https://developer.gnome.org/gtk3/3.10/gtk-migrating-GtkGrid.html
 - Split gtk3 code into separate module
 - Update plugin GTK3 code
 - Remove deprecated from UI files with a script.
 - Document the migration process to aid plugin authors.
 - Review and fix all FIXME comments.


cas--:
@kbdserver Did you mean to post in #200 that looks really useful though, yes share the code please :+1:

kbdserver:
With pleasure
[deluge.7z.zip](https://github.com/deluge-torrent/deluge/files/2420576/deluge.7z.zip)

PR #200, opened 2018-07-01
Feature GTK3UI

cas--:
To prevent confusion I think it will be best to create a new module for GTK3, it will make development a bit harder with duplicate code but should help with moving users to GTK3. Also, it can be independent of GTK2 and released as it becomes ready and hopefully make it easier for plugin authors to add support.

This branch does display the main windows with torrents (even under Python 3) but there are a lot of errors still needing to be fixed. 

- [x]  Split gtk3 code into a separate module
- [x]  Update plugin GTK3 code
- [x]  Remove deprecated from UI files with a script.
- [x]  Document the migration process to aid plugin authors.
- [x] Bump plugin versions.
- [x]  Review and fix all FIXME comments.
- ~Move from GtkBox to GtkGrid for future-proofing: https://developer.gnome.org/gtk3/3.10/gtk-migrating-GtkGrid.html~
 
More fixes from hugo merging PR:

- [x] Remove `interface-naming-policy project-wide`
- [x] Add `Generated with`
- [x] Add alternative to purged props (align, padding)
- [x] Fix icon-size
- ~Add test for windowing~

Found a useful porting guide:

https://github.com/sugarlabs/sugar-docs/blob/master/src/gtk3-porting-guide.md


cas--:
Also useful for debugging the UI
```
gsettings set org.gtk.Settings.Debug enable-inspector-keybinding true
GTK_DEBUG=interactive deluge-gtk
```

Based upon https://chipx86.github.io/gtkparasite/ which does support gtk3 too.

cas--:
Two tracebacks to investigate:

```
  File "/home/calum/projects/deluge/deluge/ui/gtk3/listview.py", line 111, in on_button_pressed
    self.emit('button-press-event', event)
TypeError: could not convert type EventButton to GdkEvent required for parameter 0
  File "/home/calum/projects/deluge/deluge/ui/gtk3/listview.py", line 225, in record_position
    self.last_sort_order[model[_iter][self.unique_column_id]] = int(str(model.get_path(iter)))
TypeError: argument iter: Expected Gtk.TreeIter, but got __builtin__.builtin_function_or_method
```

luxus:
I‘m not sure but did you see Hugo’s effort?
https://github.com/hugosenari/deluge/commits/Feature/GTK3UI

kbdserver:
For future UI design:
    added support for hi-dpi displays (use svg system and app icons);
    changed pieces bar;
    added session pause/resume support from main window;
    use GIcon to files_tab to show icons according to system mime-types;
    сonnection manager became more friendly to user;
    updated country flag tooltip;
    adapted to GTK3 and tested Scheduler, Notifications, ItConfig plugins.
Solved UI bugs:
    sorting in torrent view;
    renaming files in files_tab problem;
    сonnection manager bugs;
    restore mainwindow vpaned, hpaned position at startup;
    right click on torrentview header opens two context menus.
[Screenshots.zip](https://github.com/deluge-torrent/deluge/files/2420772/Screenshots.zip)
[deluge.7z.zip](https://github.com/deluge-torrent/deluge/files/2420774/deluge.7z.zip)

kbdserver:
BUG: in TorrentView and PeersTabView, when moving the mouse pointer through rows, the value in download and upload speed columns can be overwritten by value from previous row cell. 
The solution is to disable return by cached condition in function cell_data_speed (torrentview_data_funcs.py).

[cell_data_speed.txt](https://github.com/deluge-torrent/deluge/files/2421940/cell_data_speed.txt)

cas--:
Do you have a git branch with these changes? It would make collaboration easier

kbdserver:
I don't have git branch and have not yet learned how to use your branch cas--:Feature/GTK3UI as original.

About caching:
According to the developer's comment:
Cache the key used to calculate the current value set for the specific cell renderer. This is much cheaper than fetch the current value and test if it's equal.

Let's to analyse:
In case of cache using we need save not only cell value, but path+value. Different path cell will have different cache.
So, we have: 
- get path from iter, then compare cache with current path and value
- then if cache not equal assign new path+value cache 
- then set current value for the specific cell

Without cache just need:
- setting current value for the specific cell. No need to "fetch the current value and test if it's equal"

I compared time for operations, "compare cache with current path and value" time is almost the same as for "setting current value for the specific cell".

I agree, that may be, in case of setting pixbuf as value (large data) or long time get value operations, using cache is better. But for string data is not. 

Let's to delete caching for ordinary cells.

cas--:
Caching is enabled on all cells so this can't be isolated to the speed cell. Could you provide more information, e.g. video, on the bug as I don't see the issue you are reporting.

kbdserver:
The point of the caching is when dealing with thousands of torrents
_I won't help for thousands of torrents because it's not correctly save only value in cache, need to save {model.get_path(row)[0], value}. Caching can be isolated for all cells except pixbuf column. I will provide video later_

kbdserver:
The point of the caching is when dealing with thousands of torrents
_It won't help for thousands of torrents because function saves in cache only value, but path is also needed - {model.get_path(row)[0], value}. Caching can be isolated for all cells except pixbuf column._

[cached.mp4.zip](https://github.com/deluge-torrent/deluge/files/2424701/cached.mp4.zip)

cas--:
Thanks for the video, I can replicate now. So the issue is that with both TorrentView and PeersTabView values are appearing in blank cells. Would this not be solved by not reusing the `cell_data_speed_down` func in Peers Tab so create a `cell_data_peer_speed_down` func?

kbdserver:
I suppose the problem is not in sharing the same cell renderer func.
On the other hand, the function is called only when the cell is updated, so don't worry about thousands of torrents, since they won't work at the same time.
There is another solution for treeview with lot of rows initial loading or sorting.

kbdserver:
Also I cannot simply merge all your changes as there are parts that I don't think are corrent.
_No need merge all, just agreed parts, others we could discuss_

kbdserver:
Many people like deluge, but they can not use because it looks ugly in a HiDPI environment.
Need to release GTK3 HiDPI version ASAP

kbdserver:
Why Linux distributives use transmission as default ? Because this project is alive. We should make Deluge project alive also.

kbdserver:
If you do not want to communicate in public, this is my e-mail: kbdserver@mail.ru

kbdserver:
Calum, solution for UnicodeWarning problem in torrentview.py:
[UnicodeWarning_patch.txt](https://github.com/deluge-torrent/deluge/files/2475425/UnicodeWarning_patch.txt)

luxus:
is this the blocker for the next release? "watching" the deluge reposity makes me really impatient

kbdserver:
According to my observations, Calum is not interested in a quick release. He makes a lot of important changes, but when the release will come is not clear.

kbdserver:
> @kbdserver I don't see how `long` has any effect on Unicode and in Python `int` and `long` were unified.

I assumed that your comment # FIXME: Seeing UnicodeWarning ?? was associated with this warning:

[Failure instance: Traceback: <type 'exceptions.OverflowError'>: 3072227291 not in range -2147483648 to 2147483647
/home/kbdserver/sources/deluge/deluge/core/torrentmanager.py:1441:on_alert_state_update
/home/kbdserver/sources/deluge/deluge/core/torrentmanager.py:1499:handle_torrents_status_callback
/usr/lib64/python2.7/site-packages/twisted/internet/defer.py:459:callback
/usr/lib64/python2.7/site-packages/twisted/internet/defer.py:567:_startRunCallbacks
/usr/lib64/python2.7/site-packages/twisted/internet/defer.py:653:_runCallbacks
/home/kbdserver/sources/deluge/deluge/ui/gtk3/torrentview.py:611:_on_get_torrents_status
/home/kbdserver/sources/deluge/deluge/ui/gtk3/torrentview.py:589:update_view
/usr/lib64/python2.7/site-packages/gi/overrides/Gtk.py:1052:set
/usr/lib64/python2.7/site-packages/gi/overrides/Gtk.py:1046:_set_lists
/usr/lib64/python2.7/site-packages/gi/overrides/Gtk.py:915:_convert_value
/usr/lib64/python2.7/site-packages/gi/overrides/GObject.py:218:__init__
/usr/lib64/python2.7/site-packages/gi/overrides/GObject.py:249:set_value
]

If so, warning is related to the type of the variable in the liststore.

cas--:
Ah well yes that is a different issue and the solution is to limit the maximum size of ETA returned by torrent status. Perhaps put a max value of `31557600`, which is a year.

kbdserver:
You're right, it's better:
[ETA_FIX.txt](https://github.com/deluge-torrent/deluge/files/2496506/ETA_FIX.txt)

kbdserver:
For future UI design Part2:
-   Changed the trackers tab - edit tracker in tab or remove, move in multi-selection mode directly. Button Bar can be activated if needed;
-   Added flags information in peer tab to understand how this peer is connected;
-   Others design improvements.
Video and source code attached.
[For_Future_Design_Part2.mp4.zip](https://github.com/deluge-torrent/deluge/files/2496960/For_Future_Design_Part2.mp4.zip)
[deluge.7z.zip](https://github.com/deluge-torrent/deluge/files/2496962/deluge.7z.zip)

cas--:
@kbdserver Please provide this as a pull request, patches or link to a branch. It is too much time and effort to do this with a source code zip file.

kbdserver:
You do not understand, Calum. I want to help the project survive. Your problem is not in a zip file

kbdserver:
Maybe you're trying to get more money from https://www.patreon.com/deluge_cas.
 But also think about the users who like the program.

cas--:
> You do not understand, Calum. I want to help the project survive. 

If you want to do that, then submitting pull requests is how you help me. Code needs reviewed before it can be merged and providing a zip is not how software development works these days. It is how the develop branch got into this mess in the first place.

>Maybe you're trying to get more money from https://www.patreon.com/deluge_cas.

Disparaging comments are not welcome here.

luxus:
@kbdserver can I help you to create a pull request?

kbdserver:
> @kbdserver can I help you to create a pull request?

I will be grateful, thanks!

kbdserver:
> > 
> 
> If you want to do that, then submitting pull requests is how you help me. Code needs reviewed before it can be merged and providing a zip is not how software development works these days. It is how the develop branch got into this mess in the first place.

This path is very slow, deluge lost the advantage.

kbdserver:
Calum, you are a professional programmer, I do not even doubt that you have invested in the project. Perhaps a reboot is necessary - Deluge 3.0.0

kbdserver:
Good luck in the project.

kbdserver:
Many want to stop the project.
Fuck off yourself  bendikro

kbdserver:
Blocklist plugin GTK3 fixes
[Blocklist_fixes_diff.txt](https://github.com/deluge-torrent/deluge/files/2526074/Blocklist_fixes_diff.txt)
blocklist

PR #201, opened 2018-07-08
[#3189] Refactor HTTPDownloader to Agent

DjLegolas:
As of twisted 16.7.0, `twisted.web.client.HTTPDownloader` have been
marked as deprecated.
This caused the tests results to show many lines of warnings about it.
This refactor uses `twisted.web.client.Agent`, as suggested by Twisted.

Before this refactor - 166 warnings.
After this refactor - 28 warnings.

https://dev.deluge-torrent.org/ticket/3189


deluge/httpdownloader.py
```diff
@@ -52,43 +79,58 @@ def __init__(
...
+            http.SEE_OTHER,
+            http.TEMPORARY_REDIRECT,
+        ):
+            location = response.headers.getRawHeaders(b'location')[0]
+            error = PageRedirect(response.code, location=location)
+            finished.errback(Failure(error))
+
+        else:
+            headers = response.headers
+            if headers.hasHeader(b'content-length'):
```
cas--: Probably can use the getRawHeaders default value instead of this test

deluge/httpdownloader.py
```diff
@@ -31,13 +33,38 @@
...
+
+class HTTPDownloaderAgent(client.Agent):
     """
-    Factory class for downloading files and keeping track of progress.
+    A File Downloader Agent
     """
     def __init__(
-        self, url, filename, part_callback=None, headers=None,
-        force_filename=False, allow_compression=True,
+            self, url, filename, part_callback=None, headers=None,
```
cas--: Remove extra indent

deluge/httpdownloader.py
```diff
@@ -31,13 +33,38 @@
 log = logging.getLogger(__name__)
 
 
-class HTTPDownloader(client.HTTPDownloader):
+class _BodyHandler(HTTPClientParser):
```
cas--: I don't think this class needs an underscore prefix

deluge/httpdownloader.py
```diff
@@ -52,43 +79,58 @@ def __init__(
...
         :type headers: dictionary
         """
+        if not headers:
+            headers = {}
+        else:
+            for key, value in headers.items():
+                headers[key] = [value]
+        headers.update({'User-Agent': ['Deluge/%s (http://deluge-torrent.org)' % get_version()]})
 
+        self.url = bytes(url)
```
cas--: use `.encode()` please

DjLegolas: Do you mean `self.url = bytes(url.encode('utf-8'))`? Or something else?

DjLegolas:
> I don't see anything for TLS SNI which was in the original code

The default behavior is a browser-like HTTPS handler, so this should be covered by default. But to be sure, do you know of any site that uses cloudflare so we could test?

deluge/httpdownloader.py
cas--: Even simpler `self.url = url.encode()`, default ascii encoding as a url strictly shouldn't have non-ascii chars.

cas--:
From that ticket, I think https://www.seo.com/ is still a usable test case

DjLegolas:
Oh, ok. `seo,com`'s works only with browsers with SNI support.
So I added it to the icon tests as it will show us the code ability to use it and download a file from it.
As you can see, the test passes.

deluge/httpdownloader.py
```diff
@@ -52,43 +79,55 @@ def __init__(
...
-                self.total_length = int(headers[b'content-length'][0])
-            else:
-                self.total_length = 0
+        self.decoder = None
+        client.Agent.__init__(self, reactor)
+
+    def request_callback(self, response):
+        finished = Deferred()
+
+        if response.code in (
```
cas--: I was thinking we should be able to use twisted RedirectAgent to deal with redirects and also ContentDecoderAgent for decoding gzip thus simplify this code. 

To deal with deflate might need to subclass GzipDecoder to add 32 instead of 16.

What are your thoughts?

deluge/httpdownloader.py
```diff
@@ -31,13 +33,38 @@
 log = logging.getLogger(__name__)
 
 
-class HTTPDownloader(client.HTTPDownloader):
+class BodyHandler(HTTPClientParser):
+    def __init__(self, request, finished, length, agent):
+        HTTPClientParser.__init__(self, request, finished)
```
cas--: Better to use super so no don't need to refer to parent class so:
```python
super(BodyHandler, self).__init__(request, finished)
```

deluge/httpdownloader.py
```diff
@@ -31,13 +33,38 @@
...
+
+    def connectionLost(self, reason):  # NOQA: N802
+        with open(self.agent.filename, 'wb') as _file:
+            _file.write(self.data)
+        self.finished.callback(self.agent.filename)
+        self.state = u'DONE'
+        HTTPClientParser.connectionLost(self, reason)
+
+
+class HTTPDownloaderAgent(client.Agent):
```
cas--: I am not sure about subclassing Agent, I would prefer to pass agent as an arg so follow the style of the other web client agents: https://github.com/twisted/twisted/blob/trunk/src/twisted/web/client.py#L1941

This would mean that url and headers should be passed to request instead of HTTPDownloaderAgent

deluge/httpdownloader.py
DjLegolas: hmm, this will simplify things...
But some things will have to the same, like how to handle a redirect when not using `RedirectAgent`. The current code of `TrackerIcon` is using `HTTPDownloader` but handles the redirect of the original page by itself and thous requires a `PageRedirect` if the page marked as redirect.

deluge/httpdownloader.py
DjLegolas: This one will need me to change to `class BodyHandler(HTTPClientParser, object):` as `HTTPClientParser` is old style object, and super needs a new style class.

cas--: Ah! I don't see that being an issue though, just explain that it in commit message.

deluge/httpdownloader.py
cas--: There is the `handle_redirects` arg (I assume Trackicons uses this) for `download_file()` that could be used to determine whether to wrap the agent with `RedirectAgent`?

DjLegolas: Yes, and in the last commit I use it for this purpose. Just explaining why some of the code must stay.

cas--:
Thanks! Merged with some minor changes applied: c7e61f8c3461d59b5

PR #202, opened 2018-07-13
Prevent time formatting crash when seconds were floats

gazpachoking:
Was getting this error in deluge-console:
```
  File "/home/chase/newdeluge/deluge/ui/console/modes/basemode.py", line 207, in doRead
    self.read_input()
  File "/home/chase/newdeluge/deluge/ui/console/modes/torrentlist/torrentlist.py", line 332, in read_input
    self.refresh(affected_lines)
  File "/home/chase/newdeluge/deluge/ui/console/modes/torrentlist/torrentlist.py", line 223, in refresh
    self.torrentview.update_torrents(lines)
  File "/home/chase/newdeluge/deluge/ui/console/modes/torrentlist/torrentview.py", line 312, in update_torrents
    todraw.append((i, i - self.curoff, draw_row(i)))
  File "/home/chase/newdeluge/deluge/ui/console/modes/torrentlist/torrentview.py", line 291, in draw_row
    [get_column_value(name, ts) for name in self.cols_to_show],
  File "/home/chase/newdeluge/deluge/ui/console/utils/column.py", line 75, in get_column_value
    return col['formatter'](*args)
  File "/home/chase/newdeluge/deluge/ui/console/utils/format_utils.py", line 33, in format_time
    return deluge.common.ftime(time)
  File "/home/chase/newdeluge/deluge/common.py", line 511, in ftime
    time_str = '{:d}w {:d}d'.format(secs // 604800, secs // 86400 % 7)
ValueError: Unknown format code 'd' for object of type 'float'
```
Updated the `ftime` function to allow formatting seconds of type float.

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -500,17 +500,17 @@ def ftime(secs):
     if secs <= 0:
         time_str = ''
     elif secs < 60:
-        time_str = '{:d}s'.format(secs)
+        time_str = '{:.0f}s'.format(secs)
     elif secs < 3600:
-        time_str = '{:d}m {:d}s'.format(secs // 60, secs % 60)
+        time_str = '{:.0f}m {:.0f}s'.format(secs // 60, secs % 60)
     elif secs < 86400:
-        time_str = '{:d}h {:d}m'.format(secs // 3600, secs // 60 % 60)
+        time_str = '{:.0f}h {:.0f}m'.format(secs // 3600, secs // 60 % 60)
     elif secs < 604800:
-        time_str = '{:d}d {:d}h'.format(secs // 86400, secs // 3600 % 24)
+        time_str = '{:.0f}d {:.0f}h'.format(secs // 86400, secs // 3600 % 24)
     elif secs < 31449600:
-        time_str = '{:d}w {:d}d'.format(secs // 604800, secs // 86400 % 7)
+        time_str = '{:.0f}w {:.0f}d'.format(secs // 604800, secs // 86400 % 7)
     else:
-        time_str = '{:d}y {:d}w'.format(secs // 31449600, secs // 604800 % 52)
+        time_str = '{:.0f}y {:.0f}w'.format(secs // 31449600, secs // 604800 % 52)
 
     return time_str
 


cas--:
I would prefer to know why console is sending float and fix it there. The function accepts int so lets leave it as that.

aresch:
Probably should add some more cases in the test: https://github.com/deluge-torrent/deluge/blob/develop/deluge/tests/test_common.py#L55

gazpachoking:
Hmm, I can't get it to crash again, not exactly sure where the problem was coming from. It was totally unusable while it was happening, with the crash traceback being uncaught, and printing out everywhere all over the screen. I agree it'd be nice to know why it was getting a float, but I'm not exactly sure why we wouldn't want the time formatter to also handle floats. (Though I also agree the docstring and tests need updated if we do go that route.)

deluge/tests/test_common.py
```diff
@@ -53,15 +53,17 @@ def test_fpeer(self):
...
-        self.assertTrue(ftime(0) == '')
-        self.assertTrue(ftime(5) == '5s')
-        self.assertTrue(ftime(100) == '1m 40s')
-        self.assertTrue(ftime(3789) == '1h 3m')
-        self.assertTrue(ftime(23011) == '6h 23m')
-        self.assertTrue(ftime(391187) == '4d 12h')
-        self.assertTrue(ftime(604800) == '1w 0d')
-        self.assertTrue(ftime(13893086) == '22w 6d')
-        self.assertTrue(ftime(59740269) == '1y 46w')
+        assert ftime(0) == ''
```
cas--: Using python assert is for pytest and will not output traceback with trial on failure so please revert this change. The assertTrue could be replaced with assertEquals but it's not really worth unless changing whole test file...

gazpachoking: Isn't pytest what's being used now?

gazpachoking:
I suspect you are right on the eta being the thing. I'll defer to you on how you want to fix.

deluge/tests/test_common.py
cas--: Well, both are in use and I can't remember what the end-goal was... It doesn't actually test the twisted deferreds properly so would need this plugin: https://github.com/pytest-dev/pytest-twisted

The point is that changing it for one subset of tests is not a good idea. The tests themselves are subclassing trial unittest so best to keep that compatibility.

gazpachoking: No prob. I thought the intention was to switch over to pytest, and had introduced a failure that wasn't giving me any good info with assertTrue.

gazpachoking:
You wanna scrap this then, in favor of 1e6c02a?

deluge/tests/test_common.py
cas--: Sure it would be good to move to pytest entirely but it's better to do it as part of an overall change rather than here and there and potentially cause more headaches.

cas--:
I'm still not sure about the idea of supporting both int and float but let's go with your addition of float support too :)

deluge/common.py
```diff
@@ -497,20 +497,21 @@ def ftime(secs):
 
     """
 
+    secs = int(secs)
```
cas--: Add a small comment to indicate truncating floats and casting to int

PR #203, opened 2018-07-13
[#2837][WebUI] Handle missing gettext.js file

DjLegolas:
WebUI code now handle a missing `gettext.js` file. 

When `DelugeWeb` initialized and set `TopLevel`, `TopLevel` will check if `gettext.js` exists and will create it in case it's not.
As `setup.py` creates the file itself, this check is only for a case in which `gettext.js` somehow got deleted and the WebUI starts.
Last thing, as a result of this behavior, `gettext.js` is now deleted from the repo.

https://dev.deluge-torrent.org/ticket/2837


deluge/ui/web/server.py
```diff
@@ -470,6 +472,13 @@ def __init__(self):
             theme = CONFIG_DEFAULTS.get('theme')
         self.__stylesheets.insert(1, 'themes/css/xtheme-%s.css' % theme)
 
+    @staticmethod
+    def _check_and_build_gettext_js():
+        js_path = os.path.join(os.path.dirname(__file__), 'js')
+        if not os.path.exists(os.path.join(js_path, 'gettext.js')):
+            from gen_web_gettext import create_gettext_js
```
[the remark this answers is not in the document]
DjLegolas: OK, now changed to mocking function `_`.

deluge/ui/web/server.py
```diff
@@ -431,8 +441,19 @@ class TopLevel(resource.Resource):
 
     def __init__(self):
         resource.Resource.__init__(self)
+        gettext_js_exists = os.path.exists(
+            os.path.join(
+                os.path.join(os.path.dirname(__file__), 'js'),
```
cas--: `os.path.join` can take any number of args and join them so only one needed. Also `isfile` is preferred to `exists` as we are definitely looking for a file. 

Also there is the `rpath` function to simplify generating absolute script paths. So you can remove this code and just use:

```python
if os.path.isfile(rpath('js', 'gettext.js'):
   self.putChild ...
```

deluge/ui/web/server.py
```diff
@@ -83,6 +83,16 @@ def render(self, request):
         return compress(template.render(), request)
 
 
+class MockGetText(resource.Resource):
+    """Mock the file `gettext.js` in case it does not exists so there will still be
```
cas--: The first line of a docstring should not continue onto next line and should end with a full-stop. 

This is the style we should be using: http://www.sphinx-doc.org/en/stable/ext/example_google.html

cas--:
Merged, thanks.

PR #204, opened 2018-07-14
[#2009][WebUI] Add About window

DjLegolas:
https://dev.deluge-torrent.org/ticket/2009


DjLegolas:
I have a dilemma as to whether to include the credits and/or license or not because they contain relatively a lot of information that I am not sure if it is necessary to include.
Of curse, it's still under develop, but mostly because of this dilemma. The main `About` window is finished and only needs a toolbar icon.

cas--:
The About should contain a short amount of information as it is a pain to maintain otherwise.

We have all the detailed information in the source code, which is sufficient for now. Therefore I would credit writing the application to the Deluge Team and leave out the translation authors (it's pretty outdated anyway). The licence can be something like `GPL v3.0 or later`. Properly including it all in documentation might be useful in future as a central resource to point to.

cas--:
Could you include a screenshot?

DjLegolas:
Connected | not connected
-|-
about_connected | about_not_connected

As you can see, I tried to make is as close to the one in the GTK. This is why I wanted to know about the credits and license.

deluge/ui/web/server.py
```diff
@@ -435,7 +435,17 @@ def __init__(self):
         self.putChild('gettext.js', GetText())
         self.putChild('flag', Flag())
         self.putChild('icons', LookupResource('Icons', rpath('icons')))
-        self.putChild('images', LookupResource('Images', rpath('images')))
+        images = LookupResource(
+            'Images',
+            rpath('images'),
+            os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
```
cas--: The data dir path for ui will not always be at this location so should use `resource_filename` with module name `deluge.ui` and I would also use a separate location to differentiate from web ui images. e.g. `putChild('ui_images', ...`

DjLegolas:
Any ideas why the tests are failing?

DjLegolas:
Ok, I want to finish this feature...
@cas--, is it fine if i'll remove the credits window, and set the License to [LICENSE](https://github.com/deluge-torrent/deluge/blob/develop/LICENSE) file?

cas--:
Yes removing credits is fine and I don't think we need the license one either. I don't see it being of use for end-users.

cas--:
Fyi: If you rebase on develop, install [pre-commit](http://pre-commit.com/) and use:

    pre-commit run --all-files

That will format the python with black and js with prettier.

If you install it then it will run on the git commit files:

    pre-commit install

I am writing some docs to better explain the setup ;)

DjLegolas:
Hmm, @cas--, looks like something is missing...
When I tried running it, I got an error: 
```
An error has occurred: InvalidConfigError:
=====> D:/progrmming/python/deluge\.pre-commit-config.yaml does not exist
```

cas--:
Yeah I only just added today: 0548bdb6 :stuck_out_tongue_closed_eyes:

DjLegolas:
LOL, OK. 😂

2 things...
1. Is there any quick way to get the connected server hash ID?
I want to get its version number, but can do so only with `deluge.client.web.get_host_status`, which requires the server ID...
2. Do you know who controls the bots of github and trac on IRC? They are both out of the main room and, therefor, no info is shown when publishing stuff.

cas--:
1. The `host_id` is not readily other than parsing the `web.get_hosts` for `host_id` with `connected` status. If you are looking for the version then using the rather obscure RPC method `client.daemon.info` returns the version of the currently connected daemon.

2. If you mean drizzle bot that's a deluge service that I maintain... I haven't been on IRC recently so didn't notice that need to be registered to join the channel now. The bot is now registered and hopefully will spam the channel once more ;) I think the github one was gazpachoking's but that's just a guess.

DjLegolas:
Ok, interesting, but I didn't find a way to get this method to execute through the WebUI. Am I missing something? Is there another way to send RPC commands?

cas--:
Ah it looks like some extra methods won't be allowed by the json_api because they are not actually in the RPCServer registered methods list. It probably would make sense to export a version method in daemon:

```diff
diff --git deluge/core/daemon.py deluge/core/daemon.py
index 84c4514a1..ebb30de2f 100644
--- deluge/core/daemon.py
+++ deluge/core/daemon.py
@@ -182,6 +182,10 @@ def get_method_list(self):
         """Returns a list of the exported methods."""
         return self.rpcserver.get_method_list()
 
+    @export()
+    def get_version(self):
+        return get_version()
+
     @export(1)
     def authorized_call(self, rpc):
         """Determines if session auth_level is authorized to call RPC.
```

DjLegolas:
Yep, this will do the trick.

BTW, do you have an idea for an icon to add to the `About` button itself?

cas--:
What do you think about using the Deluge icon and text on the left-hand side as the button to show about?

DjLegolas:
Like this?
deluge-web-about

cas--:
Actually mean making Deluge clicking and popping up the about dialog

DjLegolas:
Ok, I think all of the work is done.

cas--:
Thanks, merged

PR #207, opened 2018-09-02
Add a background position and size to sidebar icons

metaa:
Without this change tracker icons that are too big render ... well, as too big.
This restrains them to fit into the sidebar list item.

What it looks like now:
image

What it will look like after:
image

(the red border serves the purpose to visualize the size of the sidebar item)


cas--:
Thanks, this was something I puzzled over before. Merged to develop

metaa:
Just for the protocol, this is the commit: 456e720b756c39408402410f2ad9d421c1a1b01b

PR #211, opened 2018-10-04
[WebUI] Copy non-minified JS file if slimit missing

DjLegolas:
Will solve [#3053](https://dev.deluge-torrent.org/ticket/3053)

### DIFF
--- a/minify_web_js.py
+++ b/minify_web_js.py
@@ -57,7 +57,10 @@ if module_exists('closure'):
 elif module_exists('slimit'):
     from slimit import minify
 else:
-    raise ImportError('Requires "slimit" package for minifying WebUI files.')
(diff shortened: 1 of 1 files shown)


minify_web_js.py
```diff
@@ -57,7 +57,8 @@ def minify_closure(file_in, file_out):
 elif module_exists('slimit'):
     from slimit import minify
 else:
-    raise ImportError('Requires "slimit" package for minifying WebUI files.')
```
cas--: Could a warning be printed that the files are not being minified

DjLegolas: yeah, sure. I'll add it.

cas--:
Merged

PR #215, opened 2018-10-11
Resolve symlinked config files

PhilHudson:
Currently, deluge replaces symlinks rather than their targets. This is *Bad*.

### DIFF
--- a/deluge/config.py
+++ b/deluge/config.py
@@ -455,6 +455,7 @@ class Config(object):
         """
         if not filename:
             filename = self.__config_file
+        filename = os.path.realpath(filename)
         # Check to see if the current config differs from the one on disk
         # We will only write a new config file if there is a difference
         try:


PhilHudson:
On a unix-alike OS -- GNU+Linux, MacOS X, *BSD, anything that supports symbolic links -- say my deluge config file is kept under source control at $HOME/scm/$OS/$HOST/$HOME/.config/deluge/core.conf and I have a symbolic link which points to it at the usual location $HOME/.config/deluge/core.conf -- currently when I save any config changes, a complete config file will replace that symlink -- *not* its target -- instead of Doing The Right Thing, which would be to follow the symlink (using os.path.realpath) and overwrite the target file it points to. Changing a file's type from symlink to text file tramples on the user's legitimate wishes and intentions, for no good reason and no benefit to anyone. The change I propose loses nobody anything and gains symlink users correct application behavior, would be my claim in support of it. (For clarity, os.path.realpath returns its argument unchanged if it is not a symlink). I hope this clarifies why the current behavior is bad.

cas--:
Thanks, I just needed to know the use-case. The badness is completely subjective.

PhilHudson:
It is unacceptable for any program to change a symbolic link to a data file without confirmation, at any time, for any reason, or as in this case, for no reason at all. Doing so is objectively a bug. This PR is the one-line fix you need. You lose nothing by doing the right thing; you gain nothing by doing the wrong thing. Please accept the change.

cas--:
Please don't muddy PR discussion with unneeded hyperbole.

This is a feature request, there has never been an expectation that Deluge config files could be symlinks and a program overwriting it's own config files is perfectly acceptable, it is perhaps undesirable for a narrow use-case but that is all.

I am willing to accept the PR however from my point of view I first need to consider what the change is, why it is needed and if/how it will impact the thousands of other users who have never required this feature before. Also what happens if a user then requests that the state files or the entire config directory should be able to be symlinked. Rarely are changes to a project trivial.

One change that will certainly be needed is the commit message to better describe why the change is being made, without any vague 'this is bad' comments.

PhilHudson:
I've updated the commit message. I think it still needs work, but is it at least a bit better now?

PR #217, opened 2018-10-12
[UI|Daemon][#3081] Move baseargparser out of deluge/ui

DjLegolas:
This file is being used by both UI and Daemon. 
Therefor, it should not be located in the UI, but in the top-level `deluge` and renamed to `argparser_base`.
The same is with the function `set_dummy_trans`, which was located in `translation_util` and is now part of `argparser_base`.

### DIFF
--- a/deluge/argparser_base.py
+++ b/deluge/argparser_base.py
@@ -16,9 +16,29 @@ import platform
 import sys
 import textwrap
 
+from six.moves import builtins
+
 import deluge.log
 from deluge import common
 from deluge.configmanager import get_config_dir, set_config_dir
+
+log = logging.getLogger(__name__)
+log.addHandler(
+    logging.NullHandler()
+)  # Silence: No handlers could be found for logger "deluge.util.lang"
+
+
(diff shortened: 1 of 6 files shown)


cas--:
Hmm so I am reconsidering whether this change is worthwhile.

DjLegolas:
The question is why?
The pros for this change are obvious. What are the cons?

cas--:
The baseargparser code was created for UIs so it's more ui code than common code.

There is another way to look at it, should daemon_entry should be in ui? Is the command `deluged` a user-interface? ;)

DjLegolas:
Well, this is a very interesting point of view.
But the answer, in my opinion, is no. The `deluged` command is not a UI.
`deluged` is only here to let users start the daemon and customize it if needed... It's just like `apache2` web server. Running it will create the daemon, but it does not the UI.

cas--:
Thanks for PR and making me think about this more. I had a few other ideas that I wanted to integrate so merged that instead.

PR #221, opened 2018-10-22
[Core] Call wait_for_alert in a thread

DjLegolas:
This spawns a thread in alertmanager to call wait_for_alert in a thread.
This reduces latency to deluge responding to events.

Based on #175.

### DIFF
--- a/deluge/component.py
+++ b/deluge/component.py
@@ -64,6 +64,11 @@ class Component:
                    paused by instructing the :class:`ComponentRegistry` to pause
                    this Component.
 
+       **pause()** - This method is called when the component is being paused.
+
+       **resume()** - This method is called when the component resumes from a Paused
+                    state.
+
         **shutdown()** - This method is called when the client is exiting.  If the
                      Component is in a "Started" state when this is called, a
                      call to stop() will be issued prior to shutdown().
@@ -83,7 +88,7 @@ class Component:
         **Stopping** - The Component has had it's stop method called, but it hasn't
                     fully stopped yet.
 
-        **Paused** - The Component has had it's update timer stopped, but will
+        **Paused** - The Component has had its update timer stopped, but will
                     still be considered in a Started state.
 
     """
@@ -111,9 +116,8 @@ class Component:
             _ComponentRegistry.deregister(self)
 
     def _component_start_timer(self):
-        if hasattr(self, 'update'):
-            self._component_timer = LoopingCall(self.update)
-            self._component_timer.start(self._component_interval)
+        self._component_timer = LoopingCall(self.update)
+        self._component_timer.start(self._component_interval)
 
     def _component_start(self):
         def on_start(result):
@@ -129,13 +133,10 @@ class Component:
             return fail(result)
 
         if self._component_state == 'Stopped':
-            if hasattr(self, 'start'):
-                self._component_state = 'Starting'
-                d = deferLater(reactor, 0, self.start)
-                d.addCallbacks(on_start, on_start_fail)
-                self._component_starting_deferred = d
-            else:
-                d = maybeDeferred(on_start, None)
+            self._component_state = 'Starting'
+            d = deferLater(reactor, 0, self.start)
+            d.addCallbacks(on_start, on_start_fail)
+            self._component_starting_deferred = d
         elif self._component_state == 'Starting':
             return self._component_starting_deferred
         elif self._component_state == 'Started':
@@ -165,14 +166,11 @@ class Component:
             return result
 
         if self._component_state != 'Stopped' and self._component_state != 'Stopping':
-            if hasattr(self, 'stop'):
-                self._component_state = 'Stopping'
-                d = maybeDeferred(self.stop)
-                d.addCallback(on_stop)
-                d.addErrback(on_stop_fail)
-                self._component_stopping_deferred = d
-            else:
-                d = maybeDeferred(on_stop, None)
+            self._component_state = 'Stopping'
+            d = maybeDeferred(self.stop)
+            d.addCallback(on_stop)
+            d.addErrback(on_stop_fail)
+            self._component_stopping_deferred = d
 
         if self._component_state == 'Stopping':
             return self._component_stopping_deferred
@@ -182,13 +180,12 @@ class Component:
     def _component_pause(self):
         def on_pause(result):
             self._component_state = 'Paused'
+            if self._component_timer and self._component_timer.running:
+                self._component_timer.stop()
 
         if self._component_state == 'Started':
-            if self._component_timer and self._component_timer.running:
-                d = maybeDeferred(self._component_timer.stop)
-                d.addCallback(on_pause)
-            else:
-                d = succeed(None)
+            d = maybeDeferred(self.pause)
+            d.addCallback(on_pause)
         elif self._component_state == 'Paused':
             d = succeed(None)
         else:
@@ -205,9 +202,10 @@ class Component:
     def _component_resume(self):
         def on_resume(result):
             self._component_state = 'Started'
+            self._component_start_timer()
 
         if self._component_state == 'Paused':
-            d = maybeDeferred(self._component_start_timer)
+            d = maybeDeferred(self.resume)
             d.addCallback(on_resume)
         else:
             d = fail(
@@ -222,9 +220,7 @@ class Component:
 
     def _component_shutdown(self):
         def on_stop(result):
-            if hasattr(self, 'shutdown'):
-                return maybeDeferred(self.shutdown)
-            return succeed(None)
+            return maybeDeferred(self.shutdown)
 
         d = self._component_stop()
         d.addCallback(on_stop)
@@ -243,6 +239,12 @@ class Component:
         pass
 
     def shutdown(self):
+        pass
+
+    def pause(self):
+        pass
+
+    def resume(self):
         pass
 
 
--- a/deluge/core/alertmanager.py
+++ b/deluge/core/alertmanager.py
@@ -15,9 +15,10 @@ This should typically only be used by the Core. Plugins should utilize the
 
 """
 import logging
+import threading
 from types import SimpleNamespace
 
-from twisted.internet import reactor
+from twisted.internet import reactor, threads
 
 import deluge.component as component
 from deluge._libtorrent import lt
@@ -31,7 +32,7 @@ class AlertManager(component.Component):
 
     def __init__(self):
         log.debug('AlertManager init...')
-        component.Component.__init__(self, 'AlertManager', interval=0.3)
+        component.Component.__init__(self, 'AlertManager')
         self.session = component.get('Core').session
 
         # Increase the alert queue size so that alerts don't get lost.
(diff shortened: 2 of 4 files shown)


deluge/core/alertmanager.py
```diff
@@ -60,12 +62,39 @@ def update(self):
...
+
     def stop(self):
         for delayed_call in self.delayed_calls:
             if delayed_call.active():
                 delayed_call.cancel()
         self.delayed_calls = []
 
+    def wait_for_alert_in_thread(self):
+        while self._component_state not in ("Stopping", "Stopped"):
+            try:
```
aresch:
```suggestion
            if self.session.wait_for_alert(1000) is None:
                continue
```

aresch: We should crash if there is an exception raised here. Alerts from libtorrent are critical for Deluge to function.

deluge/core/alertmanager.py
```diff
@@ -60,12 +62,39 @@ def update(self):
...
+            except Exception:
+                log.exception("during wait_for_alert")
+            # wait_for_alert returns None on timeout
+            if alert is None:
+                continue
+            # If we don't pop alerts before the next loop, the wait will
+            # return immediately and we'll get into a tight spin. This can
+            # happen when AlertManager is paused. Optimize this by pre-checking
+            # if we would handle alerts.
+            if self._component_state != "Started":
```
aresch: I think we could do this better by overriding pause() and clearing a threading event[0] and likewise with resume() setting the threading event. Then just have a event.wait() call here to block until the component is resumed. This would require us to expose these methods in component since we currently don't do that.

[0] - https://docs.python.org/3/library/threading.html#event-objects

deluge/core/alertmanager.py
```diff
@@ -60,12 +62,39 @@ def update(self):
         self.delayed_calls = [dc for dc in self.delayed_calls if dc.active()]
         self.handle_alerts()
```
aresch: This should be removed.

deluge/core/alertmanager.py
DjLegolas: Done.

deluge/core/alertmanager.py
DjLegolas: Hmm, good idea. Done.

deluge/core/alertmanager.py
DjLegolas: Thought I removed it... Done.

deluge/component.py
```diff
@@ -192,6 +192,8 @@ def on_pause(result):
             if self._component_timer and self._component_timer.running:
                 d = maybeDeferred(self._component_timer.stop)
                 d.addCallback(on_pause)
+                if hasattr(self, 'pause'):
```
aresch: I think this should be similarly to how _component_stop() is done. In other words, we should call pause() and then add on_pause() as a callback which does a _component_timer.stop().

This way it's consistent with the other behaviour and the pause() resume() signatures do not need to include a result from a previous deferred.

deluge/component.py
```diff
@@ -214,6 +216,8 @@ def on_resume(result):
         if self._component_state == 'Paused':
             d = maybeDeferred(self._component_start_timer)
             d.addCallback(on_resume)
+            if hasattr(self, 'resume'):
```
aresch: Same comment as above with pause().

deluge/core/alertmanager.py
```diff
@@ -55,17 +56,42 @@ def __init__(self):
...
 
+    def pause(self):
+        self._event.clear()
+
+    def resume(self):
+        self._event.set()
+
+    def wait_for_alert_in_thread(self):
+        while self._component_state not in ('Stopping', 'Stopped'):
+            if self.session.wait_for_alert(1000) is None:
```
cas--: @aresch I saw you've used `set_alert_notify` in spritzle... better than timer perhaps?

deluge/component.py
```diff
@@ -250,6 +265,12 @@ def update(self):
     def shutdown(self):
         pass
 
+    def pause(self):
```
cas--: Need to update the `Component` docstring for these new methods

deluge/component.py
```diff
@@ -187,9 +187,16 @@ def on_stop_fail(result):
     def _component_pause(self):
         def on_pause(result):
             self._component_state = 'Paused'
+            if need_stop and self._component_timer and self._component_timer.running:
+                self._component_timer.stop()
 
+        need_stop = True
         if self._component_state == 'Started':
-            if self._component_timer and self._component_timer.running:
+            if hasattr(self, 'pause'):
```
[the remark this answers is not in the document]
cas--: In fact the pause method is defined in the class so the object would always have the `pause` attribute. I think the is an artefact of initial rewrite: b0714f625f50

deluge/component.py
```diff
@@ -210,10 +217,18 @@ def on_pause(result):
     def _component_resume(self):
         def on_resume(result):
             self._component_state = 'Started'
+            if need_start:
```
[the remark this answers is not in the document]
DjLegolas: I was thinking about preventing the next lines from being called in case we already called them...
For example, in case we are in `_component_resume` and need to start the timer. If the component has the attribute `resume`, we need to call `resume()` and to `on_resume()` as a callback, in which we will start the timer.
But, if the component don't this attribute, I call `_component_start_timer` and only than `on_resume()`.
In this situation, I'll be calling `_component_start_timer` twice. So I need to prevent it.

I'll change the signature of `on_resume` and `on_pause` so both will get this flag as parameter and not from the outer scope.

deluge/core/alertmanager.py
aresch: Yea, I would suggest using it but I'm not sure if it's in a released libtorrent. I think it would only be available in 1.2 with this commit: https://github.com/arvidn/libtorrent/commit/c2ea38fdfe4cc8f5bd9a9525f6d179ebcf4ab4d6#diff-87803408b3e578d01ff14ecd078cfb4c

We probably don't want to raise the minimum libtorrent dependency to 1.2 quite yet.

deluge/component.py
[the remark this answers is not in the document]
aresch: Yea, I don't think we need hasattr() anymore, you're right, we define the functions in the class now.

aresch:
This pull request looks good to me.  It would be nice to have some better tests for alert manager, as we don't cover much of what has changed. Feel free to follow-up with some expanded tests if you're interested :)

deluge/core/alertmanager.py
cas--: Hmm well it is in 1.1 so that binding should have be exposed there...

cas--: @aresch The `wait_for_alert` is now `1s` whereas the AlertManager was polling at `0.3s`, do you think this is perhaps too long?

DjLegolas:
Hmm, ok, I'll give it a try.
Correct me if I'm wrong, but to fully test `AlertManager`, I'll need to make `libtorrent` post some alerts and to handle them correctly.
One way is to mock it, and the other is to somehow force is to set an alert.
Which is the better way?

deluge/core/alertmanager.py
aresch: I don't think it's a problem since the only thing done in update() is the clean-up of delayed calls. I was actually going to propose removing that entirely but it was out of scope for this PR. I'll mention that idea now though.

We could wrap all handler functions with another function that removes the delayed call from the list on completion which would remove the need to run a clean-up on the list. We'd probably want to change delayed_calls to a dictionary to make this work efficiently.

```python
def run_handler(self, handler, alert):
    handler_func(alert)
    del self.delayed_calls[(handler, alert)]
```

In handle_alerts() we'd change the creation of the delayed call as such:
```python
for handler in self.handlers[alert_type]:
  self.delayed_calls[(handler, alert)] = reactor.callLater(0, self.run_handler, handler, alert)
```

Anyways, if we think this is a good idea we should do it in a separate PR.

aresch: > Hmm well it is in 1.1 so that binding should have be exposed there...

I don't see it in the RC_1_1 branch, I think it's only in 1.2.

cas--: I mean it is in C++ code for [RC_1_1](https://github.com/arvidn/libtorrent/blob/RC_1_1/include/libtorrent/session_handle.hpp#L960) so the binding should have been fixed there

aresch: Oh, yea, we'd need to add it to the bindings. I think that for now, this approach is fine and there isn't much advantage to switching to use set_alert_notify.

aresch:
> Hmm, ok, I'll give it a try.
> Correct me if I'm wrong, but to fully test `AlertManager`, I'll need to make `libtorrent` post some alerts and to handle them correctly.
> One way is to mock it, and the other is to somehow force is to set an alert.
> Which is the better way?

Yea, you'll want to mock out session.pop_alerts() and session.wait_for_alerts(). You can check out how we did it in spritzle for an idea: https://github.com/spritzle/spritzle/blob/master/spritzle/tests/test_alert.py

DjLegolas:
So I wrote some test units... What do you guys think @aresch @cas--?

deluge/tests/test_alertmanager.py
```diff
@@ -30,6 +55,31 @@ def handler(alert):
...
+
+    def test_pause_not_pop_alert(self):
+        def check_event(value):
+            event = self.am._event.isSet()
+            self.assertFalse(event)
+
+        def dummy_alert_1_handler(alert):
+            self.fail()
+
+        self.am.session = SessionMock()
```
aresch: I think this should probably be put in set_up().

deluge/tests/test_alertmanager.py
```diff
@@ -30,6 +55,31 @@ def handler(alert):
...
+        d = component.pause(['AlertManager'])
```
aresch: I'm worried about the timing here. I think the pause should happen and then alerts should be added to the SessionMock, then check to make sure an alert wasn't handled. The way its done right now, I think the alerts in the SessionMock could be handled before the pause occurs and the test isn't really testing anything since SessionMock.alerts == [].

deluge/tests/test_alertmanager.py
DjLegolas: I tried moving it to `set_up()`, but got errors on other tests because of it.
So currently I'm not changing it.

deluge/tests/test_alertmanager.py
DjLegolas: Reasonable point. Changed.

DjLegolas:
Any thoughts why this change cases `tox -e py3` to fail?
I tested it with python 3.5, 3.6, 3.7 and only with 3.7 all tests pass.

deluge/tests/test_alertmanager.py
aresch: This still doesn't address my original comments. Since you are setting self.am.session = SessionMock() before the pause happens, those alerts that are in the SessionMock.__init__ have already been processed. In other words, by the time you set the handler, there are no alerts to handle anyways so the test isn't actually testing what you expect.

What you need to do here is simulate alerts coming from libtorrent while AlertManager is paused will not trigger a handler to run, so I think SessionMock.alerts needs to be built _after_ the AlertManager is paused.

DjLegolas: 4 years later, done 😅

deluge/tests/test_alertmanager.py
DjLegolas: 4 years later, done 😅

DjLegolas:
Rebased and fixed last 2 changes requested by @aresch

deluge/tests/test_alertmanager.py
```diff
@@ -27,6 +56,34 @@ def handler(alert):
...
+
+        def check_alert(value):
+            event = self.am._event.isSet()
+            self.assertFalse(event)
+
+        def dummy_alert_1_handler(alert):
+            self.fail()
+
+        d = component.pause(['AlertManager'])
+        self.am.session.set_alerts()
```
aresch, line 83: Since component.pause() is asynchronous, wouldn't you need to call set_alerts() in a callback on the deferred? Otherwise you may be calling set_alerts() before the component has actually been paused.

gazpachoking: @aresch immediately shows up to comment on an updated 4 year old ticket. 😮

DjLegolas: @aresch this is what happens when one writes code at 2AM.. fixed.
@gazpachoking lol

cas--:
@DjLegolas I imagine that the segfault in tests is due to https://github.com/arvidn/libtorrent/issues/6437

DjLegolas:
Saving the segfault [job](https://github.com/deluge-torrent/deluge/runs/4963505515) for later use.
@cas-- how do you want to handle this? We could try to perform a deep-copy of the alerts

aresch (approved):
LGTM

cas--:
I am not sure this is stable, I just had it crash on me as soon as I tested the branch out

I am not willing to merge this while there is the potential for segfaults in alert handling.

cas--:
@DjLegolas I have merged the changes for component 0745c0eff85ea4e4a3646c454098503ed08c56ff and 25a2b113e290600043338e22b4f116cd2dbf2886 but I am finding that the alertmanager thread changes is increasing the propagation of segfaults due to the alert pointer going out of scope... I will investigate further on how we can workaround the issue.

cas--:
Thanks @DjLegolas 

I applied a refactor to remove the alert copy and avoid segfaults: https://github.com/deluge-torrent/deluge/commit/54d6f502319ea60de6cbe34f1c22e7eed367f1ee

I'm not sure what effect this will have for alert handling so might need to keep an eye on it and revisit if needed.

PR #226, opened 2018-11-12
gtk3

doadin:


cas--:
Ah okay, I was testing with x64.

As I said adding `GObject.threads_init` does nothing here. The change that made the difference is the removal of the `Gdk` `threads_init`, which has disabled threading. 

I did read the following so likely solution is to disable the use of Gtk/Gdk theading on Windows but perhaps it should removed entirely since it is deprecated in Gtk3.

> Using GTK+ from multiple threads on Windows doesn't work at all, and would require substantial work and complexity to fix.

doadin:
so far I have tested 3 VMs. working on a fourth. Trying to eliminate any environment issues. All builds I have tested using appveyor gtk build.
1. win7 ult x64 with x86 and x64 python = threading issue(my build ive used for over a year now)
2. fresh win7 ult x86 with python x86 = threading issue
3. fresh win10 enterprise x64 with python x86 = NO threading issue
4. fresh win7 ult x64 with python x86 = in testing

Perhapse with a bit more testing we can maybe just exclude win7 in threading and leave for others. Im willing to test xp and 8.1 if this is a route to be considered.

cas--:
Those are interesting results.

It's really not an issue to disable threading, just need to be aware that this has been done and that in future if there are any UI lockups when running certain code then that code will need to be run in a thread.

requirements.txt
```diff
@@ -7,7 +7,7 @@ mako
 chardet
 six
 setproctitle
-pywin32; sys_platform == 'win32'
+pypiwin32; sys_platform == 'win32'
```
cas--: Please provide details in this commit message as to what if being fixed by this change. From my perspective both provide pywin32 and I havn't seen any problems.

doadin: In further testing this appears to be a issue with modules getting installed with the setup.py script(at least during a "python setup.py develop") After this setup pywin32 is "installed" but doesn't work, when going through the uninstall process you can see theres a big difference in files. and since it is "installed" following up with a pip install -r requirements.txt just says its already installed. However if you uninstall pywin32 then install with pip -r... it installs properly.

cas--:
Thanks for the suggested changes.

pywin32 is the official and latest version so this is not going to be changed.

PR #228, opened 2018-12-10
[UI|Console] added console option to specify move completed folder

hukal:
This option helps to split items to different folders. Useful for Plex media server for example.
Would be nice to see in ubuntu ppa.

### DIFF
--- a/deluge/tests/test_ui_console_add.py
+++ b/deluge/tests/test_ui_console_add.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Deluge and is licensed under GNU General Public License 3.0, or later, with
+# the additional special exception to link portions of this program with the OpenSSL library.
+# See LICENSE for more details.
+#
+
+from __future__ import unicode_literals
+
+import argparse
+
+from deluge.tests.test_ui_entry import ConsoleUIBaseTestCase
+from deluge.ui.console.cmdline.commands.add import Command
+
+
+class UICommonTestCase(ConsoleUIBaseTestCase):
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def test_add_args_short_option_name(self):
+        completed_path = 'completed_path'
+        parser = argparse.ArgumentParser()
+        cmd = Command()
+        cmd.add_arguments(parser)
+        args = parser.parse_args(['torrent', '-c', completed_path])
+        self.assertEqual(args.completed, completed_path)
+
+    def test_add_args_long_option_name(self):
+        completed_path = 'completed_path'
+        parser = argparse.ArgumentParser()
+        cmd = Command()
+        cmd.add_arguments(parser)
+        args = parser.parse_args(['torrent', '--completed', completed_path])
+        self.assertEqual(args.completed, completed_path)
--- a/deluge/tests/test_ui_console_add.py
+++ b/deluge/tests/test_ui_console_add.py
@@ -9,11 +9,12 @@ from __future__ import unicode_literals
 
 import argparse
 
-from deluge.tests.test_ui_entry import ConsoleUIBaseTestCase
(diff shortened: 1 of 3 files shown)


DjLegolas:
Can you fix the failed test (regarding coding style)?
In addition, please try to add a test for this.

DjLegolas:
I'll start with the failed test: The current coding style is new and is being validated with a dedicated test.
It is something @cas-- added some time ago.
The best way to check this, is to install `tox` and run `tox -e lint`. It will run the same test and, IIRC, modify the file as needed (note: don't run only once, but after each fix).

About adding new test, you are correct.. There are no tests for the console ui... So just start one from scratch, and test only what you added for now.
You can ask here for help.

And look at the [running tests](https://deluge.readthedocs.io/en/develop/contributing/testing.html) docs.

Finally, you closed this PR 😆

DjLegolas:
A little update, there are tests for the Console.
Look at the `test_ui_entry.py` file. In it there are most of the UI entry tests, including `ConsoleUIBaseTestCase` class, which tests some the Console's functionality.

deluge/tests/test_ui_console_add.py
```diff
@@ -0,0 +1,37 @@
...
+
+from __future__ import unicode_literals
+
+import argparse
+
+from deluge.tests.test_ui_entry import ConsoleUIBaseTestCase
+from deluge.ui.console.cmdline.commands.add import Command
+
+
+class UICommonTestCase(ConsoleUIBaseTestCase):
```
DjLegolas: should inherit from `basetest.BaseTestCase` as well for the tests to be run.

hukal:
Hi @DjLegolas ,
I've fixed inheritance and naming.
But it looks like I can't find a proper way to make an end to end test.

Things I have tried:
1. Directly call `handle` method of `Add` command.
In this case I'm getting _"ConsoleUI component is not registered"_ error.
Event if I register it (mock?) I can't do anything since `Add` command requires daemon.

2. In `test_ui_entry.py` we have `test_console_command_status` which works with daemon.
I was able to make a similar one to test 'add' command but this command returns only "Torrent added" without any id/hash. Due to this fact I can't validate the file in the way it validated in `test_add_torrent_file` from `test_core`.

3. In addition to  point 2 I've tried to run `status` command right after `add` in one method. (I was hoping to grep hash from there). It also caused errors. (Am I correct that current test structure supports 1 console call per method?)

Could you please advise how to test it?

Thanks

deluge/tests/test_ui_console_add.py
hukal: Done.

DjLegolas:
Ok, I managed to run the test by adding a call to `status` command between `add` and `manage` commands. In addition, I added a test for the `add` command itself.
Changes pushed it to @hukal fork.
I'm sure it's not the best way to handle this, but for now it will do. I hope 🤣

hukal:
Thanks @DjLegolas !
I've checked the tests locally - all green.
I'm only getting a warning: 
`common.py:270: PendingDeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5`
My local env is on Python 3.6.7

hukal:
@DjLegolas Build error, again.
`INTERNALERROR> PluginValidationError: unknown hook 'pytest_namespace' in plugin <module 'pytest_twisted' from 'c:\projects\deluge\.tox\py27\lib\site-packages\pytest_twisted.py'>`
Is it because of my changes?

hukal:
OK, let's wait. Thanks!

DjLegolas:
@cas-- What do you say? Want to merge this?

cas--:
It looks good, I will get around to reviewing and merging :+1:

cas--:
Thanks for this and the test :+1: 

I have made some changes (-m instead of -c) and merged it: f885edd7fc85f

PR #229, opened 2019-01-21
[TESTS] Force usage of pytest 4.0.2 only

DjLegolas:
As of `pytest` 4.1.0, there is an issue with `pytest-twisted` 1.8.
Until `pytest-twisted` releases a new version (fixed in git), we should use the previous version of `pytest`.


DjLegolas:
Never mind, it was released today.

PR #232, opened 2019-03-26
Use gzip content encoding only if supported by client

tkSimon:
At the moment deluge responds to requests for certain resource with gzip content encoding regardless of the requests `accept-encoding` header. It should only use gzip encoding if supported by the client.

To reproduce, run:
curl 127.0.0.1:8112/js/ext-extensions.js -o /dev/null -v

The response will have a "Content-Encoding: gzip" header, and it's payload will be gzipped even though the request doesn't contain an "Accpet-Encoding: gzip" header.

### DIFF
--- a/deluge/tests/test_httpdownloader.py
+++ b/deluge/tests/test_httpdownloader.py
@@ -236,7 +236,7 @@ class DownloadFileTestCase(unittest.TestCase):
         def cb(result):
             print(result)
 
-        d.addCallback(self.assertNotContains, b'fail', file_mode='rb')
+        d.addCallback(self.assertContains, b'fail')
         return d
 
     def test_page_redirect_unhandled(self):
--- a/deluge/ui/web/common.py
+++ b/deluge/ui/web/common.py
@@ -10,10 +10,12 @@
 from __future__ import unicode_literals
 
 import gettext
+import re
 import zlib
 
 from deluge.common import PY2, get_version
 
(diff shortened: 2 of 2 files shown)


cas--:
I doubt I will be accept this, it adds extra code to maintain. It might not be quite protocol correct to always return compressed but we want it enabled and it has been documented that gzip is used for responses. I don't see any pressing need for it otherwise?

FYI browsers always send the gzip header ([1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#The_Accept-Encoding_header))

tkSimon:
hmm, sorry to hear that. The code is taken directly from twisted: https://github.com/twisted/twisted/blob/29cbea324e1b1bb1d38863fbc0ba640d4d218177/src/twisted/web/server.py#L584

Also, the test I updated is passing in it's current form (so without my commits,) but technically it shouldn't.

tkSimon:
Personally, I wanted deluge to respond with uncompressed assets because i have an nginx reverse proxy in front of deluge that's doing brotli (and gzip compression,) and caching those. The caveat issue being that nginx only compresses when the origin (deluge) sends them uncompressed. This is usually achieved by configuring nginx to not forward the accept-encoding header.

cas--:
OK thanks for the further info, I'll have a think

cas--:
Closing for #233

PR #233, opened 2019-04-01
[#3244] [Web] Add support for accept-encoding header

cas--:
Trac ticket: https://dev.deluge-torrent.org/ticket/3244

Use EncodingResourceWrapper to replace compress function so that the proper checks for accept-encoding header are made

Also ensure only text if compressed and images are uncompressed.

- [ ] Update commit message with change details.
- [x] Remove compress function and replace usage in other code.

### DIFF
--- a/deluge/tests/common_web.py
+++ b/deluge/tests/common_web.py
@@ -67,10 +67,3 @@ class WebServerMockBase(object):
             pass
 
         self.patch(auth, 'check_request', check_request)
-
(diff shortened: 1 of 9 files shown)


gtbuchanan:
Thank you for considering this. I use IIS as a reverse proxy and the URL Rewrite module refuses to rewrite already encoded responses so I've just been dealing with the fact some of the Deluge responses don't come through successfully.

PR #234, opened 2019-04-11
Replace platform.linux_distribution function

DjLegolas:
Will close https://dev.deluge-torrent.org/ticket/3247

As of python 3.5, this function is marked as deprecated.
So, distro[1] is the one we will use (this package is listed at the
example package in the python's docs[2]).

[1] https://pypi.org/project/distro/
[2] https://docs.python.org/3/library/platform.html#platform.dist

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -46,9 +46,6 @@ except ImportError:
     from urlparse import urljoin  # pylint: disable=ungrouped-imports
     from urllib import pathname2url, unquote_plus  # pylint: disable=ungrouped-imports
 
-if platform.system() == 'Linux':
-    import distro
-
 # Windows workaround for HTTPS requests requiring certificate authority bundle.
 # see: https://twistedmatrix.com/trac/ticket/9209
 if platform.system() in ('Windows', 'Microsoft'):
@@ -56,12 +53,17 @@ if platform.system() in ('Windows', 'Microsoft'):
 
     os.environ['SSL_CERT_FILE'] = where()
 
-# gi makes dbus available on Window but don't import it as unused.
+
 if platform.system() not in ('Windows', 'Microsoft', 'Darwin'):
+    # gi makes dbus available on Window but don't import it as unused.
     try:
         import dbus
     except ImportError:
         dbus = None
+    try:
+        import distro
+    except ImportError:
+        distro = None
 
 log = logging.getLogger(__name__)
 
@@ -269,8 +271,8 @@ def get_os_version():
     elif osx_check():
         os_version = list(platform.mac_ver())
         os_version[1] = ''  # versioninfo always empty.
-    elif linux_check():
-        os_version = distro.linux_distribution(full_distribution_name=False)
+    elif distro:
+        os_version = distro.linux_distribution()
     else:
         os_version = (platform.release(),)
 
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -46,6 +46,9 @@ except ImportError:
     from urlparse import urljoin  # pylint: disable=ungrouped-imports
     from urllib import pathname2url, unquote_plus  # pylint: disable=ungrouped-imports
 
+if platform.system() == 'Linux':
+    import distro
+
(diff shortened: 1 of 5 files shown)


deluge/common.py
```diff
@@ -267,7 +270,7 @@ def get_os_version():
         os_version = list(platform.mac_ver())
         os_version[1] = ''  # versioninfo always empty.
     elif linux_check():
-        os_version = platform.linux_distribution()
+        os_version = distro.linux_distribution(full_distribution_name=False)
```
cas--: `full_distribution_name` should be True to match original line.

deluge/common.py
```diff
@@ -46,6 +46,9 @@
     from urlparse import urljoin  # pylint: disable=ungrouped-imports
     from urllib import pathname2url, unquote_plus  # pylint: disable=ungrouped-imports
 
+if platform.system() == 'Linux':
+    import distro
```
cas--: Since the distro details are not that important this would be better as optional package

deluge/common.py
```diff
@@ -267,7 +270,7 @@ def get_os_version():
         os_version = list(platform.mac_ver())
         os_version[1] = ''  # versioninfo always empty.
     elif linux_check():
```
cas--: Change this to test for `disto` or `linux_distribution` perhaps?

deluge/common.py
```diff
@@ -46,6 +46,9 @@
     from urlparse import urljoin  # pylint: disable=ungrouped-imports
     from urllib import pathname2url, unquote_plus  # pylint: disable=ungrouped-imports
 
+if platform.system() == 'Linux':
```
cas--: Technically `distro` supports BSD too so perhaps just check not Windows or Darwin?

deluge/common.py
DjLegolas: How do I do that? Isn't it by stating in the requires section in `setup.py` (which I forgot to add)?

cas--: This will importerror if distro is not installed, have a look at dbus below for example

DjLegolas: Oh, OK.
But now I got a question... Should I add `distro` to `requirements.txt` (or any other file of them) and to `install_requires` in `setup.py`?

cas--: > Should I add `distro` to `requirements.txt` (or any other file of them) and to `install_requires` in `setup.py`?

Yes add to both. `install_requires` list is automatically installed by pip and requirements.txt is for manual environment setup with `pip -r`. 

I have put optional packages in `install_requires` because they are soft-required for best user-experience and would be a headache to put in `extra_require` and ensure they are installed. 

My reasoning for making `distro` package optional is to help with packaging. For example there is no deb package on Ubuntu so Deluge would break if it's missing.

cas--: Oh don't forget to update the DEPENDS file too.

requirements.txt
```diff
@@ -11,3 +11,4 @@ pywin32; sys_platform == 'win32'
 py2-ipaddress; sys_platform == 'win32' and python_version == '2'
 certifi; sys_platform == 'win32'
 zope.interface>=4.4.2
+distro; platform_system != 'Windows' and platform_system != 'Microsoft' and platform_system != 'Darwin'
```
DjLegolas: Not sure if this condition is good.
What do you think @cas-- ?

cas--: I would simplify to `platform_system != 'Windows' and platform_system != 'Darwin'`. I don't think the Microsoft check is required these days.

I do wonder what happens if try to install distro on those OS but it's fine with the check for now.

An amusing alternative would be `'win' not in sys_platform` since it would cover `cygwin`, `win32` and `darwin` but a bit too obscure ;)

DjLegolas: I installed `distro` on windows successfully, but all the functions return empty string to something like it.

And I changed the line to the obscure way :wink:

cas--:
Thanks, merged: 2296906

PR #235, opened 2019-04-17
Remove detox

DjLegolas:
As of `tox` 3.7, `detox` is no longer needed as `tox` added a native parallel environment execution.


cas--:
Merged, thanks!

PR #237, opened 2019-05-09
adding intellij idea files to .gitignore

abecadel:


cas--:
Thanks for the PR but I'm afraid this is not really specific to deluge development (i.e. editor-agnostic) and those files can be ignored via a user specific file: https://stackoverflow.com/q/5724455/175584

PR #240, opened 2019-05-27
[WiP] Update to libtorrent-1.2

DjLegolas:
Update libtorrent to version >= 1.2.
Info about the upgrade:
https://github.com/arvidn/libtorrent/blob/RC_1_2/docs/upgrade_to_1.2.rst

TODOs (may change):
- [ ] Change test environments to lt-1.2
- [ ] Fix usage and handling of resume-data
- [ ] Alerts

### DIFF
--- a/deluge/_libtorrent.py
+++ b/deluge/_libtorrent.py
@@ -24,8 +24,8 @@ try:
 except ImportError:
     import libtorrent as lt
 
-REQUIRED_VERSION = '1.1.2.0'
-LT_VERSION = lt.__version__
+REQUIRED_VERSION = '1.2.0.0'
+LT_VERSION = lt.version
 
 if VersionSplit(LT_VERSION) < VersionSplit(REQUIRED_VERSION):
     raise ImportError(
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -45,10 +45,10 @@ from deluge.event import (
 log = logging.getLogger(__name__)
 
 LT_DEFAULT_ADD_TORRENT_FLAGS = (
-    lt.add_torrent_params_flags_t.flag_paused
-    | lt.add_torrent_params_flags_t.flag_auto_managed
-    | lt.add_torrent_params_flags_t.flag_update_subscribe
-    | lt.add_torrent_params_flags_t.flag_apply_ip_filter
+    lt.torrent_flags.paused
+    | lt.torrent_flags.auto_managed
+    | lt.torrent_flags.update_subscribe
+    | lt.torrent_flags.apply_ip_filter
 )
 
 
@@ -361,11 +361,11 @@ class TorrentManager(component.Component):
         add_torrent_params['flags'] = (
             (
                 LT_DEFAULT_ADD_TORRENT_FLAGS
-                | lt.add_torrent_params_flags_t.flag_duplicate_is_error
-                | lt.add_torrent_params_flags_t.flag_upload_mode
+                | lt.torrent_flags.duplicate_is_error
+                | lt.torrent_flags.upload_mode
             )
-            ^ lt.add_torrent_params_flags_t.flag_auto_managed
-            ^ lt.add_torrent_params_flags_t.flag_paused
+            ^ lt.torrent_flags.auto_managed
+            ^ lt.torrent_flags.paused
         )
 
         torrent_handle = self.session.add_torrent(add_torrent_params)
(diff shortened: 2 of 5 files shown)


deluge/_libtorrent.py
```diff
@@ -24,8 +24,8 @@
 except ImportError:
     import libtorrent as lt
 
-REQUIRED_VERSION = '1.1.2.0'
-LT_VERSION = lt.__version__
+REQUIRED_VERSION = '1.2.0.0'
+LT_VERSION = lt.version
```
cas--: This should not be changed, the de-facto standard for getting Python lib version is `__version__`

deluge/core/torrentmanager.py
```diff
@@ -45,10 +45,10 @@
...
 
 LT_DEFAULT_ADD_TORRENT_FLAGS = (
-    lt.add_torrent_params_flags_t.flag_paused
-    | lt.add_torrent_params_flags_t.flag_auto_managed
-    | lt.add_torrent_params_flags_t.flag_update_subscribe
-    | lt.add_torrent_params_flags_t.flag_apply_ip_filter
+    lt.torrent_flags.paused
+    | lt.torrent_flags.auto_managed
+    | lt.torrent_flags.update_subscribe
+    | lt.torrent_flags.apply_ip_filter
```
cas--: Using `lt.torrent_flags` is a breaking change for 1.1, lt 1.2 with deprecated enabled should still have `add_torrent_params_flags_t`

deluge/_libtorrent.py
```diff
@@ -24,8 +24,8 @@
 except ImportError:
     import libtorrent as lt
 
-REQUIRED_VERSION = '1.1.2.0'
-LT_VERSION = lt.__version__
+REQUIRED_VERSION = '1.2.0.0'
```
cas--: I don't want to rely on 1.2 solely yet

deluge/tests/test_core.py
```diff
@@ -116,7 +117,13 @@ def on_shutdown(result):
     def add_torrent(self, filename, paused=False):
         if not paused:
             # Patch libtorrent flags starting torrents paused
-            self.patch(deluge.core.torrentmanager, 'LT_DEFAULT_ADD_TORRENT_FLAGS', 592)
+            self.patch(
+                deluge.core.torrentmanager,
+                'LT_DEFAULT_ADD_TORRENT_FLAGS',
+                lt.torrent_flags.auto_managed
+                | lt.torrent_flags.update_subscribe
+                | lt.torrent_flags.apply_ip_filter,
```
cas--: I'm sure that using the value is intentional to avoid relying on `libtorrent` in a test. A comment would suffice or create variables with those values.

deluge/tests/test_torrent.py
```diff
@@ -73,12 +73,14 @@ def get_torrent_atp(self, filename):
         filename = common.get_test_data_file(filename)
         with open(filename, 'rb') as _file:
             info = lt.torrent_info(lt.bdecode(_file.read()))
-        atp = {'ti': info}
-        atp['save_path'] = os.getcwd()
-        atp['storage_mode'] = lt.storage_mode_t.storage_mode_sparse
-        atp['add_paused'] = False
-        atp['auto_managed'] = True
-        atp['duplicate_is_error'] = True
+        atp = lt.add_torrent_params()
```
cas--: These changes are breaking for 1.1, please revert

deluge/tests/test_core.py
DjLegolas: The problem with this hard-coded value is the change of values in lt-1.2.
For testing this properly, without relying on `libtorrent`, we should set some environment flag to know which version of it we are using and set the proper value accordingly.

Do you have any suggestions for how to do this?

DjLegolas:
> I don't want to make it 1.2-only yet so can this PR be just those changes to fix running on 1.2 (with deprecated enabled)

Ok, no problem.
For now the code will look a little silly as it will check in the relevant places with which version of `libtorrent` we currently run.

The only test I still didn't managed to resolve with 1.2 is `test_torrent_error_resume_data_unaltered()`.
The `state` of the torrent is `Downloading` while the test expects `Error`.

deluge/tests/test_core.py
cas--: Ah ok, then import and use the old-style lt flags.

deluge/tests/test_torrent.py
```diff
@@ -73,12 +73,24 @@ def get_torrent_atp(self, filename):
...
-        atp['storage_mode'] = lt.storage_mode_t.storage_mode_sparse
-        atp['add_paused'] = False
-        atp['auto_managed'] = True
-        atp['duplicate_is_error'] = True
+        if lt.is_version_1_1:
+            atp = {
+                'ti': info,
+                'save_path': os.getcwd(),
+                'storage_mode': lt.storage_mode_t.storage_mode_sparse,
+                'add_paused': False,
```
cas--: Instead of testing for 1.2. Replace these three dict with the following that should work for both versions:

```
atp['flags']  = (
    LT_DEFAULT_ADD_TORRENT_FLAGS  
    | lt.add_torrent_params_flags_t.flag_duplicate_is_error
    & ~lt.add_torrent_params_flags_t.flag_paused
)
```

deluge/tests/test_core.py
```diff
@@ -116,7 +117,11 @@ def on_shutdown(result):
     def add_torrent(self, filename, paused=False):
         if not paused:
             # Patch libtorrent flags starting torrents paused
-            self.patch(deluge.core.torrentmanager, 'LT_DEFAULT_ADD_TORRENT_FLAGS', 592)
+            self.patch(
+                deluge.core.torrentmanager,
+                'LT_DEFAULT_ADD_TORRENT_FLAGS',
+                592 if lt.is_version_1_1 else 168,
```
cas--: Replace with boolean flag operation that would work for both versions

cas--:
> The only test I still didn't managed to resolve with 1.2 is `test_torrent_error_resume_data_unaltered()`.
> The `state` of the torrent is `Downloading` while the test expects `Error`.

That could be a problem where the test is leaving files behind that are being found on next run. I have a branch with a fix for that. Could be a file in `~/Downloads'?

DjLegolas:
@cas-- can you add lt-1.2 binding to deluge's develop ppa?

in addition, there are no files in ~/Downloads no ~/downloads.

PR #241, opened 2019-05-30
fix blurry icons in gnome

isac322:
Fixed blurry icon by changing `StartupWMClass` value from `"Deluge"` to `"deluge"` in `deluge.desktop.in`.

### `StartupWMClass="Deluge"`

image

### `StartupWMClass="deluge"`

image


cas--:
Merged thanks!

PR #242, opened 2019-06-01
[WebUI] Add option to copy magnet URI

Gargaj:
(I didn't make an icon, I'm sure some pixel wizard will be happy to.)

### DIFF
--- a/deluge/ui/web/js/deluge-all/CopyMagnetURIWindow.js
+++ b/deluge/ui/web/js/deluge-all/CopyMagnetURIWindow.js
@@ -1,0 +1,89 @@
+/*
+ * Deluge.CopyMagnetURI.js
+ * 
+ * Copyright (c) Damien Churchill 2009-2010 <damoxc@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


matiasw:
I have a candidate for the icon drawn; but how do I get the code to load it? It seems there is already code in place (the "iconCls: "x-deluge-copy-magnet-uri-icon""), but what is the search path? What formats are accepted?

I've attached a PNG rendering of the icon. The original is in SVG format. Is SVG usable in the WebUI as is? Browsers today can render it. Or does it need to be a bitmap? If so, what should the dimensions be?

x-deluge-copy-magnet-uri-icon

cas--:
@matiasw The `IconCls` classes and icons paths are specified in `deluge/ui/web/css`. I would prefer if we have both png and svg stored in deluge/ui/data/pixmaps. We can then generate different sized png icons as required and copy to `deluge/ui/web/icons`.

matiasw:
@cas-- Thanks - the original SVG is at https://github.com/matiasw/deluge/blob/magnet-copy-icon/deluge/ui/data/pixmaps/copy-magnet.svg

I tried to make the UI load it, but couldn't quite make it work. What I did was added a section matching the one in `web/js/deluge-all/CopyMagnetURIWindow.js` to `deluge/ui/web/css/deluge.css`, but that doesn't seem to be enough.

matiasw:
Ok, there is a working branch that loads the icon file for this new Copy Magnet URI menu item, as well as the currently hidden & disabled "Add magnet" (aka. "Infohash") icon in the Add dialog: https://github.com/matiasw/deluge/tree/magnet-copy-icon/deluge

cas--:
Thanks! Merged in c38f913948ed

PR #247, opened 2019-06-19
Fallback functionality for dbus file manager

int3l:
Fallback to the xdg-open option in case, when D-Bus is failing to give us File Manager.
Fixes: https://dev.deluge-torrent.org/ticket/3272


cas--:
Thanks, merged: 833b5a1f306dad

PR #249, opened 2019-06-22
[common] fix creation of pidfile

DjLegolas:
https://dev.deluge-torrent.org/ticket/3278

### DIFF
--- a/deluge/argparserbase.py
+++ b/deluge/argparserbase.py
@@ -325,7 +325,7 @@ class ArgParserBase(argparse.ArgumentParser):
 
             # Write pid file before chuid
             if options.pidfile:
-                with open(options.pidfile, 'wb') as _file:
+                with open(options.pidfile, 'w') as _file:
                     _file.write('%d\n' % os.getpid())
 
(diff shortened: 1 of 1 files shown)


deluge/argparserbase.py
```diff
@@ -326,7 +326,7 @@ def _handle_ui_options(self, options):
             # Write pid file before chuid
             if options.pidfile:
                 with open(options.pidfile, 'wb') as _file:
-                    _file.write('%d\n' % os.getpid())
+                    _file.write(b'%d\n' % os.getpid())
```
cas--: Actually the pidfile should be opened in text mode so revert this and change `wb` to `w` above

cas--:
The commit title should relate to what was fixed not how it was fixed. i.e. fix creating a pidfile

https://dev.deluge-torrent.org/wiki/GitTips#CommitMessages

DjLegolas:
Fixed both commit message and code.

cas--:
Thanks merged: 1b4ac88

PR #250, opened 2019-06-24
[Plugin] Fix argument type for the notification popup hint

int3l:
As per the documentation, since Notify >= 0.6, Notification.set_hint method gets string key and GLib.Variant (instead of string) value.
Ref: http://lazka.github.io/pgi-docs/#Notify-0.7/classes/Notification.html#Notify.Notification.set_hint
http://lazka.github.io/pgi-docs/GLib-2.0/classes/Variant.html#GLib.Variant

Fixes: [#3267](https://dev.deluge-torrent.org/ticket/3267)


cas--:
Thanks I already had a branch with this fix which has been merged: 63a4301a8be8

PR #251, opened 2019-06-24
Add pygeoip support

cas--:
https://dev.deluge-torrent.org/ticket/3271

### DIFF
--- a/DEPENDS.md
+++ b/DEPENDS.md
@@ -41,7 +41,7 @@ All modules will require the [common](#common) section dependencies.
 ## Core (deluged daemon)
 
 - [libtorrent] _>= 1.1.1_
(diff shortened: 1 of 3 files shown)


paravoid:
Not a great idea - this is a deprecated module (archived on GitHub by upstream), as is the existing GeoIP module. They are both libraries for the GeoIP Legacy database formats, which is now deprecated (data is not shipped in this format by MaxMind).

The right target would be MaxMind's [Python GeoIP2 library](https://github.com/maxmind/GeoIP2-python), or the lower level [MaxMindDB DB Reader Python implementation](https://github.com/maxmind/GeoIP2-python).

cas--:
Closed via 65e5010e7

PR #254, opened 2019-07-04
[GTK] Fix ordering in status tab

thatguystone:
The {download,upload}-{total,speed} columns were a bit mixed up. This sets them right.


stefantalpalaru:
Patch works as advertised. Added to my Gentoo overlay.

PR #259, opened 2019-07-20
Ctl+Q to quit Deluge GTK without killing daemon

BackSlasher:
I'm missing this shortcut, and would like to avoid killing the daemon when quitting the UI


cas--:
Thanks, merged

PR #262, opened 2019-08-14
Update systemd-service.md

eN0Rm:
-d should be -f I believe


cas--:
The docs are for deluge 2.0 where this is the new way of daemonizing the webui, same as deluged.

PR #263, opened 2019-09-24
Fix privilege dropping

devplayer0:
`os.setgid()` should be called to set the GID, and it should be called before `os.setuid()` to prevent reinstatement of privileges.


cas--:
Thanks for spotting this, merged!

PR #265, opened 2019-10-11
Fix status tab labels being reversed

tsegers:
Swap "Uploaded" and "Down speed" label positions in Glade layout

Fixes: #3281


tsegers:
Closing since this is a duplicate of #254

PR #266, opened 2019-10-29
Set pickle.load() to use encoding='utf-8'

myakove:
use  encoding='utf-8' when loading state file

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -809,7 +809,7 @@ class TorrentManager(component.Component):
 
             try:
                 with open(filepath, 'rb') as _file:
-                    state = pickle.load(_file)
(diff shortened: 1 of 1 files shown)


myakove:
After upgrade deluge-daemon to 2.0.3-2 it fails to load with error:
state = pickle.load(_file)
builtins.UnicodeDecodeError: 'ascii' codec can't decode byte 0xd7 in position 44: ordinal not in range(128)

when trying to read torrents.state and torrents.state.bak

Force encoding='utf-8' fix it.

cas--:
Thanks for the pull request. I have pushed a different fix 23b3f144fce3 instead as needed a test to be sure of the issue and the fix to be compatible with Python 2 

Trac ticket for reference https://dev.deluge-torrent.org/ticket/3298

PR #267, opened 2019-11-03
Fixes potential "dictionary changed size during iteration" on shutdown

JohnDoee:
See https://dev.deluge-torrent.org/ticket/3317 for description.


cas--:
Merged ed4bc5fa17ec5

PR #268, opened 2019-11-03
fixing encoding issue with deluge-console config --set

JohnDoee:
Config settings are not decoded correctly, this error can be reproduced with a command like `deluge-console -c /config/ "config --set download_location /downloads"`

```
'str' object has no attribute 'decode'
14:52:04 [ERROR   ][deluge.ui.console.cmdline.command                     :138 ] 'str' object has no attribute 'decode'
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/command.py", line 135, in exec_command
    ret = self._commands[options.command].handle(options)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 104, in handle
    return self._set_config(options)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 143, in _set_config
    val = simple_eval(val)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 72, in simple_eval
    res = atom(src, next(src))
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 57, in atom
    return token[-1].decode('string-escape')
AttributeError: 'str' object has no attribute 'decode'
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/command.py", line 135, in exec_command
    ret = self._commands[options.command].handle(options)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 104, in handle
    return self._set_config(options)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 143, in _set_config
    val = simple_eval(val)
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 72, in simple_eval
    res = atom(src, next(src))
  File "/usr/lib/python3/dist-packages/deluge/ui/console/cmdline/commands/config.py", line 57, in atom
    return token[-1].decode('string-escape')
AttributeError: 'str' object has no attribute 'decode'
```

### DIFF
--- a/deluge/ui/console/cmdline/commands/config.py
+++ b/deluge/ui/console/cmdline/commands/config.py
@@ -41,7 +41,7 @@ def atom(src, token):
             else:
                 if token[1].startswith('0x'):
                     # Hex number so return unconverted as string.
-                    return token[1].decode('string-escape')
+                    return token[1].encode('utf-8').decode('unicode-escape')
                 else:
                     return int(token[1], 0)
         except ValueError:
@@ -54,11 +54,11 @@ def atom(src, token):
     elif token[1].lower() == 'false':
         return False
     elif token[0] is tokenize.STRING or token[1] == '/':
-        return token[-1].decode('string-escape')
+        return token[-1].encode('utf-8').decode('unicode-escape')
     elif token[1].isalpha():
         # Parse Windows paths e.g. 'C:\\xyz' or 'C:/xyz'.
         if next()[1] == ':' and next()[1] in '\\/':
-            return token[-1].decode('string-escape')
+            return token[-1].encode('utf-8').decode('unicode-escape')
 
     raise SyntaxError('malformed expression (%s)' % token[1])
 


JohnDoee:
Hm, don't think this is the only broken part, seems like the whole console setting stuff needs some real testing. Can't get --set listen_ports to work either

deluge/ui/console/cmdline/commands/config.py
```diff
@@ -54,11 +54,11 @@ def atom(src, token):
     elif token[1].lower() == 'false':
         return False
     elif token[0] is tokenize.STRING or token[1] == '/':
-        return token[-1].decode('string-escape')
+        return token[-1].encode('utf-8').decode('unicode-escape')
```
cas--: This replacement does not feel correct so I did some research, see https://stackoverflow.com/a/24519338/175584

My conclusion is that we remove the decode with string-escape as I can't see which escapes (newline, tab, etc.) would actually be needed for console commmands. Do you agree?

However if later on someone feels that escaping is useful then it can be revisited with concrete examples and unit tests using the regex function from that SO answer.

JohnDoee: I found the way the functionality a bit weird overall but wanted to stay true to the original design and just make it work in Python 3.

Yes, I agree that it can just be removed for now, not sure how widely used it is anyways and newlines and such in a config setting isn't common.

cas--:
Closing for #270

@JohnDoee I decided that the old code was not fit for purpose, feel free to comment on my alternative PR

PR #269, opened 2019-11-07
deluge-web has no option -d

jandetlefsen:
fixed an error that threw me off while following the instructions on https://deluge.readthedocs.io/en/latest/how-to/systemd-service.html
removed option -d and it works as expected


cas--:
The option was change in 2.0 so it is correct for the docs

jandetlefsen:
Oh i see thanks. Some sort of hint would help that there are no docs for the current stable or at least not on that page.

PR #270, opened 2019-11-15
[Console] Fix AttributeError setting config values 

cas--:
GitHub user JohnDoee reported that config settings are not decoded
correctly, this error can be reproduced with a command like

    deluge-console -c /config/ "config --set download_location /downloads"

    > AttributeError: 'str' object has no attribute 'decode'

The tokenize code was using 'string-escape' to decode strings but there
is no direct replacement in Python 3 but also unnecessary. However the
tokenize code is complex and buggy and not really suitable for the task
of evaluating config values.

A better alternative is to evaluate the config values using the json
decoder with some additional logic to allow for previous syntax usage,
such as parentheses.

Added a comprehensive set of tests to check for potential config values
passed in from command line.

### DIFF
--- a/deluge/tests/test_ui_console.py
+++ b/deluge/tests/test_ui_console.py
@@ -11,6 +11,7 @@ import argparse
 
 from deluge.common import windows_check
 from deluge.ui.console.cmdline.commands.add import Command
+from deluge.ui.console.cmdline.commands.config import json_eval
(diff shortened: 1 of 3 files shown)


JohnDoee:
looks good to me, also does away with features probably nobody used :+1:.

PR #271, opened 2019-11-17
[Logging] Fix Python 3.8 compatibility

minus7:
Deluge's logger class extends Python's `logging.Logger`. Since Python
3.8, it takes an additional argument `stacklevel`.
The implementation in Deluge does not support that. Instead of working
around the problem by adding the parameter and ignoring it (or copying
code from `logging`), remove the brittle override.
Log messages still seem to work correctly.


minus7:
`**kwargs` is not a solution since it's positional arguments. Though `*args` + `**kwargs`  works.

PR #272, opened 2019-11-24
update russian localization

zluca:


cas--:
Thank you for this translation PR but we use Launchpad for translations: https://translations.launchpad.net/deluge/2.0

Since I cannot verify these translations I will have to close this.

PR #276, opened 2020-04-01
Add new REST-based Webapi and interactive webapi Docs with Swagger-UI

bendikro:


deluge/httpdownloader.py
```diff
@@ -328,3 +381,62 @@ def on_download_fail(failure):
     )
     d.addCallbacks(on_download_success, on_download_fail)
     return d
+
+
+def download_file_to_mem(
```
cas--, line 386: Can we move this changeset to a separate PR? Also no sure about the naming of this method but will talk about that in new PR

cas--:
The openapi docs are served by the app but perhaps we also want them in docs with https://github.com/sphinx-contrib/openapi

bendikro:
> Swagger is a good choice, I would prefer not storing the source code in codebase. Can we install with npm and copy what is needed in setup/install?

Certainly. I just included the source for simplicity. How (where) would you suggest doing this?

> Is _webapidoc_ is bit long-winded? Would dropping the web prefix _apidoc_ make it too generic and confused with deluge api? _wapidoc_, _restdoc_? Not sure here if there is much other option, any thoughts?

Can change to `api` to make it shorter.

> Also the endpoints seem very long too `'/webapidocs/web/deregister_event_listener'`, at least drop the 'webapidocs' part? Wonder if we customize, such as:
> 
> ```
> /torrent/files
> /torrent/info
> /events/register
> /events/deregister
> ```

`webapidocs` is the base prefix, so after changing that to `api`, the next part, `web` reflects the path of the function in the source. This matches the values given in the `get_methods` function for the old `json` api endpoint.

```
curl -X GET "http://0.0.0.0:8112/api/web/get_methods?id=1&pretty=pretty" -H "accept: application/json"
{
  "result": [
    "web.add_host",
    "web.add_torrents",
    "web.connect",
    "web.connected",
....
...
```

deluge/httpdownloader.py
bendikro: Will do

deluge/plugins/WebUi/deluge_webui/core.py
```diff
@@ -30,17 +30,28 @@
 
 log = logging.getLogger(__name__)
 
-DEFAULT_PREFS = {'enabled': False, 'ssl': False, 'port': 8112}
+DEFAULT_PREFS = {
+    'enabled': False,
+    'ssl': False,
+    'webapidoc_enabled': False,
```
cas--, line 36: * Do we need a config to enable/disable docs? 
* If so should we enable it by default? Users will never know about it and like me annoyed when they need to set a config ;)
* And rename this setting to just *docs*?

deluge/plugins/WebUi/deluge_webui/core.py
```diff
@@ -75,6 +86,7 @@ def start_server(self):
 
         self.server.port = self.config['port']
         self.server.https = self.config['ssl']
+        self.server.setup_webapidoc(enabled=self.config['webapidoc_enabled'])
```
cas--, line 89: I don't this setup should be part of the plugin code, the server itself should check config and do the setup

deluge/ui/web/server.py
```diff
@@ -748,6 +831,14 @@ def start_normal(self):
         ip = self.socket.getHost().host
         ip = '[%s]' % ip if is_ipv6(ip) else ip
         log.info('Serving at http://%s:%s%s', ip, self.port, self.base)
+        if self.webapidoc_enabled:
```
cas--, line 834: I'm not sure what value this log entry adds

deluge/ui/web/server.py
```diff
@@ -765,6 +856,14 @@ def start_ssl(self):
         ip = self.socket.getHost().host
         ip = '[%s]' % ip if is_ipv6(ip) else ip
         log.info('Serving at https://%s:%s%s', ip, self.port, self.base)
+        if self.webapidoc_enabled:
```
cas--, line 859: I'm not sure what value this log entry adds

deluge/ui/web/server.py
```diff
@@ -740,6 +819,10 @@ def start(self):
 
         component.get('Web').enable()
 
+        # TODO: Check if correct
+        # Plugins must be enabled here to register exported json functions
+        self.plugins.enable_plugins()
```
cas--, line 824: Were they not exported before?

deluge/ui/web/server.py
```diff
@@ -773,7 +872,10 @@ def stop(self):
         except KeyError:
             pass
 
-        self.plugins.disable_plugins()
+        # TODO: Check if this is correct
+        # Disabling plugins removes them from enabled_plugins list in web.conf???
+        # self.plugins.disable_plugins()
```
cas--, line 877: Interesting, could we put these plugin issues in a separate commit?

deluge/ui/web/server.py
```diff
@@ -724,6 +798,11 @@ def win_handler(ctrl_type):
 
             SetConsoleCtrlHandler(win_handler)
 
+    def setup_webapidoc(self, enabled=True):
```
cas--, line 801: Duplicate method naming, what is the purpose of this one?

deluge/ui/web/server.py
```diff
@@ -536,6 +585,28 @@ def __init__(self):
             theme = CONFIG_DEFAULTS.get('theme')
         self.__stylesheets.insert(1, 'themes/css/xtheme-%s.css' % theme)
 
+    def setup_webapidoc(self, enabled=True):
```
cas--, line 588: I feel that enabling the docs should be read from config not passed as arg, thoughts?

deluge/ui/web/server.py
```diff
@@ -536,6 +585,28 @@ def __init__(self):
...
+        """
+        path = 'swagger-ui'
+
+        if enabled:
+            if self.js.has_script(path):
+                return
+
+            self.putChild(self.webapidoc_path.encode(), self.webapidoc_resource)
+            self.js.add_script_folder(path, rpath('js', path))
+        else:
```
cas--, line 604: Is this to handle users switching the config from enable to disable?

setup.py
```diff
@@ -536,6 +536,8 @@ def run(self):
 
 setup_requires = ['setuptools', 'wheel']
 install_requires = [
+    "apispec; python_version >= '3'",
```
cas--, line 539: Just thought that I'm not sure we need to pin this to Python 3? Only apispec v3 is Python 3 only

setup.py
```diff
@@ -536,6 +536,8 @@ def run(self):
 
 setup_requires = ['setuptools', 'wheel']
 install_requires = [
+    "apispec; python_version == '3'",
```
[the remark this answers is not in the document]
bendikro: Fixed

deluge/plugins/WebUi/deluge_webui/core.py
bendikro: I'm not sure. You think it should always be available? You'll have to log in to do anything of importance anyways, so maybe we can drop the setting.

deluge/ui/web/server.py
bendikro: It shows where the docs are served. Isn't that useful in itself?

deluge/ui/web/server.py
bendikro: I think the problem I had was that the plugins list in the config was empty unless enabled here.

deluge/ui/web/server.py
bendikro: Yes, it is. Would be easier to just have it always enabled

setup.py
bendikro: Honestly, I'm not sure as I haven't tested on python 2.

Erarshad:

PR #277, opened 2020-04-03
[Console] Fix incorrect test for when a host is online

bendikro:
The tests in connectionmanager for when a host is online are broken
and always considers a host as online.

When an error occurs in e.g. in _on_connect_fail, report_message()
in PopupsHandler expects the message to be string, not a Twisted Failure.
Fix by checking if message is string and log a warning before converting
to string.

### DIFF
--- a/deluge/ui/console/modes/connectionmanager.py
+++ b/deluge/ui/console/modes/connectionmanager.py
@@ -37,7 +37,11 @@ class ConnectionManager(BaseMode, PopupsHandler):
         self.update_select_host_popup()
 
     def update_select_host_popup(self):
-        selected_index = self.popup.current_selection() if self.popup else None
+        selected_index = None
+        if self.popup:
+            if not isinstance(self.popup, SelectablePopup):
+                return
+            selected_index = self.popup.current_selection()
 
         popup = SelectablePopup(
             self,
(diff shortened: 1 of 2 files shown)


deluge/ui/console/widgets/popup.py
```diff
@@ -341,6 +343,10 @@ def __init__(
         width_req=0.5,
         **kwargs
     ):
+        if not isinstance(message, six.string_types):
+            log.warn("Expected string type for message. Got '%s'", type(message))
```
cas--, line 347: `warn` has been deprecated: https://docs.python.org/3/library/logging.html#logging.warning

cas--: I also don't think this check should be here and instead the decoding should be done in `_on_connect_fail`

deluge/ui/console/modes/connectionmanager.py
```diff
@@ -37,7 +37,11 @@ def __init__(self, stdscr, encoding=None):
         self.update_select_host_popup()
 
     def update_select_host_popup(self):
-        selected_index = self.popup.current_selection() if self.popup else None
+        selected_index = None
+        if self.popup:
+            if not isinstance(self.popup, SelectablePopup):
```
cas--, line 42: What issue is this catching?

deluge/ui/console/modes/connectionmanager.py
```diff
@@ -93,7 +98,10 @@ def _on_connect_fail(self, result):
             log.exception(result)
 
     def _host_selected(self, selected_host, *args, **kwargs):
-        if selected_host in self.statuses:
+        if (
+            selected_host in self.statuses
+            and self.statuses[selected_host][1] == 'Online'
```
cas--, line 103: I actually think it is useful to retain selecting offline hosts e.g. DNS hostname issue would be displayed...

image

PR #278, opened 2020-04-10
Fix deluge-web.service systemctl

cololi:
Because deluge-web don't have `-d` option, delete it.
```
deluge-web.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
```


bendikro:
`deluge-web` in v1.3.X does not have the `--do-no-daemonize` option, but in v2.X it does.

PR #283, opened 2020-04-19
Fix warning related to gettext

neeshy:
According to the upstream documentation, the name parameter of gettext.install should be a sequence of strings, not a string itself: https://docs.python.org/3.8/library/gettext.html#gettext.install

This was causing a warning when deluge was built against python 3.8.


cas--:
Thanks! Merged

PR #286, opened 2020-04-26
Add more linters

bwitt:
Add a few more linters and update files to be compliant

The changes are
* Convert to double-backtick in rst files   https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-literals
* add encoding statements consistently across all python files


bwitt:
this is probably excessive

PR #287, opened 2020-04-27
[#3309|GTK] Fix cmp function for None types

cas--:
Comparisons on Python 3 are much stricter resulting in the following
error comparing with None:

    TypeError: '>' not supported between instances of 'NoneType' and 'str'

Fix this by getting the type of the other value and getting it's default
value.


cas--:
https://dev.deluge-torrent.org/ticket/3309

PR #288, opened 2020-05-01
[Servers] Moved check_ssl_keys and generate_ssl_keys to crypto_utils.py

DjLegolas:
Also updated testssl.sh version to 3.0.1 (SECURITY_TEST)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,24 @@ jobs:
           key-server: keyserver.ubuntu.com
           install: python3-libtorrent-dbg
 
+      - name: Sets env var for security
+        if: (github.event_name == 'pull_request' && contains(github.event.pull_request.body, 'security_test')) || (github.event_name == 'push' && contains(github.event.head_commit.message, 'security_test'))
+        run: echo "SECURITY_TESTS=True" >> $GITHUB_ENV
+
       - name: Install dependencies
         run: |
           pip install --upgrade pip wheel
           pip install -r requirements.txt -r requirements-tests.txt
           pip install -e .
+
+      - name: Install security dependencies
+        if: contains(env.SECURITY_TESTS, 'True')
+        run: |
+          wget -O- $TESTSSL_URL$TESTSSL_VER | tar xz
+          mv -t deluge/tests/data testssl.sh-$TESTSSL_VER/testssl.sh testssl.sh-$TESTSSL_VER/etc/;
+        env:
+          TESTSSL_VER: 3.0.6
+          TESTSSL_URL: https://codeload.github.com/drwetter/testssl.sh/tar.gz/refs/tags/v
 
       - name: Setup core dump directory
         run: |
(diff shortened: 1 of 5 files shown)


Erarshad:

cas--:
Looks good, I'll merge it next time I'm working on the code. 

I thought we might need to leave core method stub for deprecation but core should not be a hard dependency for a UI component. Try to include your reasoning for making changes in the commit message.

cas--:
I did actually find that the Console UI is also depending on a core module: https://dev.deluge-torrent.org/ticket/3491

DjLegolas:
I'll look into it.

In addition, I have a problem with adding the security only unittest to the CI.
Do you have an idea?

.github/workflows/ci.yml
```diff
@@ -59,6 +59,16 @@ jobs:
           python -c 'from deluge._libtorrent import lt; print(lt.__version__)';
           catchsegv python -X dev -m pytest -v -m "not (todo or gtkui or security)" deluge
 
+      - name: Test security
+        # if: "contains(github.event.commits[0].message, 'security_test')"
+        run: |
+          SECURITY_TESTS=True
```
cas--, line 65: This doesn't actually work since it's not exported for pytest. To use `env` job keyword to set env var in an action: https://docs.github.com/en/actions/learn-github-actions/environment-variables

Thinking further maybe instead remove the `or security` from pytest step and add a step before `install dependencies` that sets the env vars, something like this:

```
      - name: Sets env var for security
        if: contains(github.event.head_commit.message, 'security_test')
        run: echo "SECURITY_TESTS=True" >> $GITHUB_ENV

      - name: Install dependencies
      ...

      - name: Install security dependencies
        if contains(env.SECURITY_TESTS, 'True')
        run: |
          wget -O- $TESTSSL_URL | tar xz
          mv -t deluge/tests/data testssl.sh-$TESTSSL_VER/testssl.sh testssl.sh-$TESTSSL_VER/etc/;
        env:
          TESTSSL_VER=3.0.1
          TESTSSL_URL=https://codeload.github.com/drwetter/testssl.sh/tar.gz/refs/tags/$TESTSSL_VER
```

If you use the latest version 3.0.6 then the TESTSSL_URL needs to a `v` prefixed e.g. `.../tags/v$TESTSSL_VER`

cas-- (changes requested):

.github/workflows/ci.yml
```diff
@@ -41,12 +41,25 @@ jobs:
...
           pip install -r requirements.txt -r requirements-tests.txt
           pip install -e .
 
+      - name: Install security dependencies
+        if: contains(env.SECURITY_TESTS, 'True')
+        run: |
+          wget -O- $TESTSSL_URL$TESTSSL_VER | tar xz
+          mv -t deluge/tests/data testssl.sh-$TESTSSL_VER/testssl.sh testssl.sh-$TESTSSL_VER/etc/;
+        env:
+          TESTSSL_VER: v3.0.6
```
cas--, line 60: The `v` prefix is the issue here.

.github/workflows/ci.yml
DjLegolas: sounds good. changed to this implementation.

cas-- (approved):

PR #289, opened 2020-05-02
[Core] Changed on_alert_tracker_error behavior for lt >= 1.2

DjLegolas:
libtorrent 1.2 added endpoint struct to each tracker, to prevent false updates
we will need to verify that at least one endpoint to the errored tracker is working.
if there is at least one working, it will not set the tracker status to error and set it to
`Announce OK`. otherwise, it will use the error message from the alert.

Should handle [3384](https://dev.deluge-torrent.org/ticket/3384).

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -24,8 +24,15 @@ from twisted.internet.defer import Deferred, DeferredList
 from twisted.internet.task import LoopingCall
 
 import deluge.component as component
-from deluge._libtorrent import lt
-from deluge.common import PY2, archive_files, decode_bytes, get_magnet_info, is_magnet
+from deluge._libtorrent import LT_VERSION, lt
+from deluge.common import (
+    PY2,
+    VersionSplit,
+    archive_files,
+    decode_bytes,
+    get_magnet_info,
+    is_magnet,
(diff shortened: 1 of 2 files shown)


deluge/core/torrentmanager.py
```diff
@@ -1392,7 +1392,22 @@ def on_alert_tracker_error(self, alert):
         log.debug(
             'Tracker Error Alert: %s [%s]', decode_bytes(alert.message()), error_message
         )
-        torrent.set_tracker_status('Error: ' + error_message)
+        if lt.version >= '1.2':
```
cas--, line 1395: This is comparing strings so the version comparison will not work as you expect it to. Look for other examples of version comparisons in the code.

DjLegolas: Fixed :)

cas--:
In the commit message, please use [Core] instead of [TorrentManager], it's too specific. Can you also include a link to the github discussion, thanks

kslr:
hey @cas--  Can you merge it ?

an0n666:
why is it not being merged? why is deluge so dead?

cas--:
thanks @DjLegolas, sorry took so long to merge

PR #291, opened 2020-05-08
Set alt/title attribute for better accessibility

wasamasa:
This allows viewing the country in textual form by hovering the flag image and displays it if the image couldn't be loaded.


cas--:
Nice! thank you, sorry took so long to merge but done now: 6d9dc9bd4

PR #292, opened 2020-05-11
[Tests] Fix console tests sometimes failing due to hard coded port

bendikro:
The tests in ConsoleUIWithDaemonBaseTestCase could fail to the hard coded
port 58900 being busy.
Fix by using the proper port found in self.listen_port

Also convert unnecessary use of assertTrue where more appropriate
assert functions should be used, such as assertEqual and assertIn


cas--:
Merged: 937afd921

PR #293, opened 2020-05-22
[Console, Gtk] Fix formatting of time units

getsnoopy:
This commits fixes the formatting of time units to be compliant
with the SI. It also updates the abbreviation of the week to the
prevalent one.


cas--:
I am aware of the SI unit spacing but this is a style choice, mostly to save horizontal space with consistency. 

If this was to be changed it would first involve a discussion.

PR #294, opened 2020-05-27
[WebUI] Fix filelistings md5sum

hatzel:
I encountered a bug in the web-ui when uploading a torrent file with file md5sums (as created by py3createtorrent when using the --md5 flag). No files would be shown in the file list for such files, the torrent could not be added.

This PR adds a test with an example torrent file as well as a fix. The fix relies on explicitly selecting expected dictionary fields, thereby avoiding a situation where json serialization is performed on a dictionary containing a byte array (which fails).

As far as I can tell other modules in deluge which also retrieve the metadata using libtorrent already do a variant of the explicit selection process that I added.

Let me know what you need to get this merged :)


hatzel:
Fixed the import order but from what I can tell the black error is not caused by my additions.

ahti:
@cas-- would you mind having a look at this? Sorry if you are the wrong one to ping, but you seem to be handling most PRs on this repo ^^

cas--:
I decided to create a more robust version to fix this issue and it is now merged to develop: 3ec23ad96b

Thanks @hatzel for the tests and potential fix.

PR #296, opened 2020-06-07
Update Install Instructions

SergioInToronto:
Seemed like a good first contribution.

There's also a snap called [deluge-lukewh](https://snapcraft.io/deluge-lukewh) but I wasn't sure if it was official.

### DIFF
--- a/docs/source/intro/01-install.md
+++ b/docs/source/intro/01-install.md
@@ -1,26 +1,24 @@
 # Installing Deluge
 
-These are the instructions for installing Deluge. Consider them a work-in-progress and
-feel free to make suggestions for improvement.
+Instructions for installing Deluge server and the GTK client.
+
 
 ## <i class="icon-ubuntu"></i> Ubuntu
 
-### PPA
+Deluge is available from Ubuntu's Universe repository.
 
-Until the stable PPA is updated, the development version of Deluge can be used:
+```
+sudo add-apt-repository universe
+sudo apt install deluge deluge-gtk
+```
 
-    sudo add-apt-repository -u ppa:deluge-team/stable
-    sudo apt install deluge
 
-### <i class="icon-python"></i> PyPi
+## <i class="icon-fedora"></i> Fedora
 
-To install from Python PyPi, Deluge requires the following system installed packages:
+```
+sudo dnf install deluge deluge-gtk
+```
 
-    sudo apt install python3-pip python3-libtorrent python3-gi python3-gi-cairo gir1.2-gtk-3.0 gir1.2-appindicator3
-
-Install with pip:
-
-    pip install deluge
 
 ## <i class="fa fa-windows"></i> Windows
 
@@ -29,9 +27,10 @@ available for Windows.
 
 Intrepid users can install Deluge from separate packages as detailed in [issue #3201].
 
+
 ## <i class="fa fa-apple"></i> macOS
 
-There is no `.app` package currently for macOS, but can try Deluge with [Homebrew].
+There is no `.app` package currently for macOS, but can install Deluge with [Homebrew].
 
 1.  Install [Homebrew]
 2.  Open a terminal.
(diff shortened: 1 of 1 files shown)


SergioInToronto:
The build failure is a lint issue. I'm confident I didn't caused it ;)

cas--:
> The build failure is a lint issue. I'm confident I didn't caused it ;)

Did you install and run pre-commit to lint code: https://deluge.readthedocs.io/en/latest/devguide/tutorials/01-setup.html#setup-pre-commit-hook

docs/source/intro/01-install.md
```diff
@@ -48,8 +47,26 @@ There is no `.app` package currently for macOS, but can try Deluge with [Homebre
...
+
+To install from Python PyPi, Deluge requires the following system installed packages:
+
+    sudo apt install python3-pip python3-libtorrent python3-gi python3-gi-cairo gir1.2-gtk-3.0 gir1.2-appindicator3
+
+Install with pip:
+
+    pip3 install deluge
+
+### Developer PPA
```
cas--, line 63: This is actually a stable PPA so shouldn't be called development PPA. It was actually a pain to maintain the original stable PPA and this 'daily' PPA tracks master branch. I think for now just call it Deluge PPA.

docs/source/intro/01-install.md
```diff
@@ -48,8 +47,26 @@ There is no `.app` package currently for macOS, but can try Deluge with [Homebre
...
+
+Deluge also has a [development ppa]
```
cas--, line 65: Deluge actually has two maintained daily build PPA, one each for develop and master branches. The stable PPA is for the latter. You have conflated the two here so in the first case it should only be mentioned as stable and link to such: https://launchpad.net/~deluge-team/+archive/ubuntu/stable/ 

It might be good to mention the other PPA but I feel a link to another document with those details is better than easily giving users access to work-in-progress code.

docs/source/intro/01-install.md
```diff
@@ -1,26 +1,24 @@
 # Installing Deluge
 
-These are the instructions for installing Deluge. Consider them a work-in-progress and
-feel free to make suggestions for improvement.
+Instructions for installing Deluge server and the GTK client.
```
cas--, line 3: There is no mention of server or client in this document so simply say `Deluge`

docs/source/intro/01-install.md
```diff
@@ -1,26 +1,24 @@
...
 
 ## <i class="icon-ubuntu"></i> Ubuntu
 
-### PPA
+Deluge is available from Ubuntu's Universe repository.
 
-Until the stable PPA is updated, the development version of Deluge can be used:
+```
+sudo add-apt-repository universe
+sudo apt install deluge deluge-gtk
```
cas--, line 12: There is no need to install both deluge and deluge-gtk, just *deluge* will do

docs/source/intro/01-install.md
```diff
@@ -1,26 +1,24 @@
...
 
-    sudo add-apt-repository -u ppa:deluge-team/stable
-    sudo apt install deluge
 
-### <i class="icon-python"></i> PyPi
+## <i class="icon-fedora"></i> Fedora
 
-To install from Python PyPi, Deluge requires the following system installed packages:
+```
+sudo dnf install deluge deluge-gtk
```
cas--, line 19: see above

docs/source/intro/01-install.md
SergioInToronto: Does Deluge consider master stable? When I think "stable" I'm only thinking about releases.

SergioInToronto: I guess since there's a develop branch, that is the unstable work.

docs/source/intro/01-install.md
SergioInToronto: I found that confusing. I installed `deluge`, then pressed super, typed "deluge", but nothing appeared. It was because only the server was installed. I think it's important to mention both.

SergioInToronto: Whoops my mistake. That only happened when installing the snap from Ubuntu's software app. The first entry is a snap, the 2nd is from apt.
image

SergioInToronto: Speaking of, what is this?
https://github.com/Lukewh/deluge-snap

SergioInToronto:
Thanks for your feedback. Let me know if there are more issues.

cas--:
Thanks for providing changes and sorry took so long to get around to merging, I actually wanted to make a few other changes and not hard link individual forum posts.

PR #297, opened 2020-07-02
[#3299|GTK] Fix spin_share_ratio min clamps to 0.5

yuri-sevatz:
In gtkui Settings -> Queue -> "Share Ratio", we cannot set values lower than 0.5.

Attempting to set lower values results in:

    * ConfigValueChanged: stop_seed_ratio: 0.5

Deluge Web UI, Deluge Console allow setting values as low as 0.0:

    * ConfigValueChanged: stop_seed_ratio: 0.0

Fixed by decreasing the lower clamp in gtkui to match webui.

### DIFF
--- a/deluge/ui/gtk3/glade/preferences_dialog.ui
+++ b/deluge/ui/gtk3/glade/preferences_dialog.ui
@@ -16,7 +16,6 @@
     <property name="page_increment">10</property>
   </object>
   <object class="GtkAdjustment" id="adjustment_share_ratio">
-    <property name="lower">0.5</property>
     <property name="upper">100</property>
     <property name="value">2</property>
     <property name="step_increment">0.10000000000000001</property>


yuri-sevatz:
Changes:

- Removed `<property name="lower">0</property>`, since the default for `GtkAdjustment.lower` is already 0

This change seems to agree with what `glade` generates for this, if I set to 0 using their WISIWYG, and then save.

yuri-sevatz:
CI seems to be broken here, unrelated to change.

doadin (approved):

cas--:
If anything it would be the other way around to update Web UI since Deluge has always kept this as a 'be fair' feature

doadin:
@cas--  I get be fair and you should seed when it comes to public torrents etc etc, however what about private torrents? Do we just not care for the sake of public torrents? We do also realize we can just manually stop, not adding this is just an inconvenience. I hope my point is clear, also just my opinion.

yuri-sevatz:
So, it's good that you're engaging the moral considerations for leeching.  Though it looks like the structure of this thread may only be scratching the surface on more detailed deliberations for "fairness", "liability", and "costs" that seem to yield more than one moral outcome, than what I think this thread may be leading usto believe here.

If I list a few examples that come to mind, as well as some I've hit myself...

- Fairness: The user may be duplicating redundant resources, which have already been provisioned via other means:
    
  Example: **Peer Teaming** When multiple clients are setup in the same subnet, to service different default routes, and each client has different resource availability.  Sometimes you'll use one client to seed and not the other.  (I used to do this while in University to help accelerate whole swarms for "spotty" torrents, it's awesome!)

  Example: **Bootstrap Pools** when a pool offers BitTorrent only as a means of mirroring existing resources already hosted elsewhere as a fallback, with no administratively required uptime in the BitTorrent swarm (e.g. WebTorrent, for archives).

- Liability: The user may be using BitTorrent to commit a crime, despite that crime in itself being a moral positive:

  Example: **Legal or human-rights related considerations** for the material being exchanged over a pool is illegal.  Religious or banned texts? (Thinking of: China, Turkey, Russia, Saudi Arabia, or any government who's best days are ahead of them or behind them)

- Costs: The user may be on a metered connection:

  Example: **Comcast Cable / Xfinity** (Yes, we're going there! ^.^)
    
From a moral standpoint, the existence of the above outliers suggests that we should probably be deferring any judgement to the fact that a BitTorrent Client can't fundamentally distinguish what is "moral seeding," from what is not, because the BitTorrent protocol itself does not fundamentally preclude using it in the above moral situations that would be ruled as immoral

All that said, this PR was more just to help others caught in similar situations as me.  But my moral assessment is that closing this PR seems like a misstep.

yuri-sevatz:
I do like @doadin 's summary a lot better than that book I just wrote 😄.  He's right about how this evolved...

BitTorrent didn't try and solve this in the client, because from a security standpoint, the client cannot be trusted by the swarm.  Trusting the client would be a faulty assumption.

So BitTorrent Trackers evolved to allow swarms to resolve the issues of leaching, by implementing tracker metrics -- which keeps the protocol decentralized, while still allowing independent authorities on "good behavior" to cultivate smaller ecosystems of trust, all while still allowing BitTorrrent Clients to stop seeding in more general situations that certain use-cases call for.

It was either that, or we would have to build a more centralized authority on trust, or some kind of public ledger of bad seeding behavior... Though I dare not say the "B" word among developers, lol.

doadin:
@cas-- I don't want to keep going at this but , you know you could set upload speed to like 0 and max number of seeds to 0 and pretty much get a similar result to just stopping as soon as the torrent is done and upload like nothing aka still leech right? I don't really see how leaving this out makes it much better. Again not to get into this a lot more but if you don't want leachers I think you might need to get stricter but I just don't really see the point anyone can just compile this client or several others to do w.e. they want to. I think most trackers that care monitor it themselves, for this reason.

yuri-sevatz:
I think @cas-- has already made up their mind.

But I think he's showing a fundamentally flawed security understanding; to think that this can guarantee seeds by writing a **client** a certain way -- when the bug he's trying to overcome is in the **protocol** (i.e. only enforceable in the tracker with the help of data from the swarm).

My main gripe with trying to do this in the client:  is that if bittorrent doesn't know what leeching is, then deluge won't either.  So pretending that we've solved this in deluge by limiting what would otherwise be a normal expression of the bittorrent spec, is somewhat flawed thinking.

PR #298, opened 2020-07-25
Remove nonexistent web argument from Linux docs

BenQuigley:
This commit fixes the issue that the docs for setting up a systemd service specify a `-d` argument for both deluged and deluge-web, that does not actually exist (anymore?) for deluge-web:
https://github.com/deluge-torrent/deluge/blob/b1cdc32f7357e9777eb6cc2d90094c7d122af2eb/deluge/ui/web/web.py

### DIFF
--- a/docs/source/how-to/systemd-service.md
+++ b/docs/source/how-to/systemd-service.md
@@ -136,7 +136,7 @@ ExecStart=/usr/bin/deluged -d -l /var/log/deluge/daemon.log -L warning
 ```
 
 ```
-ExecStart=/usr/bin/deluge-web -d -l /var/log/deluge/web.log -L warning
+ExecStart=/usr/bin/deluge-web -l /var/log/deluge/web.log -L warning
 ```
 
 See `deluged -h` for all available log-levels.
--- a/packaging/systemd/deluge-web.service
+++ b/packaging/systemd/deluge-web.service
@@ -8,7 +8,7 @@ Wants=deluged.service
 Type=simple
 UMask=027
 
-ExecStart=/usr/bin/deluge-web -d
+ExecStart=/usr/bin/deluge-web
 
 Restart=on-failure
 


petersasi:
AFAICS this is about 
Correct me if I am wrong but as far as I know the deluge-web-debug version of each executable is the one to be used for installation as a service. I mean one keeps writing to stdout while the other forks off and exits but I  don't remember which is which.
1. Maybe the above distinction is windows only?
2. I do find this very misleading, because in reality -L debug is the command line parameter that enables debug logging.
3. Not sure if it is my place to say, but I do usually promote using -L warning by default, this will log events that point to performance bottlenecks (e.g. reaching certain buffer / list  size / bandwidth limits).

BenQuigley:
> Maybe the above distinction is windows only? - EDIT: i have just installed a Focal, yes it, sorry it is confusing vs. Windows Deluge

No; you can see in the source code that there's no `-d` option for `deluge-web` (which as you point out is a separate executable).

> I do find this very misleading, because in reality -L debug is the command line parameter that enables debug logging.
 
In the `deluged` executable, `-d` is for `--do-not-daemonize`. I think (but have not checked) this option may have once existed in `deluge-web` too, but either way, it needs to be removed from these service docs because when someone tries to build a systemd service that tells the system to run `deluge-web` with `-d`, it is going to fail:

```
$ deluge-web -d
Usage: deluge-web [options] [actions]

```

bendikro:
> No; you can see in the [code](https://github.com/deluge-torrent/deluge/blob/b1cdc32f7357e9777eb6cc2d90094c7d122af2eb/deluge/ui/web/web.py) that there's no `-d` option for `deluge-web` (which as you point out is a separate executable).

On [line 63](https://github.com/deluge-torrent/deluge/blob/b1cdc32f7357e9777eb6cc2d90094c7d122af2eb/deluge/ui/web/web.py#L63) you find the statement `self.parser.add_process_arg_group()` which is a call on [deluge/ui/baseargparser.py:BaseArgParser.add_process_arg_group()](https://github.com/deluge-torrent/deluge/blob/b1cdc32f7357e9777eb6cc2d90094c7d122af2eb/deluge/ui/baseargparser.py#L347) which adds the `--do-not-daemonize` option when not on windows.

BenQuigley:
Hmm, thanks for the correction @bendikro. 

I encountered that "no such option" error on deluge-web version 1.3.13.

bendikro:
> Hmm, thanks for the correction @bendikro.
> 
> I encountered that "no such option" error on a Raspberry Pi running deluge-web version 1.3.13.

Yeah, the `-d` option in deluge-web is new in deluge v2

BenQuigley:
Thanks!

PR #299, opened 2020-07-31
rm -c flag conflicts with --config shortcut

Lucas-C:
cf. https://github.com/deluge-torrent/deluge/blob/master/deluge/argparserbase.py#L191

This bug triggers "Password does not match" errors when "rm -c" is used


Erarshad:

PR #302, opened 2020-11-24
[UI] fix tracker icon download function in python3

amefs:
to avoid such error in python3 caused by different string type
```
Error occurred downloading file from "http://b'acg.rip'/": invalid hostname: b'acg.rip'
```

### DIFF
--- a/deluge/ui/tracker_icons.py
+++ b/deluge/ui/tracker_icons.py
@@ -12,6 +12,7 @@ from __future__ import unicode_literals
 import logging
 import os
 from tempfile import mkstemp
+from six import string_types
 
 from twisted.internet import defer, threads
 from twisted.web.error import PageRedirect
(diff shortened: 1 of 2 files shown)


cas--:
I think the problem is the caller not the function, it should be providing a string not bytes to TrackerIcons get() or fetch(). All decoding should be done early. Can you provide a full trace of the issue?

amefs:
Yep, I traced the code again. Then, find the byte was given by `Tracker.getChild`
```
request.tracker_name = path
```
and used by `Tracker.render`, which called `tracker_icons.fetch` method
```
d = self.tracker_icons.fetch(request.tracker_name)
```
can I fix it here?

cas--:
Yes we would fix it in that `Tracker.render` call by decoding `request.tracker_name` since the request attributes are always bytes. No need for six check just use `.decode()`

cas--:
Thanks for reporting and providing a fix, I have merged as single commit with comments: a4da8d29f

PR #304, opened 2020-12-19
Update systemd-service.md

groom-polar-outage:
deluge-web doesn't have option -d


Himel55:
I ran into this issue as well, would be nice if the documentation was updated.

groom-polar-outage:
Yeah, without -d option.

cas--:
See #298 #262

PR #305, opened 2021-01-23
Update deluge-web.service

markus2929:
removed -d option as it doens't exist for deluge-web anymore


bwitt:
`-d` is new for 2.0; see https://github.com/deluge-torrent/deluge/pull/304

PR #306, opened 2021-02-15
[GTKUI] Added detection of torrent URL on GTK UI focus 

matiasw:
There is a feature in Transmission Remote GUI that pops up the "Add torrent" dialog automatically if it detects that there is a torrent URL on the clipboard when the program gets focus. This is a preliminary implementation of that feature in Deluge.

At the moment, it works, but it does not have the intelligence to check if the torrent on the clipboard is already being downloaded. On IRC, I was adviced to "grab the .torrent and check to see if the hash already exists", but this seems like a rather heavy operation to perform each time the program gets focus with a torrent URL on the clipboard.

### DIFF
--- a/deluge/ui/gtk3/mainwindow.py
+++ b/deluge/ui/gtk3/mainwindow.py
@@ -14,17 +14,17 @@ import os.path
 from hashlib import sha1 as sha
 
 import gi
-from gi.repository import Gdk, Gtk
+from gi.repository import Gtk
 from gi.repository.Gdk import DragAction, WindowState
 from twisted.internet import reactor
 from twisted.internet.error import ReactorNotRunning
 
 import deluge.component as component
-from deluge.common import decode_bytes, fspeed, resource_filename, is_url, is_magnet
+from deluge.common import decode_bytes, fspeed, is_magnet, is_url, resource_filename
 from deluge.configmanager import ConfigManager
 from deluge.ui.client import client
 
-from .common import get_deluge_icon, windowing, get_clipboard_text
+from .common import get_clipboard_text, get_deluge_icon, windowing
 from .dialogs import PasswordDialog
 from .ipcinterface import process_args
 
@@ -148,7 +148,7 @@ class MainWindow(component.Component):
             'NewVersionAvailableEvent', self.on_newversionavailable_event
         )
 
-        self.previous_text = ''
+        self.previous_clipboard_text = ''
         self.first_run = True
 
     def connect_signals(self, mapping_or_class):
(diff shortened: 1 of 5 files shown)


DjLegolas (changes requested):
Please fix the code formatting errors (from `lint`):
1. fix `import`s order
2. use only single-quote

Best way is to run the `tox` tests yourself.

In addition, please change the commit message as it is too long for one line.
Something like:
```
[GTKUI] Added detection of torrent URL on GTK UI focus

In case deluge GTK gets focus with a new torrent URL on the clipboard, the "Add Torrent from URL" dialog will pop up automatically
```

matiasw:
@DjLegolas Thanks for the feedback. I changed quotes into single quotes and changed the import order; not sure if it's still the way you'd like?

deluge/ui/gtk3/mainwindow.py
```diff
@@ -12,9 +12,10 @@
 import logging
 import os.path
 from hashlib import sha1 as sha
+from urllib.parse import urlparse
 
 import gi
-from gi.repository import Gtk
+from gi.repository import Gtk, Gdk
```
DjLegolas, line 18: Just switch the order like this:
```python
from gi.repository import Gdk, Gtk
```

DjLegolas (changes requested):

deluge/ui/gtk3/mainwindow.py
matiasw: Ok, done.

Erarshad:

cas-- (changes requested):

matiasw:
Ok, so I changed the code to use these premade functions. It is a bit neater, now.

cas-- (changes requested):
Looking much better, can we do some final linting cleanup. Use `isort` to auto-fix the linting issues. 

I was just testing it out and wonder if this should be a user preference since without any way for the user to disable it might be annoying. 

You can search for the `gtkui_config['show_piecesbar']` preference to use as a guide to implementing it.

image

matiasw:
I have made the requested modifications and added the Preferences setting for this feature.

cas--:
Excellent, thanks! Just need to fix the failing linting issues

image

If you install [pre-commit](pre-commit.com) then your commits are linted automatically

matiasw:
Uh, so, I did install pre-commit, and on my machine, with `pre-commit run --all-files`, I get all PASSED, but the CI test on Github still fails. If I knew why, I could fix it?

cas--:
That is strange, I see this: 

image

Don't worry I'll can resolve it when merging it :+1:

cas--:
Thanks @matiasw for this useful feature!

PR #308, opened 2021-03-28
remove deluge-web -d param

Aeonos:
deluge-web does not support a -d parameter, which results in a problem, when one tries to start it like in this tutorial. I would therefore recommend to remove it. I dont know, if the parameter was removed or something, but it is not valid at the moment.


cas--:
Search first

PR #314, opened 2021-10-27
Update deluge-web.service

suijurisinfowarrior:
The flag -d does not exist and will cause the following error:

deluge-web -d -l /var/log/deluge/web.log -L warning
Usage: deluge-web [options] [actions]

uname -srvo
Linux 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) GNU/Linux


cas--:
That applies to 1.3 not 2.0 see the many duplicate PRs...

PR #315, opened 2021-10-27
Fix plugins still needing specific python version.

mhertz:


cas--:
Ah did I make a mistake in implementing this in 3433a911. Could you mention this in the commit message?

The improved fix would be to specify the [Environment kwargs](https://setuptools.pypa.io/en/latest/pkg_resources.html#environment-objects) so it clear what is being set to None e.g.:

```
Environment(search_path=plugin_dirs, platform=None, python=None)
```

mhertz:
Thanks alot for your reply and review Cas :) I'm sorry but stupidly deleted my forked repo and branch, and tried bring it back by pulling the PR and repushing, which worked but didn't update this here(noob at git and PRs in general), only mine, so I remade new PR and will close this, and obviously not do that mistake again. Sorry about that.

PR #317, opened 2021-11-24
[GTK3] Support using the Ayatana fork of indicators.

unit193:
As this fork is maintained in Debian, and as of Impish/21.10 is the supported variant in Ubuntu as well, as such try it first.


cas--:
Thanks, looks good :+1:

PR #318, opened 2021-12-17
[GTKUI] Enable indicator support by default.

Unit193:
Howdy,

On GNOME, I don't believe any tray icon or indicator support is enabled by default, but one can easily install an extension for indicators whereas I'm not sure about tray icons, but the 'top icons' extension I believe has had a flaky history.  I'm not entirely sure how KDE handles things, but out of the box it too has indicator support and I believe that is preferred with the move to Wayland.  In Xfce, the tray icon area is called "Status tray" and has support for both tray icons and indicators.

I think at this point it might be preferred to default to indicators, since (I believe) those are more supported in GNOME and KDE.

Thanks!


cas--:
Does seem like a good idea to switch it now. It will mess with my PPA patches that switch this value :wink:

Unit193:
@cas-- Mind changing those patches so this can be merged in? ;)

PR #319, opened 2021-12-17
Support announcing to all trackers of the same tier.

unit193:
This fixes bug http://dev.deluge-torrent.org/ticket/1395


DjLegolas:
Looks good.
Also, if you can, please add this to the WebUI (and console) as well.

unit193:
Ah yes, good plan.  I haven't tested those of course, but I've been using this patch otherwise for quite some time.

doadin (approved):

cas--:
Can I just clarify what implementation we trying to achieve here with regards tracker announcing since there are two options:

`announce_to_all_trackers` 

> all trackers in the same tier are announced to in parallel. If all trackers in tier 0 fails, all trackers in tier 1 are announced as well.

`announce_to_all_tiers` 
> one tracker from each tier is announced to. This is the uTorrent behavior.

I have updated the Trac ticket with relevance discussion points but my take on this is to use the latter `announce_to_all_tiers` as per uT.

deluge/core/preferencesmanager.py
```diff
@@ -85,6 +85,7 @@
     'max_active_downloading': 3,
     'max_active_limit': 8,
     'dont_count_slow_torrents': False,
+    'announce_to_all_trackers': True,
```
cas--, line 88: This should default to `False`

Unit193:
I'd have gone for the former, it makes a lot more sense to me as you can configure exactly which trackers are fallback, and makes more logical sense to treat trackers set as the same priority in the list, well, the same.

It seems they're taking an entirely different definition to the tiers though, which if selecting this option would surprise me as a user as I'd have different expectations.  Perhaps a mouseover if the latter is used?

Unit193:
@cas-- I'm not specifically attached to the one way of doing this, I do have a preference but in the end either way will allow me to do what I want (without locally patching.)  If you have wording suggestions for mouseover text or whatnot, that'd be appreciated too (unless you want to just adapt it.)

Fs00:
I'm surprised to see this PR left stale for years - it's a very small feature that can make a big difference!
I agree with @cas-- that this should use `announce_to_all_tiers`. I don't see the point in making Deluge behave differently from uTorrent, qBittorrent and other popular clients, considering that there are sites providing tracker lists (like [newTrackon](https://newtrackon.com/list)) which assume that clients use the uTorrent behavior.

PR #320, opened 2021-12-21
[Console] Removed Core dependency from Console UI

DjLegolas:
UIs should not depend on core directly, so removing the dependency in ConsoleUI.
This is done by adding a hard-coded variable.

Closes https://dev.deluge-torrent.org/ticket/3491

### DIFF
--- a/deluge/ui/console/widgets/statusbars.py
+++ b/deluge/ui/console/widgets/statusbars.py
@@ -11,8 +11,9 @@ from __future__ import unicode_literals
 
 import deluge.common
 import deluge.component as component
-from deluge.core.preferencesmanager import DEFAULT_PREFS
 from deluge.ui.client import client
+
+DEFAULT_DAEMON_PORT = 58846
 
 
 class StatusBars(component.Component):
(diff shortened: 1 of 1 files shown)


DjLegolas:
@cas-- what do you think about adding a test for making sure that the daemon port in both the `console` and the `core` are the same?
This way, in the event that the default port will be changed in the future, there will be a test for making sure that it will be changed in both places.
The problem is adding a dependency for the core in the console test...

deluge/ui/console/widgets/statusbars.py
```diff
@@ -11,9 +11,11 @@
 
 import deluge.common
 import deluge.component as component
-from deluge.core.preferencesmanager import DEFAULT_PREFS
 from deluge.ui.client import client
 
+# The same as the value in core.preferencesmanager.DEFAULT_PREFS['daemon_port']
```
cas--, line 16: We don't need this comment, it's fine to assume

cas--:
It's not that important to test for it, if anything I'd spend the time creating a common constant but again not that important since we'll likely never change this value.

deluge/ui/console/widgets/statusbars.py
DjLegolas: ok. removed.

cas-- (approved):

PR #321, opened 2021-12-21
[WebUI] Fixed ETA sorting in WebUI

DjLegolas:
When sorting the according to ETA values, all torrents with infinite value were being
considered a lower value (INF -> 12 -> 32) instead of largest (12 -> 32 -> INF).
This is due to the fact that the INF symbol is placed to lower value (<= 0).
Now the lower values are being treated as the largest JS number when sorting.

Closes https://dev.deluge-torrent.org/ticket/3413

### DIFF
--- a/deluge/ui/web/js/deluge-all/TorrentGrid.js
+++ b/deluge/ui/web/js/deluge-all/TorrentGrid.js
@@ -67,7 +67,9 @@
     }
 
     function etaSorter(eta) {
-        return eta * -1;
+        if (eta === 0) return Number.MAX_VALUE;
+        if (eta <= -1) return Number.MAX_SAFE_INTEGER;
+        return eta;
     }
 
     function dateOrNever(date) {
@@ -75,7 +77,9 @@
     }
 
     function timeOrInf(time) {
-        return time < 0 ? '&infin;' : ftime(time);
+        if (time === 0) return '';
+        if (time <= -1) return '&infin;';
+        return ftime(time);
     }
 
     /**


cas--:
I think there is a bit more to this and would like to fix it correctly. 

We have have three states that the torrent.get_eta method can return and the eta can be for either downloading or seeding:

* 0 (no transfer so no eta)
* -1 (more than a year)
* 1-31556952 secs (up to a year)

Therefore:
* Only display the infinity symbol for negative integers
* For zero we show nothing i.e. empty string

There is an issue that `ftime` function is returning `inf` for 0 where I think it should return `''`. So either we fix `ftime` or fix `timeOrInf` method to return `''` for 0.

Where does that leave the sorting that the user should see? I am assuming for ascending it would look like this:

```
1 
5
-1
0
```
Perhaps assign -1 to `Number.MAX_SAFE_INTEGER` and 0 to `Infinity` you may get the correct sorting.

DjLegolas:
Well, looks like now it if fixed. Including giving an empty value to 0.

deluge/ui/web/js/deluge-all/TorrentGrid.js
```diff
@@ -67,15 +67,27 @@
...
+        }
     }
 
     function dateOrNever(date) {
         return date > 0.0 ? fdate(date) : _('Never');
     }
 
     function timeOrInf(time) {
-        return time < 0 ? '&infin;' : ftime(time);
+        if (time < 0) {
```
cas--, line 84: We can simplify this nesting to make it more readable by using 'guard clauses':

```
if (time === 0) return '';
if (time <= -1) return '&infin;';
return ftime(time);
```

deluge/ui/web/js/deluge-all/TorrentGrid.js
```diff
@@ -67,15 +67,27 @@
     }
 
     function etaSorter(eta) {
-        return eta * -1;
+        if (eta < 0) {
```
cas--, line 70: See using guard clauses below, perhaps putting the 0 assertion first, since perhaps more likely outcome

cas-- (commented):
Looks great! If you don't mind could you refactor it slightly, thanks

PR #323, opened 2021-12-23
Add Keywords property to desktop file

RedBearAK:
Deluge fails to appear in some app launchers (GNOME app search, Albert launcher) when searching for just "torrent" or other keywords, rather than "bittorrent". This is due to the lack of a Keywords header/property in its desktop entry file. Adding this line should solve the issue. 

I don't know if the underscore "_" is actually necessary for this line, I just copied the appearance of the lines above it when inserting. Please check that this comes out without the underscore in the final file after processing.


cas--:
Thank you, the underscore is required so it can be translated into other languages :+1:

RedBearAK:
> Thank you, the underscore is required so it can be translated into other languages 👍

@cas-- 

Thanks, that confirms what I suspected to be the case. 👍

PR #324, opened 2021-12-24
[Gtk] Fixed edit torrents dialogs windows close issues

DjLegolas:
Up until now, when closing the Add or Edit dialogs, of the `Edit Torrents`, using the top-right X
button or using the Escape key, they were being destroyed without any way to reopening them, and
it was breaking the `Edit Torrents` window itself.
Now both dialogs have the right handlers for handing the closing process without breaking anything.


DjLegolas:
@cas-- This PR finishes the problem I had when implementing [#2434](https://dev.deluge-torrent.org/ticket/2434).
You can close it as well when merging IMO.

PR #325, opened 2021-12-25
Drop Python2 support

DjLegolas:
As python2 is already at its EOF state, we can drop all of the support to it.

This PR will contain all of the changes to this end.

Plan:

- [x] Remove `__future__` imports
- [x] Remove `six` dependency
- [x] Remove `future_builtins` imports
- [ ] Change the way we perform string formats (maybe even move to `f-strings` only?)
- [x] Make sure there are no sections that handles Py2 compatibilities
- [ ] Remove files coding header

If you know about any other things to change, please write.


deluge/crypto_utils.py
```diff
@@ -105,9 +103,7 @@ def generate_ssl_keys():
     """
     This method generates a new SSL key/cert.
     """
-    from deluge.common import PY2
-
-    digest = 'sha256' if not PY2 else b'sha256'
+    digest = b'sha256'
```
cas--, line 106: This doesn't seem correct, should be string not bytes. 

Discussion for reference: https://github.com/pyca/pyopenssl/issues/730

deluge/core/alertmanager.py
```diff
@@ -28,13 +26,7 @@
...
 log = logging.getLogger(__name__)
 
-try:
-    SimpleNamespace = types.SimpleNamespace  # Python 3.3+
-except AttributeError:
-
-    class SimpleNamespace(object):  # Python 2.7
-        def __init__(self, **attr):
-            self.__dict__.update(attr)
+SimpleNamespace = types.SimpleNamespace  # Python 3.3+
```
cas--, line 29: Can we remove the unneeded comment

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 2.0.6 (WIP)
```
cas--, line 3: This will need to be 2.1.0 since it's a breaking compatibility change

cas-- (commented):
If we are making wholesale changes, we can remove the following from all files: 

`# -*- coding: utf-8 -*-`

deluge/core/alertmanager.py
cas--: Actually just import SimpleNamespace from types

pyproject.toml
```diff
@@ -10,4 +10,4 @@ skip-string-normalization = true
 [tool.isort]
 profile = "black"
 # Python 2 stdlib
-extra_standard_library = ["urlparse", "HTMLParser", "urllib2"]
+# extra_standard_library = ["urlparse", "HTMLParser", "urllib2"]
```
cas--, line 13: Remove these lines instead of commenting out

setup.py
```diff
@@ -9,29 +9,21 @@
...
-
 import glob
 import os
 import platform
 import sys
-from distutils.command.build import build as _build
-from distutils.command.clean import clean as _clean
-from distutils.command.install_data import install_data as _install_data
-from shutil import rmtree
+from distutils.command.build import build as _build  # TODO: deprecated in PY3.12
```
cas--, line 16: Please leave don't include TODO comments, this will be dealt with when the time comes

deluge/crypto_utils.py
DjLegolas: Stupid docstring... The IDE warned me about it being bytes and not string because of it.
Will change in the next push.

CHANGELOG.md
DjLegolas: No problem.
Do you want me to tag it already as 2.1.0 or add a new one for future releases like `develop`?

cas--:
Thanks for getting this work started. 

Maybe look at using this: https://github.com/asottile/pyupgrade with `--py36-plus` since with can bump to >=3.6 (I know 3.6 EOL)

CHANGELOG.md
cas--: Sure you can change the version number. I'll remember to bump to dev tag when merging

bwitt:
+1 for using https://github.com/asottile/pyupgrade with `--py36-plus` ; 
also suggest using https://github.com/asottile/reorder_python_imports with `--py36-plus` to fix up a few more things.

DjLegolas:
> Thanks for getting this work started.
> 
> Maybe look at using this: https://github.com/asottile/pyupgrade with `--py36-plus` since with can bump to >=3.6 (I know 3.6 EOL)

Do you want to add it permanently to the linting process?

cas--:
What I would suggest is that this PR is wrapped up as is and we merge it. So things to finalise for this PR:
 
* remove urllib imports in PreferencesManager to pass the linting
* Any tests or reference to Python 2 use `git grep -Ei "python.?2|py2[ 0-9]"`

Then create a second PR that uses pyupgrade since that will be another huge changeset. Adding pyupgrade to pre-commit is helpful since if added as a pre-commit `[manual]` stage we can run it quickly with

`pre-commit run pyupgrade -a `

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.1.0 (WIP)
+
+### Codebase
+
+- Dropped support for Python 2
+
 ## 2.0.6 (WIP)
```
cas--, line 9: Best to remove this since we'll add it back when needed

PR #327, opened 2021-12-29
Windows CI improvments

DjLegolas:
This PR adds several things:

- Windows CI will use python3.8 - the same as the Linux CI (I don't see any reason why not)
- Added `SVG` MIME type for icons as all browsers [support it](https://www.w3schools.com/html/html_favicon.asp) as favicon.
- Enabled TrackerIcons tests for Windows as `os.path.samefile` is now supported on Windows (as of [python3.2](https://docs.python.org/3/library/os.path.html#os.path.samefile)).


deluge/tests/test_tracker_icons.py
```diff
@@ -24,8 +22,8 @@
 @pytest.mark.internet
 class TrackerIconsTestCase(BaseTestCase):
 
-    if windows_check():
-        skip = 'cannot use os.path.samefile to compair on windows(unix only)'
+    # if windows_check():
+    #     skip = 'cannot use os.path.samefile to compair on windows(unix only)'
```
cas--, line 26: We should always aim to delete instead of commenting out code

cas--:
Thanks merged: f3784723a

deluge/tests/test_tracker_icons.py
DjLegolas: forgot I commented out this code.. whoops

PR #331, opened 2022-01-04
Update Windows Packaging

doadin:
- [x] Check that deluge-console works.

- [x] Plugins Execute & Extractor Won't start:
```
09:58:44.350 [ERROR   ][deluge.pluginmanagerbase          :78  ] Unable to instantiate plugin 'Execute' from 'e:\\program files\\deluge\\deluge\\plugins\\execute-1.3-py3.10.egg'!
09:58:44.350 [ERROR   ][deluge.pluginmanagerbase          :78  ] No module named 'twisted.internet.utils'
  File "deluge\pluginmanagerbase.py", line 147, in enable_plugin
  File "e:\program files\deluge\deluge\plugins\execute\deluge_execute\__init__.py", line 14, in __init__
    from .core import Core as _pluginCls
  File "e:\program files\deluge\deluge\plugins\execute\deluge_execute\core.py", line 14, in <module>
    from twisted.internet.utils import getProcessOutputAndValue
ModuleNotFoundError: No module named 'twisted.internet.utils'
```
   Answer: https://github.com/deluge-torrent/deluge/pull/331/commits/3aec79f30f39d76d3afefa8e64895f331784b133
- [x] Check for compatibility with newer python versions.
- [x] Twisted error: https://github.com/twisted/twisted/pull/1679
   Answer: https://github.com/deluge-torrent/deluge/pull/331/commits/e8874aab8fa0ab526845d424bdf69fe1ab398889
- [x] Edit path for openssl? Why?
       Answer: libtorrent + pyinstaller requires a lib(ssl/crypto)-1_1.dll and lib(ssl/crypto)-1_1-x64.dll odd quirk but solveable by just having two copies. Maybe later compiling our own libtorrent.
- [x] [vc-redist](https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170) check
       Answer: Not needed libtorrent isn't build static and pyinstaller grabs the redist files.
- [x] With UI/Plugin files copied to data, do not need the deluge folder.
       Answer: Not needed but comes from collect_all('deluge'):hiddenimports which is needed.


DjLegolas:
Looks great, thanks.
If you can, please add a job to the CI for automatic creating the installer. It will need to have the ability to compile everything it needs (GTK3, OpenSSL, etc.).

doadin:
Im working on that atm. The hardest part kind of is compiling GTK. Others have public windows builds. Could use a build host for this but artifacts don't last forever. And GTK takes quite a bit of time to compile.

doadin:
@DjLegolas
The best option would probably be to have a build of GTK hosted somewhere then use that plus the already available builds of others eg. just pip install libtorrent and use prebuilt openssl from the ci host, as most things are available or easy to get like with pip, GTK is the only thing really hard to get.

cas--:
@doadin I agree if we can start simple with dependencies the better, where needed we can modify it later to suit.  For GTK we could fork the gvsbuild repo and create releases or store them on deluge ftp, either way we want a simple reproducible process.

doadin:
@cas-- Yes I plan on on making an CD git git actions, since that seems to be what we are using for other things, that will making an installer using prebuilts from github actions image, which the only thing I see needed out side of that is GTK ill mkae the script to use from appveyor since thats where im currently setting up a build, should be a easy change in url to a different host depending on where you would like to get it from.

doadin:
@DjLegolas @cas-- 
It seems that github actions uses powershell to run so I added a sample CD.yml based on that which should only need a correction for the URL where we want to get GTK from. The deluge spec file already has path setup to be able to find openssl and such so unless the wheel of libtorrent from pip requires something else should be able to find all the deps deluge needs.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,114 @@
...
+          # Look to see if there is a cache hit for the corresponding requirements file
+          key: ${{ runner.os }}-pip-${{ hashFiles('tox.ini', 'setup.py', 'requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: |
+          $WebClient = New-Object System.Net.WebClient
+          $WebClient.DownloadFile("https://ci.appveyor.com/api/buildjobs/wgsvc79ov8jsy8dc/artifacts/gvsbuild-vs16-x86.zip","C:\GTK.zip")
```
DjLegolas, line 40: Doesn't this needs to be a x64 version of GTK?

DjLegolas:
BTW, great work @doadin 🥇

packaging/win/deluge-win64-installer.nsi
```diff
@@ -0,0 +1,215 @@
+# Deluge Windows installer script
```
cas--, line 1: There is only two lines difference between these nsi files so there should only be one nsi script that be be called with optional argument for 32-bit build (64-bit default?)

packaging/win/delugensis.py
```diff
@@ -0,0 +1,53 @@
...
+# Copyright (C) 2012-2015 Calum Lind <calumlind@gmail.com>
+# Copyright (C) 2010 Damien Churchill <damoxc@gmail.com>
+# Copyright (C) 2009-2010 Andrew Resch <andrewresch@gmail.com>
+# Copyright (C) 2009 Jesper Lund <mail@jesperlund.com>
+#
+# This file is part of Deluge and is licensed under GNU General Public License 3.0, or later, with
+# the additional special exception to link portions of this program with the OpenSSL library.
+# See LICENSE for more details.
+#
+# isort:skip_file
```
cas--, line 13: there should be no reason to skip isort on this file

.github/workflows/cd.yml
doadin: yes the x64 build had not completed yet so I didn't have a url. Its done now and I just woke up so I'm going to update.

packaging/win/deluge-win64-installer.nsi
doadin: I am not really familiar with NSIS, I will look into that.

packaging/win/delugensis.py
doadin: Was just a copy paste from the bbfreeze file, sorry.

doadin:
@DjLegolas  @cas-- I think I have addressed the points you guys have brought up, still uncertain about nsis use again I'm not very familiar but I think other than it being cleaner its at least close to proper syntax?

doadin:
@DjLegolas @cas--  So im really stuck its apparently not able to load a dll for gtk, which I have no clue. I have the spec file setup to add GTK\lib folder and the VC redist to path but its still missing a dll from somewhere idk.

DjLegolas:
@doadin - don't worry, we will help you with that.
Seem like there are several dlls missing:

1. python - `No such file or directory: 'c:\\hostedtoolcache\\windows\\python\\3.7.9\\x64\\DLLs\\'`. Maybe it's on another location or name?
2. make sure that the GTK lib and bin paths are in the PATH environment variable so the the python libs will find them.

doadin:
@DjLegolas 
1. I did a print of the dir and that dir does exist not sure why it says it doesn't. (side note: do wish we could have rdp to the vm like with appveyor.)
2.I am using pathex in pyinstaller to add the GTK lib and bin paths. I could try and add it in the build script vs pyinstaller adding it but should work from pyinstaller.

doadin:
@DjLegolas I tried adding the GTK dir to path from the github actions and still no go. I wish I knew what DLL it can't find would be probabbly really easy to find if I knew what I was looking for lol.

mhertz:
Really stupid question, but doesn't help using x64 gtk(for x64 build) ? It's seemingly through bindepend.py/pefile.py trying check a file related to gi(hook) , which links to gtk dlls, running matchdllarch on them, or something. Strange error regardless. As said on forum, the pathhex is irrelevant, atleast on real system, with that spec, when in PATH as djlegolas said, just to not waste time on adding there this and that. Anyway, really impressed, looking great, thanks for all the nice work.

doadin:
The script installs pycairo and pygobject from wheel from gtk build folder and is "successful" so I see no reason why it would not be available. I could be wrong but I think the missing dll isn't a product of missing gi but other way around it tries to import gi which tries to load the dll which fails ending in the importing failing.

mhertz:
Sorry doadin, thanks for explaining, and sounds like you're right. Anyway, I remember during looking through cmd output during builds, that it stated sometimes, probably unrelated, looking for ctypes in DLLs, though not specially for gi, but saw there where a ctypes pyd in that DLLs folder, so maybe check if there, if possible, though strange mentions dir not found in error, when there. Sorry for noise and I'll butt out now - best of luck you guys :)

doadin:
@mhertz I never turn down help/ideas. :)

mhertz:
Thanks mate :)

Just very quick, seemingly is issue of gdkpixbuff hook tries to find path to gi library dir it gets the wrong one and fails colecting any pixbuff-loaders - it tries to do a dll check I think, but cannot on a folder, as the error just mentions folder not file.

Anyway, I got this from looking closer at the output and googling it, where issue with little other, but slight related, very slight but regardless - It was stated that e.g. the user having issues had poisened his build environment by having gtk3/gdk dll's in current folder or next to script or something, messing that part up.

A suggestion was to check if can see the path correct by adding in spec file, just for troubleshooting... well, you can see yourself what to add, a single line printing it to cmd during freeze - I didn't yet read it through, so possibly better hints later, and again sorry if way off:

https://issueexplorer.com/issue/pyinstaller/pyinstaller/6282

doadin:
I have a theory I am testing now. Normally I believe when you install gtk python bindings you end up with a struckture something like python/lib/site-packages/gnome/gi-repository I just made a build attempt and made it print out the site-packages folder and the gi-repository which I believe normally has libs in it only had like 1 file and was a __init__.py or something like that. Im thinking maybe for some reason the wheel is incomplete. Im currently running a build on my test VM to see the folder structure.

mhertz:
When installing pygobject, then always makes gi dir under site_packages and bunch other gi dirs there. Indeed if only a __init__.py, then the issue as per error, as the _gi*.pyd not there, which tries imported - so the install of wheel went wrong, or incomplete as you state. Good catch, sorry for my ramblings before.

doadin:
I'm trying to run a test suggested in that github issue and see where it is looking for the lib then we can check that folder to confirm if the files needed are there or not.

mhertz:
The gi regardless is f'ed up - you need a _gi*.pyd there for import, and i'm pretty sure that's the root cause(if you say there only is a init.py under site_packages/gi). But sure, test that out.

doadin:
@mhertz we have that same struckture https://github.com/deluge-torrent/deluge/runs/4717092096?check_suite_focus=true#step:7:4767 it is the other folder i'm uncertain about.

https://github.com/deluge-torrent/deluge/runs/4717092096?check_suite_focus=true#step:7:4821 https://github.com/deluge-torrent/deluge/runs/4717092096?check_suite_focus=true#step:7:4821 . shows "Directory: C:\hostedtoolcache\windows\Python\3.7.9\x64\Lib\site-packages\gi\repository" with only a pycache dir and a "__init__.py" file im fairly sure there is supposed to be more. Again waiting for my VM to finish building to confirm. However if this is the case then we either have to build GTK in the build process or wait for gvsbuild to fix the wheel creation? idk.

mhertz:
Sorry doadin, I get myself confused, and forgot what yopu said initially about the import failing because missing file of that gdkpixbuff or whatever, doh :) There should only be a __init__.py there.

doadin:
with a test python script with
`from PyInstaller.utils.hooks.gi import get_gi_libdir
get_gi_libdir('GdkPixbuf', '2.0')
`
we get
`
  File "C:\test.py", line 2, in <module>
    get_gi_libdir('GdkPixbuf', '2.0')
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\PyInstaller\utils\hooks\gi.py", line 35, in get_gi_libdir
    path = findSystemLibrary(lib.strip())
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\PyInstaller\depend\bindepend.py", line 942, in findSystemLibrary
    return getfullnameof(name)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\PyInstaller\depend\bindepend.py", line 71, in getfullnameof
    if os.path.exists(npth) and matchDLLArch(npth):
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\PyInstaller\depend\bindepend.py", line 178, in matchDLLArch
    pe = pefile.PE(filename, fast_load=True)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pefile.py", line 2743, in __init__
    self.__parse__(name, data, fast_load)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pefile.py", line 2808, in __parse__
    "Unable to access file '{0}'{1}".format(fname, exception_msg)
Exception: Unable to access file 'C:\': [Errno 2] No such file or directory: 'C:\\'
`
Which is interesting 1. that pyinstaller get involved and 2. It seems to just search the dir the script is ran from?

mhertz:
Hmm, that dir is not present at all??? I don't get where gets that part from? SOrry over my head, please let me know if can do something to help test in a vm, but else hope you good people can figure it out :)

doadin:
Yea I think this is way over my head looking at https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/depend/bindepend.py#L42 it searches sys.path for dll's to which in our spec file we extend to include the gtk build folder, so how is it not finding them? Again unless its not actually the GTK libs we are missing and it is looking for msvc dll's or something to which again it would be helpful if it told us which dll it is looking for and not finding so we could add that dll's dir to out path.... but that would make this too easy I guess...

mhertz:
Yeah indeed puzzing. I have no input, but would personally try without any pathex's, as maybe trips something up, like that pyinstaller thread I posted before about poisening his own environment - in my many testings of this, I needed zero pathex's, but just all in PATHs. I just tested a spec today actually without such, and posted it in my edit to forum-thread.

Hope you succede, or get help from qualified people, thanks again!

doadin:
Im starting to actually thinking maybe this is a issue of github actions running some command as cmd and some with powershell and therefore we have different path environment variables? Im not sure thats a thing?

doadin:
@mhertz well the test script doesn't use the spec file and also fails so unfortunatly I don't think thats an issue. at least not yet.

mhertz:
Doh!

Are you setting paths like so: https://stackoverflow.com/questions/60169752/how-to-update-the-path-in-a-github-action-workflow-file-for-a-windows-latest-hos

doadin:
@mhertz you might be on to somethinig I was setting them using the normal windows method but in the Install dependencies step not the Freeze Deluge step which that post seems to suggest that they don't carry over between runs. Maybe thats it.

doadin:
@mhertz I think I also kind of just confirmed this in my last test we get

```
  File "<string>", line 2, in <module>
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\gi\__init__.py", line 47, in <module>
    from . import _gi
ImportError: DLL load failed: The specified module could not be found.
['C:\\', 'C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\python37.zip', 'C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\DLLs', 'C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\lib', 'C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64', 'C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64\\lib\\site-packages']
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pefile.py", line 2795, in __parse__
    fd = open(fname, "rb")
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\'
```

which I added print(sys.path) to the test python script and what do you know no gtk folder.

mhertz:
Glad you hopefully figured it out, and thanks for conversation - again, if fails still, please let me know if can do something to help test.

Take care.

doadin:
@DjLegolas @cas-- @mhertz It seems to be working now, the only issue is the icon file location its not finding but that worst case scenario could be hard coded to the deluge source path. However the gi imports working now. So we have made progress.

cas--:
@doadin Nice work, much appreciated :tada:

doadin:
since this is really close now a few things to note going forward
1. Depends on gtk artifact from appveyor/zip file that has the structure like so:
    gvsbuild-vs16-x64.zip/release/bin
    gvsbuild-vs16-x64.zip/release/lib
    etc etc
2. the pyinstaller spec file is coded to at set deluge.ico path specific to github actions checkout
3. I don't know how to use nsis command line args to make is produce different installers for x86 vs x64
    why was this needed? I had to nsis scripts to change the installer file name so we can provide x86 and x64 installers without confusion of which a user downloaded. and it changes the default install location to "program files" for x64 or "program files (x86)" for x86

doadin:
@cas-- something I can change now is do we want the CD to run on pull requests so we can verify changes easier? Or just on tags?

Do we want to make github releases?

mhertz:
Very quickly while you're messing with it anyway, then as per my last edit on forum-thread, using 'pip install .' like deluge official instructions states, followed by 'python setup.py install_scripts', fixes deluge-console working again, and is more proper installed anyway, without double deluge package folders(one egg one normal, mixing up). I noticed you don't include hidden_imports for gi and cairo, but I guess you've tested that this isn't needed - was previously, but maybe not now with added spec-additions. Btw, I tested spec also runs fine in packaging win32 folder, also icons support relative path in spec btw. 

You don't have to answer, I know you're busy, but just so you don't need read through my way to long edits on forum about this :)

Thanks again.

doadin:
@mhertz Adwaita icons are being used.

mhertz:
Sure doadin, and obviously your call, but I just did this to unofficial installer for 1+ years now, without errors reported, well not my builds atleast. I left the 'scalable' folder there, and index file, which has all the symbolic deluge icons as far as I know, or there would be errors in deluge-debug.exe which I checked. Honestly don't know if this is particular correct, but however does work I know. As said gi hook copies much of share over needlessly, just in case. Thanks for reply, and again, ping me if can help somehow.

doadin:
@mhertz Sorry didn't see you left scale and the index, working on nsis atm then I'll test that in a min.

cas--:
@doadin I would not want this to run on every PR (fine here for testing), tags might be useful (those without `-dev0`), if we also have a manual option to run it that is good. 

Oh the PR runs could be gated as we did for the security tests so would need to include a PR key to run it.

darkred:
/off-topic

Greetings

I'm so glad for the ongoing development of Deluge, and I've been constantly wondering: why not enabling Issues and Discussions in this repo? And, also why not migrate issue tracker from Trac to GitHub? 
I believe it would help in getting more attention to the project and more people to actively contribute in Deluge's development.
I made this thread in the forum [[Suggestions] Make use of 'GitHub Actions' + Migrate issue tracker from Trac to GitHub](https://forum.deluge-torrent.org/viewtopic.php?f=8&t=55442) 2.5 years ago, but never got a response from a dev.

*Sorry for this off-topic, but I couldn't find any other way to ask your opinion directly.*

/off-topic

doadin:
@cas-- 
> @doadin I would not want this to run on every PR (fine here for testing), tags might be useful (those without `-dev0`), if we also have a manual option to run it that is good.
> 
It should now have this

> Oh the PR runs could be gated as we did for the security tests so would need to include a PR comment value to run it.
>
Not sure about this. Also I don't know how to make the nsis change its providing a def but idk how to process it in the nsis script.

doadin:
@cas-- Also is there a version of python we want to target for builds? I currently have a github action for GTK setup to target 3.8 which should be ready to copy over if you want to. And it should be easy to change the deluge CD action as well now. I think you have test run on 3.8? Looks like python supports each new version for +1 year from the last and 3.8 is supported till 2024 so that could be a decent target. I see deluge code was updated to work with 3.10 though so we could go for newer.

cas--:
@darkred Very offtopic.. I have considered it but also don't want to be at the whim of Github...

@doadin For the windows build I am happy with almost latest Python version e.g. 3.9. We want something that is stable and not outdated as you mentioned. If it's easier to stick with 3.8 then that is perfect with me :+1: Obviously Deluge needs to support much older versions so it's Python 3.6+ now

doadin:
@cas-- 3.9 sounds like a good target its not latest but has had a year to get stable and get improvements, and also only a year old so not too dated. I'm starting a gtk build with 3.9 now to test make sure it goes ok, With how its setup now its just change the number in three places for each job in the action file and off it goes. And for the deluge action only in one spot per job.

doadin:
@cas-- As of now as long as the GTK build with python 3.9 goes ok all I need to do is adjust the url of where to get the GTK.zip build from which I'm not sure how to do releases.

doadin:
@cas-- I have not tested yet but I have found https://github.com/marketplace/actions/fetch-github-release-asset it appears we can just use that to fetch the latest build from a repo you fork so we don't have to worry about changing urls for builds of gtk. Unless you want to just use deluge servers. Its pretty much all set except this part so just need to know which direction to go.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,136 @@
...
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: |
+          $WebClient = New-Object System.Net.WebClient
+          $WebClient.DownloadFile("https://ci.appveyor.com/api/buildjobs/vggwhkyrv0cpoewf/artifacts/gvsbuild-vs16-x64.zip","C:\GTK.zip")
+          7z x C:\GTK.zip -oc:\GTK
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          $Env:Path
```
cas--, line 46: Is this a typo? Also this does not match the x86 step

.github/workflows/cd.yml
```diff
@@ -0,0 +1,136 @@
...
+          Set-Location -Path $env:GITHUB_WORKSPACE\packaging\win
+          python delugensis.py
+          makensis /Darch=x64 deluge-win-installer.nsi
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: my-artifact
+          path: packaging/win/*.exe
+
+  Build-Windows-x86:
```
cas--, line 77: Can we use [strategy matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix) to deduplicate the Build-Windows steps.

So perhaps set a `gvsbuild` for download link, `arch` key set to `x86` or `x64` and `wheel_arch` to `win32` or `amd64`  etc...

.github/workflows/cd.yml
```diff
@@ -0,0 +1,136 @@
...
+        run: |
+          $WebClient = New-Object System.Net.WebClient
+          $WebClient.DownloadFile("https://ci.appveyor.com/api/buildjobs/vggwhkyrv0cpoewf/artifacts/gvsbuild-vs16-x64.zip","C:\GTK.zip")
+          7z x C:\GTK.zip -oc:\GTK
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          $Env:Path
+          python -m pip install --upgrade pip wheel
+          pip install C:/GTK/release/python/pycairo-1.20.1-cp37-cp37m-win_amd64.whl
+          pip install C:/GTK/release/python/PyGObject-3.42.0-cp37-cp37m-win_amd64.whl
+          python -m pip install libtorrent==1.2.*
```
cas--, line 50: Can we install latest specific version to avoid surprises

cas-- (commented):
Really nice work on this

How and where are you currently building the gvsbuild, is there anything custom to it? Have they moved to github actions?

doadin:
@cas-- Sorry not sure why https://github.com/deluge-torrent/deluge/commit/23141dff826af7adf409b03bcd842c0f6fe839c1 is'nt showing as the current changes.

doadin:
> Really nice work on this
> 
> How and where are you currently building the gvsbuild, is there anything custom to it? Have they moved to github actions?
@DjLegolas @cas-- 
They use github actions but don't release builds. I was building on AppVeyor but switched to github actions so really if you fork the repo you can go with either. AppVeyor build process doesn't make github releases only AppVeyor artifacts. The github actions I updated recently so it will make a release on tag, Downloading that release I'm not sure of a better way other than just downloading from a url.

doadin:
@cas-- The github action I have for GTK currently build on any tag and now has two strategy matrix.
`
          - platform: x64
            vs_ver: 16
            pythonversion: 3.9
          - platform: x86
            vs_ver: 16
            pythonversion: 3.9
`
And the end result is a github release with two zip files in the format:
gvsbuild-pyPYTHONVERSION-vsVS_VER-ARCH.zip

doadin:
So it appears if you re-run a build the release with delete the old and upload the new so the url doesn't have to change with using github releases for GTK. Alternatively  could use something like https://github.com/dsaltares/fetch-gh-release-asset to just download the latest release.

DjLegolas:
@doadin @cas-- I think we can build everything for py3.6-3.9 (don't know if py3.10 will work, not to mention 3.11), and libtorrent 1.2 and 2.0.
It all depends on GTK to be built for all of the python versions (and it look like @doadin is already compiling it to 3.7-3.9).
As a result, we will have every supported python version with every supported libtorrent (still a WIP :sweat_smile:)

The difference is only in the `matrix` section in the `cd.yml` file:
```yml
matrix:
    pythonVersion: [3.7, 3.8, 3.9]
    libtorrent: [1.2.15, 2.0.5]
    arch: [x64, x86]
    include:
      - arch: x64
        wheel_arch: amd64
      - arch: x86
        wheel_arch: win32
```
And the GTK zip download link was moved to the `Install Dependencies` section (the only place it is needed) as follows:
```yml
GVSBUILD: "https://github.com/doadin/gvsbuild/releases/download/1-7-2022-5/gvsbuild-py${{matrix.pythonVersion}}-vs16-${{matrix.arch}}.zip"
```

cas--:
@DjLegolas As a stand-alone install we don't need to build other python versions, we pick one that is stable and is compatible with Deluge dependencies. 

It might be worthwhile having a libtorrent 2.0 for testing but not sure as a wide release :thinking:

DjLegolas:
> @DjLegolas As a stand-alone install we don't need to build other python versions, we pick one that is stable and is compatible with Deluge dependencies.
> 
> It might be worthwhile having a libtorrent 2.0 for testing but not sure as a wide release 🤔

It's a valid point, and it is you decision at the end 🥇.
Some think points:

- create a libtorrent2.0 variant which will be always marked as testing release, up until we finish with the fixes for it.
- add the matrix for the CI phase, and test on all supported python versions (>= 3.6), also on LT1.2/2.0.
- @doadin - change your `CD.yml` file for gvsbuild, for better and shorter file, to: 
   ```yml
    matrix:
        pythonversion: [3.7, 3.8, 3.9]
        vs_ver: 16
        platform: [x64, x86]
        include:
          - platform: x64
            arch: x64
          - platform: x86
            arch: win32
   ```

doadin:
@cas-- @DjLegolas 
The build process seems to work with anything python 3.7-3.9, except for with 3.7 & 3.9 the end exe's give this issue:
`
  File "deluge-script.pyw", line 33, in <module>
  File "deluge-script.pyw", line 22, in importlib_load_entry_point
  File "importlib\metadata.py", line 524, in distribution
  File "importlib\metadata.py", line 187, in from_name
importlib.metadata.PackageNotFoundError: deluge
`
From what I read this was due to a change early on in 3.9 which now seems maybe backported to 3.8.(googleing gives a lot of results from people saying 3.9 issue went to 3.8 and its fine)

doadin:
> > @DjLegolas As a stand-alone install we don't need to build other python versions, we pick one that is stable and is compatible with Deluge dependencies.
> > It might be worthwhile having a libtorrent 2.0 for testing but not sure as a wide release 🤔
> 
> It's a valid point, and it is you decision at the end 🥇. Some think points:
> 
> * create a libtorrent2.0 variant which will be always marked as testing release, up until we finish with the fixes for it.
> * add the matrix for the CI phase, and test on all supported python versions (>= 3.6), also on LT1.2/2.0.
> * @doadin - change your `CD.yml` file for gvsbuild, for better and shorter file, to:
>   ```yaml
>    matrix:
>        pythonversion: [3.7, 3.8, 3.9]
>        vs_ver: 16
>        platform: [x64, x86]
>        include:
>          - platform: x64
>            arch: x64
>          - platform: x86
>            arch: win32
>   ```

Invalid workflow file : .github/workflows/cd.yml#L17
The workflow is not valid. .github/workflows/cd.yml (Line: 17, Col: 19): Unexpected value '16'

Maybe I did it wrong but I just copy pasted.

doadin:
@mhertz I am getting https://github.com/deluge-torrent/deluge/runs/4750750505?check_suite_focus=true#step:7:326 when adding datas += [ ('deluge', 'deluge-2.0.5-py3.7.egg/deluge') ]

mhertz:
Yeah, sorry, that was from pretty old edit, and at the time I didn't really know it was supposed to build from packages/win, and so just changed to build from deluge-source dir, as you did at time, instead of my previous testing direct from Scripts dir.

Anyway, that's old as said, and I posted better solution later, so unneeded. Also as said, for fixing deluge-console, and to follow general best practise guide-lines, imho consider installing through 'pip install .' or alternatively, if must use setup.py, then I checked works through adding install command '--old-and-unmanageable', or alternatively what pip does, and use '--single-version-externally-managed --record record.txt' to setup.py.

Last, the excludes you added won't work, and will need e.g what I did I.e folders not supported there. Took me a bunch of tests to get working, though never claimed being the brightest neither :)

doadin:
@mhertz I guess I copied pasted old excludes as well. Ill go look at what you posted.

doadin:
@mhertz Thanks!

doadin:
@DjLegolas Sorry if you saw my comment I figured it out, so nvm. :)

mhertz:
doadin, i'm going to bed now, pretty late/early here in my side of the woods, but just wanted to say that in my spec/setup I have crypto dlls auto-picked up and used e.g. added a https tracker manually and checked "announce OK", which won't happen without. Just noticed you messed with the pathex/datas additions. Again, amazing work all around!

doadin:
@mhertz Unfortunately in github action environ its not picking up the x64 dll for openssl even though its in the os PATH environment variable.

mhertz:
doadin, quick note just, hope okay. I wanted to look into the stuff you mentioned regarding newer python's making issues, so just started doing little testings. I always thought probably being issues when going to py3.8 and above(but wanted wait to if you maybe asked for help regarding it), tbecause in unofficial installer I had to specifically code into __init__.py os.add_dll_directory(dir),well used *.pth initially but regardless, because of changes in how python worked in those versions, to avoid dll mismatch I also was plauged with initially with bug-reports by forum-users before py3.8. I wasen't sure if would work anyway for you, without mods, because the gvsbuild devs included os.add_dll_directory into __init__.py's of pycairo and pygobject after I raised that there where newly introduced issues, and one of the devs explained me the pitfall, and made/comitted fix still there i.e. checks PATH for needed gtk dll's and add the dirs found into os.add_dll_directory. There can still then be mismatches, because some have dirs in PATH from other apps, with said dll's, or in system32 with incompatible versions, completelly breaking deluge(system32 takes precedense always, before this os.add_dll_directory change introduced, saving me from lotta trouble at the time), so I in my scripts/builds(unofficial installer) patch that out of __init__.py of pycairo/pygobject and in deluge's __init__.py I hardcode the bin dir of gtk , which I included in app dir, as only call to os.add_dll_directory happening, as if keeping the pycairo/pygobject's __init__.py's and add mine infront or some other place, then the docs state it cannot promise load order of multiple os_add_dll_directories, hence I did it that way just in case. Anyway, I suspect you need openssl location in os_add_dll_dir, but if having it next to the libtorrent.pyd under site-packages, then works regardless, so I just had my scripts(unofficial installer) copy the two built openssl dlls to there, but could also be copied to gtk bin dir as included in os.add_dll_dir if not patching out the mods by pycairo/gobject, or if override and doing it yourself like I did. Forgive me for having a problem explaining stuff shortly :) So, initially deluge couldn't import libtorrent, even unfreezed with python 3.9.9, but after copying dll's in place next to libtorrent.pyd, works, and gets picked up through freeze because is next to the libtorrent pyd, most likely would also have worked if copied to gtk bin dir, but whatever.

I'm thinking this possibly is your github-actions openssl freezing issue also, because maybe messed with newer python's when happened, Sorry if wrong, but regardless, wanted to say that works for me atleast with my first test of py3.9.9, just waiting for freeze to finish(so PATH for openssl not enough for even unfreezed, is my point, on py3.8 and up - also with deluged not even loading the __init__.py's of pycairo/pygobject as not gtk app).

Okay finished freezing, and fails loading deluge because error stated in gi's __init__.py as paramter incorrect for add_dll_directory it states - as said it by default enumerates PATH for cairo/gi libs(cairo.dll) and adds those found to os.add_dll_directory, which on clean system is bin dir of gtk, but that is failing now - maybe pyinstaller sets up a fake environment, where such isn't allowed i.e. parsing through PATH, or using other dll's outside freezed dir, and regardless shouldn't work like that freezed obviously anyway. Regardless might work if patching that out and using e.g. my deluge __init__.py, or some other trick, but i'll check it out and post back when progress. This must also have been brought up with pyinstaller in general before i'm sure, but didn't look it up yet. Sorry again for wordyness. Last, why it's not breaking before(under py3.8) with wrong parameter error, is that it check if os.add_dll_dir attribute available, and if not skips, in __init__.py of pycairo/pygobject, so first time using that section from there now, as upped python version, so attribute available.

I'll make some testing and will make edits when hopefully finding something that works, and try to keep short, unless you beat me to it with solution of-course - and sorry haven't looked at your latest commits, so hope you haven't already fixed, and i'm speaking redundantly here. Thanks you.

packaging/win/delugewin.spec
```diff
@@ -0,0 +1,324 @@
+
+# -*- mode: python ; coding: utf-8 -*-
+from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import collect_all, copy_metadata
+
+datas = []
+binaries = []
+if is_win:
+    if "GITHUB_ACTIONS" in os.environ:
```
cas--, line 9: This seems very fragile having github actions specific lines in the spec file. Also this is added for x86 arch which probably not what we want?

What is causing pyinstaller not to include them? Could this issue not always be the case any x64 build not just github actions?

Is there a way of specifying it via argument or perhaps without full path? e.g. `libcrypto-1_1-x64.dll`

.github/workflows/cd.yml
```diff
@@ -0,0 +1,98 @@
...
+  workflow_dispatch:
+
+jobs:
+  Build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          - arch: x64
+            wheel_arch: amd64
```
cas--, line 21: Is this being used? Remove if not

I was just thinking that the actual wheel platform is `win_amd64` and a better name would have been `py_platform` since it's platform tag generated by Python.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,98 @@
...
+          #gci env:* | sort-object name
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          pyinstaller --clean $env:GITHUB_WORKSPACE\packaging\win\delugewin.spec --distpath $env:GITHUB_WORKSPACE\packaging\win\freeze
+          Get-ChildItem $env:GITHUB_WORKSPACE\packaging\win\freeze\* -Include requires.txt -Recurse | Remove-Item
+
+      - name: Make Deluge Installer
+        env:
+          ARCH: ${{ matrix.arch }}
+        run: |
+          Set-Location -Path $env:GITHUB_WORKSPACE\packaging\win
```
cas--, line 91: Can simplify to using run command `working-dir`keyword: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsrun

.github/workflows/cd.yml
```diff
@@ -0,0 +1,98 @@
...
+          python delugensis.py
```
cas--, line 92: Can we rename `delugensis.py` to something more meaningful such as `setup_nsis.py`

.github/workflows/cd.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          path: '%LOCALAPPDATA%\pip\Cache'
+          # Look to see if there is a cache hit for the corresponding requirements file
+          key: ${{ runner.os }}-pip-${{ hashFiles('tox.ini', 'setup.py', 'requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        env:
```
cas--, line 54: I dont' think we need to complicate this by setting envs for matrix options

packaging/win/deluge-win-installer.nsi
```diff
@@ -105,54 +110,24 @@ Function finishpageaction
     CreateShortCut "$DESKTOP\Deluge.lnk" "$INSTDIR\deluge.exe"
 FunctionEnd
 
-# Test if Visual Studio Redistributables 2008 SP1 installed and returns -1 if none installed
```
cas--, line 108: What is the current status of the redistributables? Will the user always have them installed?

cas--:
I just created a Deluge repo for storing the builds: https://github.com/deluge-torrent/gvsbuild-release/releases

It's just run manually for now for a single Python version but if you are having issues can extend it to build other Python versions...

doadin:
@cas-- Yes I still have not had the time to figure out the issue with newer python versions so would be nice to at least extend that to include python 3.7 so we can get builds out sooner rather than later? It's been a long time since we have seen a new windows build.

I also updated to address all but 2 of your comments, if I missed any let me know, and let me know what you thing about those two.

doadin:
3.8+ gives:

`
  File "deluge-gtk-script.pyw", line 33, in <module>
  File "deluge-gtk-script.pyw", line 22, in importlib_load_entry_point
  File "importlib\metadata.py", line 524, in distribution
  File "importlib\metadata.py", line 187, in from_name
importlib.metadata.PackageNotFoundError: deluge
`
Which from some googleing suggests using pyinstaller copy_metadata would fix but unfortunatly does not.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,93 @@
...
+
+jobs:
+  Build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          - arch: x64
+            pythonVersion: 3.7
+            gvsbuild: "https://github.com/deluge-torrent/gvsbuild-release/releases/download/latest/gvsbuild-py3.7-vs16-x64.zip"
```
cas--, line 22: Easier to maintain if move `gvsbuild` as an env to install deps section, use python and arch to populate filename.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,93 @@
...
+  workflow_dispatch:
+
+jobs:
+  Build:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          - arch: x64
+            pythonVersion: 3.7
```
cas--, line 21: Can we simplify by moving this as a matrix list:

```
      matrix:
        arch: [x64, x86]
        python: [3.7]
        libtorrent: [2.0.5]
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,93 @@
...
+          $WebClient = New-Object System.Net.WebClient
+          $WebClient.DownloadFile($env:GVSBUILD,"C:\GTK.zip")
+          7z x C:\GTK.zip -oc:\GTK
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          python -m pip install --upgrade pip wheel
+          $pycairopath = Get-Childitem –Path "C:\GTK\release\python\" -Include pycairo*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          $PyGObjectpath = Get-Childitem –Path "C:\GTK\release\python\" -Include PyGObject*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          pip install $pycairopath
+          pip install $PyGObjectpath
+          python -m pip install libtorrent==$env:LIBTORRENT
```
cas--, line 65: Can you replace this env with `${{ matrix.libtorrent }}`

.github/workflows/cd.yml
cas--: Set-Location line can be removed then

mhertz:
doadin, it doesn't here, so either is a gtk-actions issue, or which I much expect, that you don't install deluge like me. Honestly don't understand why you insist on installing through setup.py and not through pip, or add flags to setup.py that don't need you to delete requirements.txt as workaround for it installs as an incomplete egg which needs extra files imported which results in two deluge packages confusing everything etc, not to mention breking deluge-console, not the most important on windows i'd speculate, but nonetheess.

Your call of-course, and i'll go ahead and make the PR gvsbuild asked me to, in mean time.

doadin:
@mhertz I've been following https://dev.deluge-torrent.org/wiki/Installing/Source#BuildingandInstalling for years now.

While it does fix the issue with 3.8+ it does not fix openssl dll missing.
It does also seem to change deluge.common.get_version() 

builds come out like deluge-2.0b2.dev521-win64-setup
instead of deluge-2.1.0.dev15-win32-setup

how this is python 3.x related when it runs a git command idk. All I can do is report what I'm seeing.

doadin:
@mhertz despite this version change I have updated the build process to include those changes as it cleans up the build script.

mhertz:
Thanks for reply, and I sincerly apologize my ton in last post, didn't mean it in any way negative honestly, and for the record am not in doubt about you being much ahead of me regarding development, to say in the slightest :) I have only tested stable deluge, and will make edits if finding something, to use for you or not of-course at your discretion - thank you :)

doadin:
@mhertz Ill be the first to admit I don't know very much I am just trying to help the community.

doadin:
@cas-- In addition to the changes I mention earlier from your review, I have added changes @mhertz has provided and we can use python 3.8+ now. Also the build process is cleaner and simpler. The only adjustment left is the spec file libssl-x64 include. Which I'm not sure why exactly it is needed. github actions has the x64 lib in its $env:Path so why pyinstaller is only getting the 32-bit bin and not the x64 I'm not sure, maybe pyinstaller searches system32 first and it finds a libssl and uses that and moves on? I'm not sure.

doadin:
@mhertz the only change I have not pulled in is for removeing the share folders and honestly I an not sure what you have it doing and why it needs to run on every script.

mhertz:
Indeed, bothered me too :) Of-course your call. As said, I think ran 50 runs to even get that working, though probably an issue with me more than anything else. But it doesn't help filtering a.datas, when we have b-h.datas with those icons anyway, getting merged at end - maybe could do something with toc(), but I just got tired of it when having found something working. It goes through all collected files for each step, x.datas, and "rewrites" it through checking if matches with the excludes filter, and if not, then adds, and if matches, then strips/avoids - reverse filtering of sorts. No biggie of course, and i'm sure could be done better, e.g. the straight datas, like on top, but couldn't make work. Also, maybe Cas/you think sit's too agressive I dunno, only found it worked as far as could see myself, and has svg decoder in gdkpixbuff loader through librsvg so why include all the pngs additiopnally, but what do I know

Thanks :)

doadin:
@mhertz well now im seeing:
deluge-debug.exe:
`
PS E:\Program Files\Deluge> .\deluge-debug
  File "deluge-debug-script.py", line 33, in <module>
  File "deluge\ui\ui_entry.py", line 39, in start_ui
    ui_entrypoints[entrypoint.name] = entrypoint.load()
  File "pkg_resources\__init__.py", line 2449, in load
  File "pkg_resources\__init__.py", line 2472, in require
  File "pkg_resources\__init__.py", line 772, in resolve
pkg_resources.DistributionNotFound: The 'mako' distribution was not found and is required by the application
[44432] Failed to execute script 'deluge-debug-script' due to unhandled exception!
`
deluge.exe:
`
  File "deluge-script.pyw", line 33, in <module>
  File "deluge\ui\ui_entry.py", line 39, in start_ui
    ui_entrypoints[entrypoint.name] = entrypoint.load()
  File "pkg_resources\__init__.py", line 2449, in load
  File "pkg_resources\__init__.py", line 2472, in require
  File "pkg_resources\__init__.py", line 772, in resolve
pkg_resources.DistributionNotFound: The 'certifi; sys_platform == "
`

All the other exes work.

mhertz:
doadin, i'm sorry, when I said "working" I meant in the previous py-versions - I did say that I haden't tested all exe's yet in py3.8=+, but sorry if formulated badly - probably - sorry.
.
There are some strange issues - first deluge-console seemingly works if after installing deluge unfreezed, then also installing windows-curses - that didn't need happen on <py3.8. Then when that is installed, then works, I think, need test more, but shows some initial error about libintl and referenced before assignment, also deluge-debug states such - though seemingly works - I just think need intl dll and whatnot other dll in main dir, that I used to copy there with unofficial installer, but will take a look.

Last, atleast unfreezed, then now have deluged showing window open all time, like the debug.exe should - I remember having this issue along time ago, and got patches by tobbez fixing this, so maybe needed here again, I dunno - hmm, no I didn't use such in last unofficial build - now i'm getting confused, will write back later, and sorry if making issues for you doadin, didn't mean to.

cas--:
I'm sorry I'm finding it very hard to keep track of these conversations so helping out is tricky. 

Can we try and keep a track of exactly what is working and what is not and be very specific about each issue and what the potential fixes are. Also when making commit updates it would be useful to actually see what and why changes are being made rather than rebase to single commit. It might also be useful for workarounds in the code to have terse comments about what is being fixed.

@doadin would you mind updating the list of tasks remaining in the PR description and tick them off as done. We can then decide if they are a priority or not. 

We want a minimum viable stable installer for Windows therefore don't need to chase latest versions:

* If there are too many issues with Python 3.9 then don't work on it, use Python 3.7. 
* We only need to verify that the `x86` installer works since it's compatible with `x64` thus concentrate on that. 
* I would also prefer if we used libtorrent 1.2 rather than 2.0 again for stability
* I don't think reducing the installer size is a priority and can come later

@mhertz Please provide a bit more information on error and fixes by providing traces, code snippets and links to files. 

Avoid sharing links to issues that are not github.com since those are sham sites making money via ad revenue e.g. giters.com and also the linking between github issues fails. 

The deluge-console should work on Windows using windows-curses now

doadin:
@cas--  I'm just going to do one last rebase to go back to what we had/what you reviewed if we find a way to improve on that, that doesn't break things then we can go from there. This new way broke two things in an attempt to just use a newer python version. Just going to scrap that and go with 3.7.

doadin:
@cas-- ok from here everything is how you last reviewed.

mhertz:
I apologize Cas.

Py3.9.9 with my spec and install, has no issues in freezed version as far as I see now.

In unfreezed is having issue of reporting intl dl translation issuel, but works, though as kinda off-topic here I guess, when not in freezed, then I didn't post output, can on request.

Only thing is that the freezed deluged works because I defined console=false in spec - bit it ends in py, not pyw, so default windows exe loader/wrapper loads through python.exe not pythonw.exe making cmd window open in unfreezed- again I guess of-topic as not in freezed.

All 8 exe's tested working.

Deluge-console only works non-interactive, for me, in all my tests, both here and in unofficial, and in "real install".

Spec used: https://paste.c-net.org/BethesdaTermite

I add unfreezed errors just in case, but again, no issue in freezed version.

````
Unable to initialize gettext/locale!
local variable 'libintl' referenced before assignment
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 143, in setup_translation
    libintl = load_libintl(['libintl-8.dll', 'intl.dll'])
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 132, in load_libintl
    if not libintl:
UnboundLocalError: local variable 'libintl' referenced before assignment
20:32:41 [ERROR   ][deluge.i18n.util  :155 ] Unable to initialize gettext/locale!
20:32:41 [ERROR   ][deluge.i18n.util  :156 ] local variable 'libintl' referenced before assignment
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 143, in setup_translation
    libintl = load_libintl(['libintl-8.dll', 'intl.dll'])
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 132, in load_libintl
    if not libintl:
UnboundLocalError: local variable 'libintl' referenced before assignment
````
````
C:\Users\fooba>deluge-console info
Unable to initialize gettext/locale!
local variable 'libintl' referenced before assignment
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 143, in setup_translation
    libintl = load_libintl(['libintl-8.dll', 'intl.dll'])
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 132, in load_libintl
    if not libintl:
UnboundLocalError: local variable 'libintl' referenced before assignment
20:34:34 [ERROR   ][deluge.i18n.util                                      :155 ] Unable to initialize gettext/locale!
20:34:34 [ERROR   ][deluge.i18n.util                                      :156 ] local variable 'libintl' referenced before assignment
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 143, in setup_translation
    libintl = load_libintl(['libintl-8.dll', 'intl.dll'])
  File "C:\Python\lib\site-packages\deluge\i18n\util.py", line 132, in load_libintl
    if not libintl:
UnboundLocalError: local variable 'libintl' referenced before assignment
[S]   100% archlinux-2022.01.01-x86_64.iso e368633c136582fc3eab4a77b1c75c36151a8eef
    DL: 899.2 M (0 B) UL: 17.9 M (0 B) ETA: -


C:\Users\fooba>
````

**Comment by mhertz** [comment]
I'm sorry, didn't see doadin/your reply, sorry guys, i'll but out and stop derailing, and only comment upon request. Thanks and sorry.

**Comment by doadin** [comment]
@mhertz 
I don't think anyone really minds help but I had a build that worked only think not tested was deluge-console which din't work in deluge 1.3.x anyways so imo not a huge lose(had to lose something you didn't really have as far as most users are concerned). The only thing really needed is deleting the requires.txt not really a big deal imo. Nothing done here has to hack/slash gtk or exe loaders or anything.

@cas-- 
Updated to use libtorrent 1.2.15

**Comment by doadin** [comment]
I think the only reason why the openssl dll addition is needed in spec is because of the mass amounts of programs in Path that have a version of those dll's. So we can leave it as is or I can reduce the Path env in script to only include dirs we need. Either way I think a change is needed and the one we have isn't that bad. Not my call though I'll do what ever.

**Comment by doadin** [comment]
@cas--  @mhertz Something I have found out. Which I updated the tasks.
3.8+ Delete Deluge\deluge-2.1.0.dev15.dist-info\METADATA: Requires-Dist lines to work
3.7 Delete Deluge\deluge-2.1.0.dev15.egg\requires.txt to work.

**Comment by mhertz** [comment]
Good job finding what works for you :)

The errors I posted in unfreezed was fixed in freezed because intl.dll was in the gtk-dir added through os.add_dll_directory() and gets included in freezed main-dir. 

Btw, in your gvsbuild-scripts, could use something like:
````
curl https://gitlab.gnome.org/GNOME/pygobject/-/commit/5535efb58a67a4782c17707fea8c7a8f1dfb2900.patch > gvsbuild\patches\pygobject\pygobject_py3_8_load_dll.patch
curl https://github.com/pygobject/pycairo/commit/69850152b5fb4fd7cfdf08a009fa6aa327911f5b.patch > gvsbuild\patches\pycairo\pycairo_py3_8_load_dll.patch
````
Untill added in gvsbuild or upstream. Powershell translated.

Thanks.

**Comment by doadin** [comment]
@mhertz I havn't had any issues with the build scripts, however if we wanted we could use that first patch idea applied to the .spec file to add the openssl-x64 dll's to build and remove the 32-bit ones maybe. And that could maybe be a less peticular system dependent change unlike my check for if running on github actions.

**Comment by mhertz** [comment]
Yes, try that - I thought I would need an os.add_dll_directory addition in spec for including openssl, but didn't need it for some reason.

I don't understand why you don't need patch the pycairo/pygobject, as is the thing that breaks deluge on py3.8+, unless you have no semicolon in PATH ending(echo %PATH%) - and from the error you posted yesterday, I believe you have. 

Regardless whatever you can make it work with :)

**Comment by doadin** [comment]
@mhertz I think you hit the nail on the head there I think its a OS Path setup issue not a gtk issue.

**Comment by mhertz** [comment]
I have also openssl dlls without 64 in name in my freezed dir, didn't think you where talking about that. This means that pyopenssl from python got seen and imported with deps, and not the x64 we build ourself only for libtorrent's usage - we need both, or else e.g. yarss will break, unless rebuilding without asm(--no-asm) and copy the x64 dlls to not have 64 in name, but why bother. Of-course you need the x64, i'm not saying why bother to that.

**Comment by doadin** [comment]
@cas-- The opensll dll's it is picking up is in fact a x64 bin it is just not named as such. How we want to deal with this im not sure. However if we just name them as libtorrent expects we don't need that line in spec file.

**Comment by doadin** [comment]
From my testing I believe pyinstaller is picking up the dll from C:\hostedtoolcache\windows\Python\3.7.9\x64\DLLs which despite being x64 does not include that in the name. From testing I see that in that folder for x86 and x64 the name remains the same simply libssl-1_1.dll. It also appears openssl 1.1 is used from python 2.7-3.9 newer versions may use a different openssl version I'm not sure.

**Comment by doadin** [comment]
@cas-- It appears to be an issue with the libtorrent build with the x64 variant expecting two sets of openssl dlls.

**Comment by doadin** [comment]
Using a dependency walker it shows only x64 dll needed yet if lib(ssl/crypto)-1_1.dll is removed it fails to import libtorrent. This must be a pyinstaller issue I guess?

**Comment by cas--** [comment]
@doadin Thanks for looking into this, I like that fix. We should let libtorrent know about this too but might be tricky to resolve without breaking it for someone...

> It also appears openssl 1.1 is used from python 2.7-3.9 newer versions may use a different openssl version I'm not sure.

As long as we document it, we can revisit the issue

**Comment by doadin** [comment]
@cas-- yea a dependency walker shows the libtorrent.pyd as only needing the x64 so it might actually be a pyinstaller hook maybe? I'm not sure.

**Comment by doadin** [comment]
If we change:
cd.yml:
    - Under Install Deluge:
          change the two lines to:
          pip install .
          python setup.py install_scripts
      - Under Freeze Deluge:
         remove Get-ChildItem $env:GITHUB_WORKSPACE\packaging\win\freeze\* -Include requires.txt -Recurse | Remove-Item
delugewin.spec:
change lines 20 to:
datas += [ ('../../deluge/ui', 'deluge/ui'),
           ('../../deluge/plugins', 'deluge/plugins') ]

This is what we get:

Python 3.9:
deluge.exe & Deluge-debug.exe:

```
  File "deluge-script.pyw", line 33, in <module>
  File "deluge\ui\ui_entry.py", line 39, in start_ui
    ui_entrypoints[entrypoint.name] = entrypoint.load()
  File "pkg_resources\__init__.py", line 2449, in load
  File "pkg_resources\__init__.py", line 2472, in require
  File "pkg_resources\__init__.py", line 772, in resolve
pkg_resources.DistributionNotFound: The 'pyopenssl' distribution was not found and is required by the application
```

Unless you remove dist lines from deluge-(build version number).dist-info METADATA
If you remove just pyopenssl it will just fail stating another was not found
All Other exe's are fine as far as starting.

deluge-console:
Starts but seems to freeze with this error seen:

```
  File "twisted\internet\base.py", line 1331, in mainLoop
  File "twisted\internet\selectreactor.py", line 101, in doSelect
  File "twisted\internet\selectreactor.py", line 39, in win32select
builtins.OSError: [WinError 10038] An operation was attempted on something that is not a socket
Unhandled Error
  File "deluge\ui\console\main.py", line 161, in start_ui
    wrapper(self.run)
  File "curses\__init__.py", line 94, in wrapper

  File "deluge\ui\console\main.py", line 282, in run
    reactor.run()
  File "twisted\internet\base.py", line 1318, in run
  File "twisted\internet\base.py", line 1331, in mainLoop
  File "twisted\internet\selectreactor.py", line 101, in doSelect
  File "twisted\internet\selectreactor.py", line 39, in win32select
builtins.OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

Python 3.7:
All exe's fail with:

```
  File "deluged-script.py", line 11, in <module>
  File "pkg_resources\__init__.py", line 490, in load_entry_point
  File "pkg_resources\__init__.py", line 2862, in load_entry_point
  File "pkg_resources\__init__.py", line 2461, in load
  File "pkg_resources\__init__.py", line 2484, in require
  File "pkg_resources\__init__.py", line 787, in resolve
pkg_resources.DistributionNotFound: The 'zope.interface' distribution was not found and is required by the application
```
Unless you remove dist lines from deluge-(build version number).dist-info METADATA
If you remove just pyopenssl it will just fail stating another was not found

deluge-console:
Starts but seems to freeze with this error seen:

```
  File "twisted\internet\base.py", line 1331, in mainLoop
  File "twisted\internet\selectreactor.py", line 101, in doSelect
  File "twisted\internet\selectreactor.py", line 39, in win32select
builtins.OSError: [WinError 10038] An operation was attempted on something that is not a socket
Unhandled Error
  File "deluge\ui\console\main.py", line 161, in start_ui
    wrapper(self.run)
  File "curses\__init__.py", line 94, in wrapper

  File "deluge\ui\console\main.py", line 282, in run
    reactor.run()
  File "twisted\internet\base.py", line 1318, in run
  File "twisted\internet\base.py", line 1331, in mainLoop
  File "twisted\internet\selectreactor.py", line 101, in doSelect
  File "twisted\internet\selectreactor.py", line 39, in win32select
builtins.OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

**Comment by doadin** [comment]
I know that was really long but we can get a wider compatibility with those changes + what we have now. I think some of the other issues where due to going back and forth with pyinstaller changes.

**Comment by doadin** [comment]
Still running tests to double check my work but I think 1066eb69340bc386a57950a502585245d2267ac2 is a solution to getting 3.8 and 3.9 builds. I don't have gtk 3.8 builds to test but iv'e been testing with 3.7 and 3.9.

**Comment by doadin** [comment]
updated first post/task list with more info.

**Comment by mhertz** [comment]
Forgive me for speaking uninvited, but really hard continuing holding in helpful info honestly. I'll be short/consize from now on(if invited), promize.

* Metadata needs patched only because missing metadata, both in "requires.txt" issue of py3.7, and "metadata" of py3.8+, and avoided entirely by in spec adding said metadata back through:
````
datas += copy_metadata('deluge', recursive=True)
````
* I made PRs at gvsbuild + upstream, but untill merged, then can add yourself, if deemed neccesary. It breaks deluge starting if in same dir as deluge and having trailing semicolon in PATH, as some installers add - test e.g. from freezed dir:
````
set path=%path%;
deluge
````
https://github.com/wingtk/gvsbuild/issues/475
* deluge-console works only non-interactive, just like deluge 1.3.x, sorry if thinking I ment all worked.
*  Unimportant, but the openssl without x64 comes from python itself/pyopenssl, and the x64 named is for libtorrent, so no pyinstaller/libtorrent issue(should report to python to fix there unusual naming). Good job fixing by copying over! As said, deleting the x64 from system32 also fixing this, but better way from doadin imho.
* If ever finding more files missing/issues, then use the collect_all for deluge line from docs I always did, though currently doadin's way better, as saves 1mb install size.

Hope OK, but will stop post uninvited again now. Thanks doadin, awesome job as usual! :)

**Comment by doadin** [comment]
@mhertz Thanks, I will test this. I have been trying to get on irc more, if you would like we could work on this more there. Though honestly seems pretty close to a good change imo.

**Comment by mhertz** [comment]
Thanks alot doadin, appreciate your post alot! I don't really have more to add honestly, never used IRC honeslty neither, so please just ping me if you think I might now/test something, or feel free to always email me, or deluge-forum PM. I'll let you finish your awesome job in peace. Take care :)

**Comment by doadin** [comment]
I tested 8 builds python 3.7-3.10 both x86 and x64 and found only the same issue which exists with old and new method.
Plugins Execute & Extractor:
```
09:58:44.350 [ERROR   ][deluge.pluginmanagerbase          :78  ] Unable to instantiate plugin 'Execute' from 'e:\\program files\\deluge\\deluge\\plugins\\execute-1.3-py3.10.egg'!
09:58:44.350 [ERROR   ][deluge.pluginmanagerbase          :78  ] No module named 'twisted.internet.utils'
  File "deluge\pluginmanagerbase.py", line 147, in enable_plugin
  File "e:\program files\deluge\deluge\plugins\execute\deluge_execute\__init__.py", line 14, in __init__
    from .core import Core as _pluginCls
  File "e:\program files\deluge\deluge\plugins\execute\deluge_execute\core.py", line 14, in <module>
    from twisted.internet.utils import getProcessOutputAndValue
ModuleNotFoundError: No module named 'twisted.internet.utils'
```

**Comment by mhertz** [comment]
I apologize doadin, I tested many of the plugins finally worked, including some 3'rd party ones, and was so happy that forgot test each and every one doh!

The last error is not important and also unfreezed, I reported some time ago here with own fix: https://dev.deluge-torrent.org/ticket/3476

**Comment by doadin** [comment]
@mhertz All good, I did the same thing :) . In changing so many things I saw the plugins I use enabled and didn't test though enough. Thats why I went ahead and made builds and tested python 3.7-3.10 x86 and 64, I don't want leave anything untested. Appreciate the link to the ticket.

**Comment by mhertz** [comment]
I'll narrow it down and make short edit, but just posting quick to not waste your time if you're also running testings. All plugins work with these added, and as said some is redundant obviously:
````
tmp_ret2 = collect_all('twisted')
datas += tmp_ret2[0]; binaries += tmp_ret2[1]; hiddenimports += tmp_ret2[2]
datas += copy_metadata('twisted', recursive=True)

````
About your fix, thanks mate, and for asking me, and check it out in a bit, though i'm not the one to ask as not "good" like you guys, but much appreciate you asked me :)

**Comment by doadin** [comment]
@mhertz I think as far as icons goes the biggest, safest one to remove I think it cursors. We will have to lokk into or just leave em. Its part of the theme and who knows what random plugin or something might use an icon from the theme so idk.

**Comment by doadin** [comment]
@mhertz Really just down to one more Item needing fixed. Thanks so much for you help!

**Comment by cas--** [comment]
@doadin I have just fixed the plugin description issue https://dev.deluge-torrent.org/ticket/3476  so no longer need bd3502bb447

I have also fixed the `libintl` error that was reported by @mhertz d2390cd24729 however that would only occur if dll was missing... :slightly_smiling_face:

**Comment by doadin** [comment]
@cas--  I just noticed, ill update this. Things are looking pretty complete still need to do some cleanup. Mostly being the plugin fix adding data hopefully can at least slim it down. And still waiting for that twisted issue to get fixed I really droped the ball there, but we can just use a fixed build of our own for now not a huge deal.

**Comment by doadin** [comment]
@cas-- As long as everything is fine your ok with how its all done, other than cleaning the commits up I think this is ready.

**Comment by cas--** [comment]
I'll review it tomorrow but a quick skim seems good to merge! I've setup my windows vm again to check how it looks

Thanks for all the effort in resolving this and the details provided for each problem has been really useful 👍

**Comment by mhertz** [comment]
> I have also fixed the `libintl` error that was reported by @mhertz [d2390cd](https://github.com/deluge-torrent/deluge/commit/d2390cd247291ed428e6707e19d362d499cfd3ed) however that would only occur if dll was missing... 🙂
@cas, thanks! :) I'm running vanilla here, no mods, and since py3.8+ then without me adding os.add_dll_directory() then not finding these dlls from gtk(PATH irrelevant) - the gtk py-bindings runs that call, but libintl requested before that, or so I'm theorizing atleast, as sure have dll. Thanks again.

**Comment by doadin** [comment]
> I'll review it tomorrow but a quick skim seems good to merge! I've setup my windows vm again to check how it looks
> 
> Thanks for all the effort in resolving this and the details provided for each problem has been really useful 👍 

Cool. I'm not gonna make any more changes other than cleaning up the number of commits.

**Comment by doadin** [comment]
@cas-- I combined some of the commits that aimed to do the same thing so its done in 5 commits now vs 12 so its still seperate to see the process that happened to implement fixes, idk if you like this better or not or want it reduced even more? Or maybe this is worse? I saved a copy of how it was in a branch on my fork if you don't like it like this.

**Comment by doadin** [comment]
the only thing not separate is the process we went through to be able to go from py3.7 to newer other than that its still all separate so the process can be seen, just in fewer steps instead of going back and forth and having commits just changing libtorrent versions.(things I think didn't need to be separate got squished) Again I have a copy of how it was before though if this is not better.

**Comment by DjLegolas** [comment]
@doadin @mhertz first of all, great work guys. The long waited installer is an important milestone for Deluge 2.0 🥇 💯 

Regarding only the commits, I would suggest to add to the reason for the patched `twisted` to the commit message because without it, one would never know why it was needed.

And again, thanks for this (>130 comments 🤣) MR!

File: .github/workflows/cd.yml [hunk 16]
```diff
@@ -0,0 +1,86 @@
...
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          python -m pip install --upgrade pip wheel
+          $pycairopath = Get-Childitem –Path "C:\GTK\release\python\" -Include pycairo*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          $PyGObjectpath = Get-Childitem –Path "C:\GTK\release\python\" -Include PyGObject*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          pip install $pycairopath
+          pip install $PyGObjectpath
+          pip install https://github.com/doadin/twisted/releases/download/latest/Twisted-21.7.0.post0-py3-none-any.whl
+          python -m pip install libtorrent==${{ matrix.libtorrent }}
+          pip install -r requirements.txt
+          pip install pyinstaller
```
CirnoT, line 59: Any reason this is done separately from `libtorrent`?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,86 @@
...
+          $WebClient = New-Object System.Net.WebClient
+          $WebClient.DownloadFile("https://github.com/deluge-torrent/gvsbuild-release/releases/download/latest/gvsbuild-py${{ matrix.python }}-vs16-${{ matrix.arch }}.zip","C:\GTK.zip")
+          7z x C:\GTK.zip -oc:\GTK
+          $env:Path = "C:\GTK\release;C:\GTK\release\bin;C:\GTK\release\lib;$env:Path"
+          python -m pip install --upgrade pip wheel
+          $pycairopath = Get-Childitem –Path "C:\GTK\release\python\" -Include pycairo*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          $PyGObjectpath = Get-Childitem –Path "C:\GTK\release\python\" -Include PyGObject*.whl -File -Recurse -ErrorAction SilentlyContinue | select -expand FullName
+          pip install $pycairopath
+          pip install $PyGObjectpath
+          pip install https://github.com/doadin/twisted/releases/download/latest/Twisted-21.7.0.post0-py3-none-any.whl
```
CirnoT, line 56: While I have nothing against it by itself, this is introducing dependency on binary release that is outside of Deluge control. We should patch it ourselves or keep patched repository under `deluge-torrent` organization.

CirnoT:
I have not checked the version that adds `hiddenimports` from `Twisted`, however can we be reasonable sure that plugins such as YaRSS2 will work? For example, YaRSS2 for v2 depends on `http.cookies` via `requests` and `email.mime.*`. Both of these are part of stdlib for Python and are presumed to be always available, however pyinstaller does not treat them as such and produced binaries might not have necessary modules available.

This was an issue with v1 too, back then YaRSS2 had its own custom code for `email.mime.*` part for Windows users, however since release of v2 it started to depend on various 3rd-party modules such as `requests`, which bring their own dependencies that plugin does not necessarily control. To my knowledge they've done best job they could trying to package these external dependencies into egg they provide, however of course Python's stdlib is out of their control.

---

The other issue that remains is how broken GTK3 is on Windows. Seeing some weird issues where clicking Deluge icon on taskbar does bring window to front but doing so again does not minimize it as one would expect. Using `GTK_CSD=1` is a solution to this (and many more). We should of course never modify user's environment globally when installing Deluge, so the question remains whether there is some way we can have Deluge disable decorations when running on Windows without relying on environment variables.

mhertz:
I specified e.g gtk_csd=1 other place before, but my understanding was that it wasent global and only enherented for running proces, and not transferred. Wrong?

cas--:
@CirnoT Thanks for suggestions :+1: 

I know that for the dependency issues (e.g. email.mime) were manual fixed in bbfreeze and py2app: https://github.com/deluge-torrent/deluge/blob/d2390cd247291ed428e6707e19d362d499cfd3ed/packaging/win32/deluge-bbfreeze.py#L84 https://github.com/deluge-torrent/deluge/blob/d2390cd247291ed428e6707e19d362d499cfd3ed/setup.cfg#L21

---

Can you more specific about Python2 breaking changes? I probably will have to create a 2.0.x release branch  anyway so certainly possible to release 2.0.5/6 for Windows.

---

> Using GTK_CSD=1 is a solution to this 

Did you mean `GTK_CSD=0`? This is only a env var. Perhaps could be done in [runtime-hook](https://pyinstaller.readthedocs.io/en/stable/usage.html?#cmdoption-runtime-hook)?

CirnoT:
> Did you mean GTK_CSD=0? This is only a env var. Perhaps could be done in runtime-hook?

Yes. If its possible to set environment variable for executable via `runtime-hook` in pyinstaller than that would be solution.

> Edit: Always been somewhat leery though, as stated to possibly have a performance penalty, didnt notice any though, but hard to judge, when only quick testing.

Less performance on drawing items is in my opinion preferable to broken interaction with Windows DWM.

> but my understanding was that it wasent global and only enherented for running proces, and not transferred. Wrong?

I am not sure where you set it. Environment variables are usually either set when running process or globally for user/computer (in which case it would apply to everything). If as mentioned above `runtime-hook` can do it for pyinstaller then it would indeed be isolated.

> Can you more specific about Python2 breaking changes? I probably will have to create a 2.0.x release branch anyway so certainly possible to release 2.0.5/6 for Windows.

Some plugins (in this example YaRSS2) still retain some level of compatibility for Python 2, even on releases that are meant for Deluge 2. See https://bitbucket.org/bendikro/deluge-yarss-plugin/issues/67/deluge-210-removed-all-py2-support for details. The common offenders will most likely be `PY2` (from `deluge.common`) and `six` (if plugin decided to use it, but unlikely).

Perhaps we could keep PY2 as a dumb stub (`PY2 = false`) to make plugins happy?

mhertz:
Very quick, as not liking intrude on your important things, but just remembered that I otherplace added requests(not stdlib) and pygame, from pip, when installing deluge from pip, so springjools's autoremoveplus fork can even be enabled/used(don't know if relevant to you), and notifications default plugin works for playing sound. Don't need to respond. Thanks for everything you good people do :)

CirnoT:
> added requests(not stdlib) and pygame, from pip, when installing deluge from pip, so springjools's autoremoveplus fork can even be enabled/used(don't know if relevant to you)

Appreciated to know, however non-stdlib requirements are something plugin developers must solve themselves, otherwise users are forced to take manual steps (like you did), so this is not something we can accommodate when packaging Windows version. For example, YaRSS2 depends on `requests` too but they do vendor it themselves.

cas--:
> Perhaps we could keep PY2 as a dumb stub (PY2 = false) to make plugins happy?

Done via https://github.com/deluge-torrent/deluge/commit/6da4c4bf6667cfb0b6f0d4dfbc8ab6d2be5ef09a

cas--:
Btw the job will not run by default on PRs but I added a condition if labelled with 'windows' so can try that out to see if it works...

DjLegolas:
About the stdlib, why not use the `Windows embeddable package` available from python?
It exists exactly for this purpose.

doadin:
@cas-- Sorry, I just woke up was going to make a commit with some of the requested changes, was just commenting and saw you merged.

CirnoT:
> About the stdlib, why not use the `Windows embeddable package` available from python?
> It exists exactly for this purpose.

That is not compatible with pyinstaller, or are you suggesting that we should abandon pyinstaller and create our own stub executables using embeddable package?

PR #332, opened 2022-01-04
[AutoAdd] Fixed error dialog not being shown on error

DjLegolas:
This issue might be the cause [3069](https://dev.deluge-torrent.org/ticket/3069).

### DIFF
--- a/deluge/plugins/AutoAdd/deluge_autoadd/gtkui.py
+++ b/deluge/plugins/AutoAdd/deluge_autoadd/gtkui.py
@@ -324,7 +324,7 @@ class OptionsDialog:
             dialogs.ErrorDialog(_('Incompatible Option'), str(ex), self.dialog).run()
 
     def on_error_show(self, result):
-        d = dialogs.ErrorDialog(_('Error'), result.value.exception_msg, self.dialog)
+        d = dialogs.ErrorDialog(_('Error'), result.value.message, self.dialog)
         result.cleanFailure()
         d.run()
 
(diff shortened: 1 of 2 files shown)


cas-- (changes requested):
We can just replace with `message` attribute since `exception_msg` was removed with changes to RPC protocol: 9b812a4eec6205b98

DjLegolas:
> We can just replace with `message` attribute since `exception_msg` was removed with changes to RPC protocol: [9b812a4](https://github.com/deluge-torrent/deluge/commit/9b812a4eec6205b98d9d5c12c14f82611832217a)

I saw it in [another place](https://github.com/deluge-torrent/deluge/blob/44dcbee5f43b454994b3a6173bdd27802120e707/deluge/ui/console/cmdline/commands/connect.py#L61) so I copied the behavior in case it still exists.
Change that too?

cas--:
Yep change there too :+1:

cas--:
That looks fine, do you mind updating the commit message to mention the reason for change and reference commit id 9b812a4

cas-- (approved):

DjLegolas:
> That looks fine, do you mind updating the commit message to mention the reason for change and reference commit id [9b812a4](https://github.com/deluge-torrent/deluge/commit/9b812a4eec6205b98d9d5c12c14f82611832217a)

Hope it is good enough 😄

cas--:
That was exactly what was needed :+1: 

Remember to add a reference to the Trac ticket in commit messages too, something like this:

image

I automatically add the pr close line to simplify workflow so don't worry about that.

PR #333, opened 2022-01-05
[TrackerIcon] fixed icon download issues

DjLegolas:
This PR contains 2 commits:

1. Fixed parse error on UTF-8 sites with non-english chars
I got the exception 
```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 2158: character maps to <undefined>
```
2. Fixed old-large icon removal 
After successful download of the icon, it wasn't shown in the UI.
Only after closing it I got, for each of the icons, the exception:
```
builtins.PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\user\\AppData\\Roaming\\deluge\\icons\\site.ico'
```

closes: [#3479](https://dev.deluge-torrent.org/ticket/3479)

### DIFF
--- a/deluge/tests/test_tracker_icons.py
+++ b/deluge/tests/test_tracker_icons.py
@@ -14,7 +14,6 @@ from . import common
 from .basetest import BaseTestCase
 
 common.set_tmp_config_dir()
-deluge.ui.tracker_icons.PIL_INSTALLED = False
 common.disable_new_release_check()
 
 
@@ -41,6 +40,15 @@ class TrackerIconsTestCase(BaseTestCase):
         # So instead we'll grab its favicon.ico
         icon = TrackerIcon(common.get_test_data_file('google.ico'))
         d = self.icons.fetch('www.google.com')
+        d.addCallback(self.assertNotIdentical, None)
+        d.addCallback(self.assertEqual, icon)
+        return d
+
+    def test_get_google_ico_hebrew(self):
+        # Google.co.il page must be read as UTF-8
+        # So we'll test whether it can be read
+        icon = TrackerIcon(common.get_test_data_file('google.ico'))
+        d = self.icons.fetch('www.google.co.il')
         d.addCallback(self.assertNotIdentical, None)
         d.addCallback(self.assertEqual, icon)
         return d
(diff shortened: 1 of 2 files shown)


deluge/ui/tracker_icons.py
```diff
@@ -289,7 +294,14 @@ def parse_html_page(self, page):
         :returns: a Deferred which callbacks a list of available favicons (url, type)
         :rtype: Deferred
         """
-        with open(page) as _file:
+        if chardet:
+            with open(page, 'rb') as _file:
```
cas--, line 298: If it's an HTML page it should not need to be opened in binary mode so can remove `'rb'` option

deluge/ui/tracker_icons.py
```diff
@@ -289,7 +294,14 @@ def parse_html_page(self, page):
...
+                result = chardet.detect(_file.read())
+                encoding = result['encoding']
+        else:
+            encoding = 'UTF-8'
```
cas--, line 302: We can drop the else statement if encoding is specified at top of function

cas-- (changes requested):
Nice work, just some small changes :+1:

deluge/ui/tracker_icons.py
DjLegolas: Without `'rb'`, python will try to open it with the default encoding again. It will lead to a `UnicodeDecodeError` error again on Windows.
And tested it also on ubuntu under WSL, which didn't have that problem.
Here is the section of `encoding` param of the `open` function from the docstring:
```
In text mode, if encoding is not specified the encoding used is platform
dependent: locale.getpreferredencoding(False) is called to get the
current locale encoding.
```
When running it from Linux:
```python
>>> locale.getpreferredencoding(False)
'UTF-8'
```

When running it from (my) Windows:
```python
>>>locale.getpreferredencoding(False)
'cp1252'
```
So I'm not sure removing this will help.

cas--: Okay no problem, I did wonder if I was missing something :smile:

deluge/tests/test_tracker_icons.py
```diff
@@ -45,6 +44,15 @@ def test_get_google_ico(self):
         d.addCallback(self.assertEqual, icon)
         return d
 
+    def test_get_google_ico_hebrew(self):
+        # Google.co.il page must be read as UTF-8
+        # So we'll test whether it can be read
```
cas--, line 49: I fixed this when merging but it's better to use a docstring when describing the test (I know others tests are using a comment here)

PR #334, opened 2022-01-06
[Gtk] Fixed magnets not showing the prefetched metadata

DjLegolas:
it is a combination of 2 errors:
1. Decompress issue due to UnicodeDecodeError in `transfer.py`
2. The `info_hash` of the magnet file is being saved as `str`. But after the
prefetch is done, we get it from the `core`/`RPC` as `bytes`.

Closes: deluge-torrent/deluge#334

### DIFF
--- a/deluge/ui/gtk3/addtorrentdialog.py
+++ b/deluge/ui/gtk3/addtorrentdialog.py
@@ -263,6 +263,7 @@ class AddTorrentDialog(component.Component):
     def _on_uri_metadata(self, result, uri, trackers):
         """Process prefetched metadata to allow file priority selection."""
         info_hash, metadata = result
+        info_hash = info_hash.decode()  # info_hash is bytes, we need it to be Unicode
         log.debug('magnet metadata for %s (%s)', uri, info_hash)
         if info_hash not in self.prefetching_magnets:
             return
--- a/deluge/transfer.py
+++ b/deluge/transfer.py
@@ -115,17 +115,21 @@ class DelugeTransferProtocol(Protocol):
             self._message_length = 0
             self._buffer = b''
 
-    def _handle_complete_message(self, data):
+    def _handle_complete_message(self, data, decode_utf8=True):
         """
         Handles a complete message as it is transferred on the network.
 
-        :param data: a zlib compressed string encoded with rencode.
+        Args:
+            data: a zlib compressed string encoded with rencode.
+            decode_utf8: try to decode as utf8. Default: True
 
         """
         try:
             self.message_received(
-                rencode.loads(zlib.decompress(data), decode_utf8=True)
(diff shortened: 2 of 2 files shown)


cas--:
Can you provide an example magnet for this issue, I think there might be a deeper problem to solve here

mhertz:
It's only issue in thinclient mode, not standalone.

Thanks alot DjLegolas, works great, much appreciated :)

deluge/ui/gtk3/addtorrentdialog.py
```diff
@@ -263,6 +263,7 @@ def add_from_files(self, filenames):
     def _on_uri_metadata(self, result, uri, trackers):
         """Process prefetched metadata to allow file priority selection."""
         info_hash, metadata = result
+        info_hash = info_hash.decode()  # info_hash is bytes, we need it to be Unicode
```
cas--, line 266: This change will break standalone mode

cas--:
The root of the problem is the sending of mixed str and bytes metadata dict to rencode. This can be solved by pickling the metadata dict and then encoding with base64 and reversing that process in the UI.

DjLegolas:
@cas-- actually, seem as the metadata itself is only bytes inside the dict.
The info_hash is the only one that is being sent as a str/bytes, depending on standalone/thinclient.

deluge/ui/gtk3/addtorrentdialog.py
DjLegolas: a possible solution is:
```python
info_hash, metadata = result
if isinstance(info_hash, bytes):
    # in thinclient mode info_hash is bytes, we need it to be Unicode
    info_hash = info_hash.decode()
```

cas--:
Sorry what I meant is it is a mix of raw bytes and encoded strs so there is no way to determine type when receiving without trying to catch the exception. 

It's certainly bad practice for the API calls to randomly switch types like this and we should be decoding as early as possible in the Python code when receiving data.

Further thought is to not bdecode the metadata in _on_uri_metadata. Instead send that metadata b64encoded and the client should use deluge.bencode.bdecode after b64decode.

It's up to you if you want to make that change, I can do it this weekend otherwise

Btw really don't worry about adding the GitHub PR close link, it's something I do from my end, just the Trac issue is required for future reference.

DjLegolas:
@cas-- tested your fix, still have an issue.
```
23:45:38.538 [WARNING ][deluge.transfer                   :72  ] Failed to decompress (26945 bytes) and load serialized data with rencode: 'utf-8' codec can't decode byte 0x87 in position 2: invalid start byte
```

cas--:
Hmm okay. Did you update the daemon to latest code? I tested it here without issue so would need more information about how to replicate that

PR #273, opened 2019-12-22
gtk3ui and macos fixes

Lord-Kamina:
This fixes some standing issues with accelerators on macOS, as well as makes it easier to produce a working app bundle using macports (I made my changes in a new script and bundle-file, in case people prefer jhbuild still; it shouldn't be too hard to mimic what I did for macports to be used with jhbuild)

Of note, it appears in the future it would be desirable to drop GtkosxApplication altogether and use Gtk.Application instead but I'm not sure I am familiar enough with either python nor GTK3 to make those changes myself.

P.S. You can take my .app for a spin here: https://www.dropbox.com/s/zy0y4u41bvp6huk/Deluge.app.zip?dl=1

PR #335, opened 2022-01-09
Update Mac Packaging

doadin:
Currently will produce a atrifact zip that contains similar structure to windows install.
If you extract that zip inside are binaries and those binaries run fine.

Currently trying to work out how to make a ".app".

### DIFF
--- a/.github/workflows/cd-mac.yml
+++ b/.github/workflows/cd-mac.yml
@@ -1,0 +1,65 @@
+name: CD-Mac
+
+on:
+  push:
+    tags-ignore:
+      - "*.dev0"
+  pull_request:
+    branches:
+      - develop
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  Build-Mac:
+    runs-on: macos-10.15
+    strategy:
+      matrix:
+        arch: [x64]
+        python: [3.9]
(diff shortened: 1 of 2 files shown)


cas--:
We already have a solution for [creating an .app](https://github.com/deluge-torrent/deluge/pull/273) so let's not make using pyinstaller a priority

doadin:
@cas--  Well, I figured out the issue and this build process works better than on windows so, idk. Up to you what you want. Its here if you want it or I can just close this. However its working now with little to no workarounds or anything like that. Other than some normal spec file things it just deletes requirments.txt from the dist. That other option is not merged and uses a different build system I think Its simpler to use pyinstaller for both, It can even use one single pyinstaller spec file if you want.(only seperate currently for testing)

cas--:
That's good to know, I just wanted to concentrate on one platform for now

doadin:
@cas-- I am going to go ahead and close this since I'm taking a break. Currently it builds and everything runs fine on 10.15 from the dist folder however the .app it makes I could not get to run not sure if its just user error or not I don't develop on mac very much so really not sure. I updated with some improvements learned from the windows builds. The only thing not working is the deluge exe is commented out even though it would run. The problem was it makes the .app fail to build I believe due to having the same name as the dist folder, If it was done on a case sensitive FS would probably be ok.

PR #336, opened 2022-01-11
Correct _entry_points for deluged/deluge-web on Windows

mhertz:
Changed from console_scripts to gui_scripts to not open cmd-prompt when run

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -481,10 +481,15 @@ if not windows_check() and not osx_check():
 
 _entry_points['console_scripts'] = [
     'deluge-console = deluge.ui.console:start',
-    'deluge-web = deluge.ui.web:start',
-    'deluged = deluge.core.daemon_entry:start_daemon',
 ]
-if windows_check():
+if not windows_check():
+    _entry_points['console_scripts'].extend(
+        [
+            'deluge-web = deluge.ui.web:start',
+            'deluged = deluge.core.daemon_entry:start_daemon',
+        ]
+    )
+else:
     _entry_points['console_scripts'].extend(
         [
             'deluge-debug = deluge.ui.ui_entry:start_ui',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -495,6 +493,8 @@ def run(self):
 _entry_points['gui_scripts'] = [
     'deluge = deluge.ui.ui_entry:start_ui',
     'deluge-gtk = deluge.ui.gtk3:start',
+    'deluge-web = deluge.ui.web:start',
```
cas--, line 496: This affects all OS's so this change is not correct. 

Is this not a pyinstaller issue though? Please double-check as they are certainly not gui_scripts

cas-- (changes requested):

setup.py
mhertz: I apologize Cas, really stupid indeed, sorry. Not issue in pyinstaller, infact I work around there by adding 'console=false' in spec for deluged/deluge-web, so only seen unfreezed. Yeah sounds strange to define gui_scripts for those, but if 'console_scripts' then set to run through 'python.exe' instead of 'pythonw.exe', effectively making them debug-exe's(cmd-prompt opens upon double-click). Anyway, I tested on windows and linux, and works as far I can see, but sure you can do prettier/better, so if that is the case, kindly just close PR and whenever time/motivation permits, then add yourself, in proper way, thanks, as not wanna waste your time going back/fourth here. Side-note, I found issue with deluge.exe and deluge-gtk.exe not starting on windows, deluge-debug,exe does though - It happens without my change, and only on develop, not master(stable) - If finding what's the problem will try make PR, though as seen, not very good at it, sorry again.

cas--: This is from the [entry point](https://packaging.python.org/en/latest/specifications/entry-points/) spec and would affect output from these applications but we do have the debug versions for that...

> The difference between console_scripts and gui_scripts only affects Windows systems. console_scripts are wrapped in a console executable, so they are attached to a console and can use sys.stdin, sys.stdout and sys.stderr for input and output. gui_scripts are wrapped in a GUI executable, so they can be started without a console, but cannot use standard streams unless application code redirects them. Other platforms do not have the same distinction.

I am thinking that this might have been intentional to keep them as console_scripts, I did mess around with it when making changes to bbfreeze: 1c3e14919f

mhertz: Thanks for explaining cas, much appreciated! I understand, somewhat tricky situation, and again, maybe just close this and implement yourself correct change(s), if deemed needed, when ready, instead of extra time spent going back and fourth with me, on top :) Your call of-course, and thanks again.

PR #338, opened 2022-01-15
IP Check Updates

doadin:


doadin:
@cas-- Should be good to go, except might not want https://github.com/deluge-torrent/deluge/pull/338/commits/0e9c0606679570a0f6c1ba7a01866dd844713e05

doadin:
@cas-- I think it would be nice to get this & f89f28f81624c8eabcc612a3acff4a7606bd2235 merged if possible for the windows release.

doadin (commented):
This works toward a greater plan which.

doadin:
@cas-- I think this is ready except this would bring the min python version to 3.4. Idk if we want to continue to work with older versions, 3.4 went eol March 18, 2019.

doadin:
@cas-- I expanded the interface check to use regex to match in case ifaddr is not available but regex only works on windows so, im not sure about a failback for linux/mac.
@DjLegolas would you mind reviewing?

deluge/common.py
```diff
@@ -974,6 +974,66 @@ def is_ipv6(ip):
...
+            regex = '^{[0-9a-zA-Z]{8}-([0-9a-zA-Z]{4}-){3}[0-9a-zA-Z]{12}}$'
+            try:
+                for iface in interface.split(','):
+                    if re.search(regex, str(iface)):
+                        log.info(
+                            'Interface found with regex! Could not verify real interface'
+                        )
+                    else:
+                        log.warning('Interface not found with regex!')
+                        raise Exception('Interface not valid!')
```
DjLegolas, line 1028: the `log.warning` call seem redundant as in the `except Exception` below we print the same line, but with the exception info.

deluge/common.py
```diff
@@ -938,12 +938,13 @@ def is_ipv4(ip):
     import socket
 
     try:
-        if windows_check():
-            return socket.inet_aton(ip)
-        else:
-            return socket.inet_pton(socket.AF_INET, ip)
+        for interface in ip.split(','):
```
DjLegolas, line 941: why split the `ip`? are we expecting it to have several options?
the same goes for `is_ipv6` function

DjLegolas (commented):
Added some comments from a quick review.
In addition, I noticed that there are a lot of interface checks of `if is_ip(inter) or is_net_interface(inter)`. Maybe create a new function to check both? Just like `is_ip` itself

deluge/common.py
doadin: Yes for multi-homing.

doadin: We could do the for loop else where where we need to verify interfaces/ips I guess.

DjLegolas: I asked because running it on the entire string will fail in case it contain several types of interfaces:
```
>>> is_ipv4('127.0.0.1,eth01')
Unable to verify IPv6 Address with socket.
False
```

doadin: True. Would it work to extend is_ip to return return is_ipv4(ip) or is_ipv6(ip) or is_net_interface(ip)

doadin: No that would still fail. Hmmm not sure how to fix this.

doadin: I think I need to move the for loop out of the ip_* etc checks.

DjLegolas: yep. maybe the outer function I suggested in the opening message.
For example (the name needs to be changed 😅):
```python
def verify_ip_or_interface(ip):
    interfaces = []
    for interface in ip.split(','):
        if is_ip(interface) or is_net_interface(interface):
            interfaces.apped(interface)
    return ','.join(interfaces)
```

DjLegolas:
BTW, I think it will be best to also add tests for the changes.

deluge/common.py
doadin: Honestly I think Multi-Homing is above my understanding.

DjLegolas: don't worry, eventually we will get it also :)

doadin:
@DjLegolas Maybe for now I will leave out multi-homing support just add support for interfaces. I wanted to do both so we don't have to edit the ip checks multiple times but at least at the moment I can't see how to get it done.

deluge/common.py
doadin: I added a test but still not sure how to clean up if is_ip(inter) or is_net_interface(inter)

doadin:
@DjLegolas the interface check using ifaddr works cross-platform however I don't have nor do I know a way to do a failback for linux/mac.

DjLegolas:
no problem. let's focus on verifying the linux interfaces w/o `ifaddr`.
One way I found is getting a list of [available interfaces](https://man7.org/linux/man-pages/man5/sysfs.5.html) like so `os.listdir('/sys/class/net/')`.
Need to check if available on other systems as well.

Or maybe, relay on `ifaddr` being always available as it is a dependency.

doadin:
@DjLegolas turns out with python >=3.3 we can use socket.if_nameindex() this is not available will 3.8 for windows.

deluge/common.py
doadin: @DjLegolas would something like https://github.com/deluge-torrent/deluge/pull/338/commits/7384191554606094c94813effce5270e314d7219 work?

deluge/common.py
```diff
@@ -938,12 +948,10 @@ def is_ipv4(ip):
...
     try:
-        if windows_check():
-            return socket.inet_aton(ip)
-        else:
-            return socket.inet_pton(socket.AF_INET, ip)
+        socket.inet_pton(socket.AF_INET, ip)
+        return True
     except OSError:
-        return False
+        log.warning('Unable to verify IPv4 Address with socket.')
```
cas--, line 954: Should not be logging a warning and needs to return `False` since:

> If the IP address string ip_string is invalid, OSError will be raised.

deluge/common.py
```diff
@@ -938,12 +948,10 @@ def is_ipv4(ip):
     import socket
```
cas--, line 948: I think we should move this to top of the module

deluge/common.py
```diff
@@ -961,24 +969,82 @@ def is_ipv6(ip):
...
+        bool: Whether Interface is valid or not.
+
+    Examples:
+        >>> is_net_interface("eth0")
+        True
+        >>> is_net_interface("{7A30AE62-23ZA-3744-Z844-A5B042524871}")
+        True
+
+    """
+    if windows_check() and not interface.isupper():
```
cas--, line 998: This is not something that a common function should care about and should not modify either

deluge/common.py
```diff
@@ -900,6 +900,16 @@ def free_space(path):
         return disk_data.f_bavail * block_size
 
 
+def verify_ip_or_interface(ip):
```
cas--, line 903: I think we should standardise on the naming. So interface in libtorrent terms is an IP or Adapter therefore `is_interface` would suffice perhaps?

cas--:
I agree let's do just one thing at a time in the PR. I would also suggest that we optionally include `ifaddr` instead of trying to hack the problem ourselves, fallback to `True` and assume user is correct. Remember that libtorrent will fail if the interface doesn't exist

deluge/common.py
```diff
@@ -961,24 +969,82 @@ def is_ipv6(ip):
...
-        import ipaddress
-    except ImportError:
+        socket.inet_pton(socket.AF_INET6, ip)
+        return True
+    except OSError:
+        log.warning('Unable to verify IPv6 Address with socket.')
+    return False
+
+
+def is_net_interface(interface):
```
cas--, line 982: Continuing from previous comment how about `is_adapter`?

deluge/common.py
doadin: I figured it would make it easyer on people to avoid simple errors and depending on how they find the network guid it may or may not provide it in upper and lead to confusion I could see people thinking "im putting it in right, why is deluge broken?" hence why it also logs a message so people know what it should be. Your call though ill take it out.

cas--: I understand but we can do that verification in the UI where we can easily provide feedback to the user.

deluge/common.py
```diff
@@ -961,23 +960,50 @@ def is_ipv6(ip):
...
+    """
+    if is_ipv4(interface) or is_ipv6(interface):
+        return False
+    try:
+        import ifaddr
+    except ImportError:
+        log.error('Unable to verify interfaces. Module ifaddr missing')
+        return False
+    adapters = ifaddr.get_adapters()
+    interface = interface.encode('utf-8')
```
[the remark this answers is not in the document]
doadin: Alternatives? Cause it doesn't work when its not. So not sure what you mean by shouldn't be.

doadin:
@cas-- 
> I agree let's do just one thing at a time in the PR (just the interface/adapter checks). I would also suggest that we optionally include `ifaddr` instead of trying to hack the problem ourselves, fallback to `True` and assume user is correct. Remember that libtorrent will fail if the interface doesn't exist

By hack do you mean the regex? Would you like me to remove that and just return true if importing ifaddr fails?

deluge/common.py
```diff
@@ -961,24 +970,77 @@ def is_ipv6(ip):
...
-        import ipaddress
-    except ImportError:
+        socket.inet_pton(socket.AF_INET6, ip)
+    except OSError:
+        return False
+    else:
+        return True
+
+
+def is_interface(interface):
```
cas--, line 983: This is the wrong way round `is_interface` should be `is_adapter`. An interface is either an IP address or adapter id

deluge/common.py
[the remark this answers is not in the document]
cas--: So this should be `str` and it looks like there is platform bug in ifaddr 0.1.7 since it's returning `bytes` on Windows for adapter name, not good... https://github.com/pydron/ifaddr/pull/32

Can install that specific commit with:

```
pip install https://github.com/pydron/ifaddr/archive/37cb5334f392f12811d38d90ec891746e3247c76.zip
```

deluge/common.py
```diff
@@ -962,23 +995,53 @@ def is_ipv6(ip):
...
-    except ImportError:
-        import socket
+        socket.inet_pton(socket.AF_INET6, ip)
+    except OSError:
+        return False
+    else:
+        return True
+
+
+def is_adapter(interface):
```
cas--, line 1005: The param should be named `adapter`

deluge/common.py
```diff
@@ -962,23 +995,53 @@ def is_ipv6(ip):
...
-            return ipaddress.IPv6Address(decode_bytes(ip))
-        except ipaddress.AddressValueError:
-            pass
-
-    return False
+        if ifaddr:
+            try:
+                adapters = ifaddr.get_adapters()
+                for adapter in adapters:
+                    if interface in adapter.name:
```
cas--: Is this correct, I think 'in' would be error prone vs equality check?

deluge/common.py
```diff
@@ -962,23 +995,69 @@ def is_ipv6(ip):
...
+
+    Examples:
+        >>> is_adapter("eth0")
+        True
+        >>> is_adapter("{7A30AE62-23ZA-3744-Z844-A5B042524871}")
+        True
 
+    """
+
+    if linux_check() or osx_check():
```
cas--, line 1022: Can we put ifaddr check first.

Also with an OSError we are unable to determine adapter status so have to return True.

We could include a fallback for Windows with your regex to do a pattern match 

So something like:
```python
if ifaddr:
    try:
        adapters = ifaddr.get_adapters()
    except OSError:
         return True
    else:
         return any([adapter == a.name for a in adapters])

if linux_check() or osx_check():
    try:
           adapters = socket.if_nameindex()
    except OSError:
         return True
    else:
         return any([adapter == a[1] for a in adapters])

if windows_check():
     regex = '^{[0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12}}$'
     return  re.search(regex, str(interface))

return True

File: deluge/common.py [hunk 13]
```diff
@@ -900,6 +909,33 @@ def free_space(path):
         return disk_data.f_bavail * block_size
 
 
+def is_interface(ip):
```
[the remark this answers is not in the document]
doadin: The coma split is for multi-homing support while we did decide to focus on one thing in this PR I think it would be better kept as is so we don't have to re-write this again later.

deluge/common.py
```diff
@@ -44,6 +45,14 @@
 
     os.environ['SSL_CERT_FILE'] = where()
 
+# On windows sock.if_nameindex() is only supported on 3.8+ and does not return a uuid.
+# Try ifaddr to verify network interfaces.
+if platform.system() in ('Windows', 'Microsoft'):
```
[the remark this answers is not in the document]
doadin: later on we do linux_check() or osx_check() to not run ifaddr on linux or mac as it should not be needed since those platforms can use socket.if_nameindex().

deluge/common.py
doadin: Would it not be better to reply first on built-in vs third party? I would think third party would be less reliable, which is why I have it only using that on windows because windows we dont have a built-in, at least not yet.

cas--: Okay, there are pros and cons to third party libs but I'll concede here that we could put ifaddr after. I've updated the snippet to also pass on OSError so that it tries `ifaddr` if it's available and also rather than testing for linux/macos we just do negative check for windows to exclude it

deluge/common.py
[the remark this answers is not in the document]
cas--: Sorry please remove and we'll create another function for that as needed

doadin:
@cas-- I think I have merged all the requested changes.

deluge/common.py
```diff
@@ -44,6 +45,13 @@
 
     os.environ['SSL_CERT_FILE'] = where()
 
+# On windows sock.if_nameindex() is only supported on 3.8+ and does not return a uuid.
```
cas--, line 48: Can you put this comment in the commit message

deluge/common.py
```diff
@@ -44,6 +45,13 @@
...
+# Try ifaddr to verify network interfaces.
+try:
+    import ifaddr
```
cas--, line 51: We need to mention the optional dependency in DEPENDS.md and add it setup.py extras

deluge/core/core.py
```diff
@@ -164,19 +164,25 @@ def __init__(
         # store the one in the config so we can restore it on shutdown
         self._old_listen_interface = None
         if listen_interface:
-            if deluge.common.is_ip(listen_interface):
+            if deluge.common.is_adapter(listen_interface):
```
cas--, line 167: This should use `is_interface`

deluge/core/core.py
```diff
@@ -164,19 +164,25 @@ def __init__(
...
-                    'Invalid listen interface (must be IP Address): %s',
+                    'Invalid listen interface (must be IP Address or Interface Name): %s',
                     listen_interface,
                 )
 
         self._old_outgoing_interface = None
         if outgoing_interface:
-            self._old_outgoing_interface = self.config['outgoing_interface']
-            self.config['outgoing_interface'] = outgoing_interface
+            if deluge.common.is_adapter(outgoing_interface):
```
cas--, line 178: Should use `is_interface`

cas-- (commented):
Thanks for updating, nearly sorted

deluge/common.py
doadin: for some reason the git ui im using is not letting me.

.github/workflows/cd.yml
```diff
@@ -66,6 +66,7 @@ jobs:
           libtorrent==${{ matrix.libtorrent }}
           pyinstaller
           pygame
+          https://github.com/pydron/ifaddr/archive/37cb5334f392f12811d38d90ec891746e3247c76.zip
```
cas--, line 69: This can be put in the requirements.txt with `sys_platform == 'win32'` to only install for Windows

deluge/common.py
doadin: nvm I got it to work.

doadin:
@cas-- not sure about working for https://github.com/deluge-torrent/deluge/pull/338/commits/9e977506e544c5cf1a6526e3ad86812efa5e8bf5

DEPENDS.md
```diff
@@ -28,6 +28,7 @@ All modules will require the [common](#common) section dependencies.
 - [setproctitle] - Optional: Renaming processes.
 - [Pillow] - Optional: Support for resizing tracker icons.
 - [dbus-python] - Optional: Show item location in filemanager.
+- [ifaddr] - Optional: Help verify network interfaces.
```
cas--, line 31: We can just say `Verify ...`

deluge/common.py
```diff
@@ -44,6 +45,13 @@
 
     os.environ['SSL_CERT_FILE'] = where()
 
+# On windows sock.if_nameindex() is only supported on 3.8+ and does not return a uuid.
+# Try ifaddr to verify network interfaces.
```
cas--, line 49: Please remove both the comments as they don't add anything here. If anything the usage of ifaddr would be mentioned in the function.

We can look at the git commit if we need to know why it was added

deluge/common.py
```diff
@@ -900,6 +908,29 @@ def free_space(path):
         return disk_data.f_bavail * block_size
 
 
+def is_interface(interface):
+    """Check if interface is a valid IP or network adapter.
+
+    Args:
+        ip (str): The IP to test.
```
cas--, line 915: This docstring needs correcting to from `ip` to `interface`

deluge/common.py
```diff
@@ -962,23 +990,51 @@ def is_ipv6(ip):
...
+        >>> is_adapter("{7A30AE62-23ZA-3744-Z844-A5B042524871}")
+        True
+
+    """
+
+    if not windows_check():
         try:
-            return ipaddress.IPv6Address(decode_bytes(ip))
-        except ipaddress.AddressValueError:
+            adapters = socket.if_nameindex()
```
cas--, line 1019: Can be simplified `if_nametoindex(adapter)` and replace in else statement any() with True

deluge/common.py
```diff
@@ -962,23 +990,51 @@ def is_ipv6(ip):
...
+        try:
+            adapters = ifaddr.get_adapters()
+        except OSError:
+            return True
+        else:
+            return any([adapter == a.name for a in adapters])
+
+    if windows_check():
+        regex = '^{[0-9A-Z]{8}-([0-9A-Z]{4}-){3}[0-9A-Z]{12}}$'
+        return re.search(regex, str(adapter))
```
cas--, line 1035: Oh re.search should be wrapped in bool()

deluge/ui/gtk3/glade/preferences_dialog.ui
```diff
@@ -2573,8 +2573,8 @@ used sparingly.</property>
                                       <object class="GtkEntry" id="entry_interface">
                                         <property name="visible">True</property>
                                         <property name="can_focus">True</property>
-                                        <property name="tooltip_text" translatable="yes">The IP address of the interface to listen for incoming bittorrent connections on. Leave this empty if you want to use the default.</property>
-                                        <property name="max_length">15</property>
+                                        <property name="tooltip_text" translatable="yes">IP addresses / network interface names to be listened for incoming bittorrent connections. Leave this empty if you want to use the default.</property>
```
cas--, line 2576: This is using plural but this change is only for singular. Could also be shortened


`IP address or network interface name to listen for incoming BitTorrent connections. Leave empty to use system default.`

deluge/ui/gtk3/glade/preferences_dialog.ui
```diff
@@ -2587,7 +2587,7 @@ used sparingly.</property>
                                   <object class="GtkLabel" id="label110">
                                     <property name="visible">True</property>
                                     <property name="can_focus">False</property>
-                                    <property name="label" translatable="yes">Incoming Address</property>
+                                    <property name="label" translatable="yes">Incoming Interfaces</property>
```
cas--, line 2590: Singular just now please

deluge/ui/gtk3/glade/preferences_dialog.ui
```diff
@@ -2812,9 +2812,9 @@ used sparingly.</property>
                                         <property name="visible">True</property>
                                         <property name="can_focus">True</property>
                                         <property name="tooltip_text" translatable="yes">
-The network interface name or IP address for outgoing BitTorrent connections. (Leave empty for default.)
+                                          IP addresses / network interface names for outgoing bittorrent TCP connections. (Leave empty for default.)
```
cas--, line 2815: See listen_interface comment

deluge/ui/gtk3/preferences.py
```diff
@@ -671,11 +671,15 @@ def set_config(self, hide=False):
             'chk_random_outgoing_ports'
         ).get_active()
         incoming_address = self.builder.get_object('entry_interface').get_text().strip()
-        if deluge.common.is_ip(incoming_address) or not incoming_address:
+        if deluge.common.is_adapter(incoming_address):
```
cas--, line 674: Should not remove the empty check and use is_interface

deluge/ui/gtk3/preferences.py
```diff
@@ -671,11 +671,15 @@ def set_config(self, hide=False):
...
             new_core_config['listen_interface'] = incoming_address
-        new_core_config['outgoing_interface'] = (
+        outgoing_address = (
             self.builder.get_object('entry_outgoing_interface').get_text().strip()
         )
+        if deluge.common.is_adapter(outgoing_address):
```
cas--, line 679: Same as incoming

cas-- (changes requested):

doadin:
@cas-- I think I have addressed all the changes you have requested.

deluge/common.py
```diff
@@ -900,6 +906,29 @@ def free_space(path):
         return disk_data.f_bavail * block_size
 
 
+def is_interface(interface):
+    """Check if interface is a valid IP or network adapter.
+
+    Args:
+        interface (str): The IP to test.
```
DjLegolas, line 913: The IP or interface name to test

deluge/tests/test_common.py
```diff
@@ -116,6 +118,19 @@ def test_is_ipv6(self):
...
+        else:
+            self.assertTrue(is_interface_name('lo'))
+
+    def test_is_interface(self):
+        if windows_check():
+            self.assertTrue(is_interface('127.0.0.1'))
+        else:
+            self.assertTrue(is_interface('lo'))
+            self.assertTrue(is_interface('127.0.0.1'))
+
```
DjLegolas, line 133: Make sure there are both `assertTrue` and `assertFalse` for all cases of tests

requirements.txt
```diff
@@ -13,3 +13,4 @@ windows-curses; sys_platform == 'win32'
 zope.interface>=4.4.2
 distro; 'linux' in sys_platform or 'bsd' in sys_platform
 pygeoip
+https://github.com/pydron/ifaddr/archive/37cb5334f392f12811d38d90ec891746e3247c76.zip; sys_platform == 'win32'
```
DjLegolas, line 16: Maybe add also `ifaddr; sys_platform != 'win32'` for other OSs

doadin:
@cas-- @DjLegolas I'm not sure how to improve the tests since windows is currently the only one that will return false, also we can't really test for true since we don't know what the guid will be on the VM its run on. Do we want ifaddr for other platforms too in requirements?

DjLegolas:
About the tests, you can have something like this:
```python
def get_windows_interface_name(self):
    import winreg

    # find a network card in the registery
    with winreg.OpenKey(
        winreg.HKEY_LOCAL_MACHINE,
        r'SOFTWARE\Microsoft\Windows NT\CurrentVersion\NetworkCards'
    ) as key:
        self.assertTrue(winreg.QueryInfoKey(key)[0] > 0)  # must have at least 1 network card
        network_card = winreg.EnumKey(key, 0)
    # get GUID of network card
    with winreg.OpenKey(
            winreg.HKEY_LOCAL_MACHINE,
            fr'SOFTWARE\Microsoft\Windows NT\CurrentVersion\NetworkCards\{network_card}'
    ) as key:
        for i in range(5):
            value = winreg.EnumValue(key, i)
            if value[0] == 'ServiceName':
                interface_name = value[1]
    return interface_name

def test_is_interface_name(self):
    if windows_check():
        interface_name = self.get_windows_interface_name()
        self.assertFalse(is_interface_name('2001:db8:'))
        self.assertFalse(is_interface_name('{THIS0000-IS00-ONLY-FOR0-TESTING00000}'))
        self.assertTrue(is_interface_name(interface_name))
    else:
        self.assertTrue(is_interface_name('lo'))
        self.assertFalse(is_interface_name('127.0.0.1'))
        self.assertFalse(is_interface_name('eth01101'))

def test_is_interface(self):
    if windows_check():
        interface_name = self.get_windows_interface_name()
        self.assertTrue(is_interface('127.0.0.1'))
        self.assertTrue(is_interface(interface_name))
        self.assertFalse(is_interface('127'))
        self.assertFalse(is_interface('{THIS0000-IS00-ONLY-FOR0-TESTING00000}'))
    else:
        self.assertTrue(is_interface('lo'))
        self.assertTrue(is_interface('127.0.0.1'))
        self.assertTrue(is_interface('127.'))
        self.assertTrue(is_interface('eth01101'))
```

doadin:
@DjLegolas That would only work if we use ifaddr on linux as well for tests.

DEPENDS.md
```diff
@@ -28,6 +28,7 @@ All modules will require the [common](#common) section dependencies.
 - [setproctitle] - Optional: Renaming processes.
 - [Pillow] - Optional: Support for resizing tracker icons.
 - [dbus-python] - Optional: Show item location in filemanager.
+- [ifaddr] - Optional: Verify network interfaces.
```
DjLegolas, line 31: Add at the end of the file `[ifaddr]: https://pypi.org/project/ifaddr/` to make it a link.

DjLegolas (changes requested):

cas--:
Thanks for your work on this @doadin

gazpachoking:
Hmm, so a couple tests added to the suite are failing for me (on Windows 10.) When ifaddr is installed, these two tests fail:
```
self.assertTrue(is_interface(interface_name))
self.assertTrue(is_interface_name(interface_name))
```
When ifaddr is not installed, these two fail:
```
self.assertFalse(is_interface('{THIS0000-IS00-ONLY-FOR0-TESTING00000}'))
self.assertFalse(is_interface_name('{THIS0000-IS00-ONLY-FOR0-TESTING00000}'))
```

doadin:
@gazpachoking for that first one I wonder what the output of https://github.com/doadin/deluge/blob/39c6a77dc9d3130869bb690a2d545f24164efb9c/deluge/tests/test_common.py#L144 is.
maybe https://github.com/doadin/deluge/blob/39c6a77dc9d3130869bb690a2d545f24164efb9c/deluge/tests/test_common.py#L138 is a problem with using the first adpater maybe its a loopback or something that ifaddr does not consider.

when ifaddr is not installed it should fail to regex which those are formated correctly. So actually they should be true.

doadin:
@gazpachoking did you test with ifaddr from requirments.txt?

gazpachoking:
I actually just saw that and reinstalled with the same issue.

gazpachoking:
ifaddr is giving a list of 5 adapter strings, but the one retrieved from the registry via test isn't in it

gazpachoking:
Here's what interface_name is successfully reading as
image
And here's the list of adapters returned by ifaddr
image
Intel(R) Ethernet Controller (2) I225-V is my actual addapter, and shows up in both, just different ids.

gazpachoking:
Oh wait, the `(2)` is pertinent in that name. My real adapter is `Intel(R) Ethernet Controller (2) I225-V`, the one read from the registry doesn't have that `(2)`, but the second item in my registry is the real one without the 2, and has the matching address.

gazpachoking:
So, I think the actual code is working properly, but the tests are just fragile. I guess maybe the one read from the registry is an interface that was deleted or something.

gazpachoking:
Speaking of that line you linked, `for i in range(1):` is just the same as `for i in [0]:`, and isn't much of a loop at all.

doadin:
> Speaking of that line you linked, `for i in range(1):` is just the same as `for i in [0]:`, and isn't much of a loop at all.

yea it was originally as @DjLegolas had for i in range(5) but would error if there where less then 5.

doadin:
@gazpachoking yea, maybe one of the adapaters didn't clean uninstall after a driver update or something so a registry entry is still lingering.

gazpachoking:
It'd be nice to have the tests be more stable, but I'm just going to delete that key from my registry for now.

cas--:
@gazpachoking I agree, I only quickly checked the tests before merging. I am not sure how much we need to test here tbh as it's somewhat outside Deluge domain to verify adapter lookup and perhaps ifaddr should be mocked out.

DjLegolas:
That code was only a PoC, and needs improvement... 😅

PR #339, opened 2022-01-15
Mutli-Homing Continued

doadin:
https://github.com/deluge-torrent/deluge/pull/300


doadin:
@cas-- Continued from the PR linked in description with pretty much just the ipv6 check change and no geoip commit.

PR #340, opened 2022-01-16
Support pygeoip for ip lookups

mhertz:
For platforms where regular GeoIP unavailable

Fixes: https://dev.deluge-torrent.org/ticket/3271

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -23,10 +23,14 @@ import deluge.configmanager
 from deluge._libtorrent import lt
 from deluge.event import ConfigValueChangedEvent
 
+GeoIP = None
 try:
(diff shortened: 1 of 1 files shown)


mhertz:
This was meant for Windows more than anything, but wouldn't include the github-action for pip installing pygeoip, as probably best seperate I thought. I just remembered geoip support on windows lacking as no wheel on pypi for GeoIP, and I trieed build it from source, but error's out atleast for me on msvc2019, unless doing it wrong, but did define libs and headers and build the C api first etc. Anyway one of the devs of the pywrapper stated to use the pygeoip instead unless really needing the speed. All credit for code @cas, and I just changed it slightly to support both backends.

cas--:
Thanks closed via 65e5010e7

PR #341, opened 2022-01-21
[3498][Core] Fixed KeyError in sessionproxy after torrent delete

DjLegolas:
When several torrents are being removed from session, an exception was
being raised in a callback function `on_status` with the `torrent_id` of
one (or more) of the removed torrents.
Now we will catch when the torrent does not exist anymore and print a
debug log about it.

Closes: https://dev.deluge-torrent.org/ticket/3498

### DIFF
--- a/deluge/ui/sessionproxy.py
+++ b/deluge/ui/sessionproxy.py
@@ -145,11 +145,14 @@ class SessionProxy(component.Component):
 
                 def on_status(result, torrent_id):
                     t = time()
-                    self.torrents[torrent_id][0] = t
-                    self.torrents[torrent_id][1].update(result)
-                    for key in keys_to_get:
-                        self.cache_times[torrent_id][key] = t
-                    return self.create_status_dict([torrent_id], keys)[torrent_id]
+                    try:
+                        self.torrents[torrent_id][0] = t
+                        self.torrents[torrent_id][1].update(result)
+                        for key in keys_to_get:
+                            self.cache_times[torrent_id][key] = t
+                        return self.create_status_dict([torrent_id], keys)[torrent_id]
+                    except KeyError:
+                        return {}
 
                 return d.addCallback(on_status, torrent_id)
         else:


DjLegolas:
Exception example:
```
10:43:50.148 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] keypress_delete
10:43:52.288 [DEBUG   ][deluge.ui.gtk3.gtkui              :64  ] RPC: Sent 276.0 KiB (4.3 KiB/s) Recv 228.5 KiB (4.7 KiB/s)
10:43:52.656 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
10:43:53.016 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
10:43:53.019 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
10:43:53.019 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
10:43:53.021 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
10:43:53.021 [DEBUG   ][deluge.ui.gtk3.torrentview        :64  ] on_selection_changed
Unhandled error in Deferred:

  File "C:\develops\python\deluge\deluge\transfer.py", line 126, in _handle_complete_message
    self.message_received(
  File "C:\develops\python\deluge\deluge\ui\client.py", line 130, in message_received
    d.callback(request[2])
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 460, in callback
    self._startRunCallbacks(result)
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 568, in _startRunCallbacks
    self._runCallbacks()
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "C:\develops\python\deluge\deluge\ui\sessionproxy.py", line 148, in on_status
    self.torrents[torrent_id][0] = t
builtins.KeyError: 'TORRENT_ID'

Unhandled error in Deferred:

  File "C:\develops\python\deluge\deluge\transfer.py", line 126, in _handle_complete_message
    self.message_received(
  File "C:\develops\python\deluge\deluge\ui\client.py", line 130, in message_received
    d.callback(request[2])
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 460, in callback
    self._startRunCallbacks(result)
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 568, in _startRunCallbacks
    self._runCallbacks()
  File "C:\Python38\lib\site-packages\twisted\internet\defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "C:\develops\python\deluge\deluge\ui\sessionproxy.py", line 148, in on_status
    self.torrents[torrent_id][0] = t
builtins.KeyError: 'TORRENT_ID'
```
Notice: the TORRENT_ID in both exceptions is the same here, but not always.

deluge/ui/sessionproxy.py
```diff
@@ -145,7 +145,13 @@ def get_torrent_status(self, torrent_id, keys):
...
                     t = time()
-                    self.torrents[torrent_id][0] = t
+                    torrent = self.torrents.get(torrent_id)
+                    if not torrent:
+                        log.debug(
+                            f'Got status for unidentified torrent id {torrent_id}'
+                        )
+                        return {}
+                    torrent[0] = t
                     self.torrents[torrent_id][1].update(result)
```
aresch, line 155: I think you could run into the same problem here if the torrent is removed after you get it from the self.torrents dict. Maybe doing the same thing we do below in get_torrent_statuses() (wrap the body of this function in a try/except on KeyError) would be the better approach?

DjLegolas: It might be a better approach. Pushed a fix.

DjLegolas:
I think I should try and add a test for this case, but having trouble with the timings.
@cas-- @aresch @gazpachoking - can you help/suggest ideas?

aresch:
Looking at this a bit more, maybe it would make sense to change how this works a little bit. How about instead of immediately updating the state in on_torrent_added(), on_torrent_removed() and on_torrent_state_changed() we store these updates in a queue (with the current time for the cache) and have a function that runs at the beginning of get_torrent_status()/get_torrents_status() that processes the queue. This should prevent possible timing issues and we can remove the try/excepts for KeyError.

What do you think?

DjLegolas:
I don't think it will work.
First, `on_torrent_added()` is the only handler which requests the torrent state in its code. The other two, `on_torrent_removed()` and `on_torrent_state_changed()`, only updates the internal data, `self.torrents` and `self.cache_times`.
Which leads to the problem in hand, other parts of the UI tries to get a torrent state update, probably just before we get the `torrent removed` event:
```python
get_torrent_state()
-> torrent_id in self.torrents, calling client.core.get_torrent_status(), with on_status() callback
-> receiving torrent_removed_event so deleting torrent_id
-> processing the above on_status() callback
```

PR #337, opened 2022-01-13
Windows Continued

doadin:
- [ ]  email.mime deps
- [x]  Include all stdlib
- [x]  Include pygame for notification plugin
- [x]  Set GTK_CSD=0 in runtime hook
- [ ]  Host patched Twisted binary under Deluge

PR #342, opened 2022-01-21
Simplify PyInstaller spec file

CirnoT:
This makes the process of editing the file much more pleasant and removes duplicate code.

This PR also brings stdlibs as hidden imports, something which was missed by @cas-- in #337.

### DIFF
--- a/packaging/win/delugewin.spec
+++ b/packaging/win/delugewin.spec
@@ -1,322 +1,200 @@
-
 # -*- mode: python ; coding: utf-8 -*-
 import os
 import sys
 import deluge.common
-from PyInstaller.compat import is_win
-from PyInstaller.utils.hooks import collect_all, copy_metadata
+from PyInstaller.utils.hooks import collect_all, collect_submodules, copy_metadata
 
 datas = []
 binaries = []
 hiddenimports = ['pygame']
-runtime_hooks_gtk = [os.path.join(SPECPATH, 'pyi_rth_gtk_csd.py')]
 
 # Collect Meta Data
 datas += copy_metadata('deluge', recursive=True)
 datas += copy_metadata('service-identity', recursive=True)
 
 # Add Deluge Hidden Imports
-tmp_ret = collect_all('deluge')
-hiddenimports += tmp_ret[2]
+hiddenimports += collect_submodules('deluge')
 
-#Add Hidden Imports for Plugins
-tmp_ret2 = collect_all('twisted')
-hiddenimports += tmp_ret2[2]
+# Add stdlib as Hidden Imports.
+# This is filtered list that excludes some common examples or stuff not useful in plugins (such as tty, mailbox, turtledemo etc.).
+# It is safe to assume that 90% of that list would already be included anyway.
+stdlib = [
+    'string',
+    're',
+    'unicodedata',
+    'struct',
+    'codecs',
+    'datetime',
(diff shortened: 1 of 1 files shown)


CirnoT:
@cas-- I strongly recommend either copying the file directly or merging this via git (either as merge or squash) - it seems that you prefer to take edits from PR and apply them yourself however you've missed quite a few of smaller changes that I've requested and @doadin made (such as unused `build_version` or inlining of useless `tmp_ret`)

packaging/win/delugewin.spec
```diff
@@ -3,320 +3,128 @@
...
+}
+
+analysis = {}
+exe = {}
+coll = []
+
+# Perform analysis
+for e, d in executables.items():
+  runtime_hooks = []
+  if !d.console:
```
doadin, line 84: I'm guessing this excludes exes with console? However we need to exclude from more than just that also the web exes as well. And does console here refer to the console window or the deluge console? Cause if its window console and not deluge console we need to exclude deluge console as well.

doadin: Sorry if that was not clear what i mean is the deluge-web-script has console false so weould fall under this check but it is not a gtk exe and therefore doesn't need the hook, is it a big deal probabbly not but the way it is now we can control that so lets not go backwards? :)

CirnoT: Please check with code:

```py
...
EXE(
  ...
  console=d['console'],
  ...)
...
```

You are right however, this can't be used this way.

doadin:
Not home to test and I can't add labels for the cd to run. I can test later but looks ok to me.

cas--:
Can we avoid the use of single letters for variables, better to use something descriptive to help readability. 

Also would be nice to lint/format the file with black since it's python code

```
black packaging/win/delugewin.spec
```

CirnoT:
>Can we avoid the use of single letters for variables, better to use something descriptive to help readability.

Any idea what we can use instead? `e` refers to entry in `executables` while `d` refers to dictionary of that entry. We could also change it to use only `i` so the result would be `executables[i]['gtk']`

DjLegolas:
> Any idea what we can use instead? `e` refers to entry in `executables` while `d` refers to dictionary of that entry. We could also change it to use only `i` so the result would be `executables[i]['gtk']`

Hmm, you can call them `executable` and `data_dict`.
In addition, maybe do everything in the `Perform analysis`, `Executable` and `Collect` sections in one loop?
Just remove the last 2 for loops, and fix the indentation of the comments.
```python
for executable, data_dict in executables.items():
    # Perform analysis
    ...

    # Executable
    ...

    # Collect
    ...
```

doadin:
> 

I could be wrong in all of this however, some thoughts. I don't think you are supposed to run collect before you are done with analysis/executable. Maybe it will not matter in our case. Also maybe instead of data_dict since we have variables data and execute maybe executables_dict would be more on the nose. Also maybe script_name instead of executable since its not actually referring to a executable and rather the name of the script we are building for.

DjLegolas:
@doadin - I have no knowledge about `pyinstaller`, so you might be correct.
I only commented about the code, which can be totally wrong :sweat_smile:

doadin:
@DjLegolas yea just small differences. Other than the things cas forgot to merge I feel like most of this is just cleanup that could be put off not really import things happening.

PR #343, opened 2022-01-21
[Core] Stopped using deprecated functions

DjLegolas:
As we intend to support LT 2.0, we should stop using the deprecated methods and functions.
We should not remove the usage of them, as we still need to support older versions of LT, but when
using the latest version of LT, we should use the latest API.

Closes: https://dev.deluge-torrent.org/ticket/3499


cas--:
Can we try to have separate PRs so the easy changes can be merged, also provide some context for the change in the commit message. e.g. explain why are were replacing linux_distribution and perhaps a link to documentation

Also for deluge 2.1.x we can bump the lt min version to 1.2. This means just replacing the deprecated code rather than trying to support lt 1.1

DjLegolas:
Good points.
New PR for the `linux_distribution` drop: https://github.com/deluge-torrent/deluge/pull/345.
I have added a commit for min version and removed version validation check in `torrent.py`.

deluge/_libtorrent.py
```diff
@@ -26,7 +26,7 @@
         raise LibtorrentImportError('No libtorrent library found: %s' % (ex))
 
 
-REQUIRED_VERSION = '1.1.2.0'
+REQUIRED_VERSION = '1.2.0.0'
```
cas--, line 29: If we bump this version we can also remove any conditionals related to 1.2 elsewhere in code

an0n666: @cas-- I suggest that if you bump the requirement to 1.2.0. then you should go for at least 1.2.14 as there's been many bug and crash fixes in the later builds.

DjLegolas:
`resume_data` in `alertmanager` is being accessed when coping the `save_resume_data_alert` alert. It will keep warn us until we stop using a python binding with deprecation support.
Nevertheless, I have replaced the usage of the deprecated `alert.resume_data` in the alert handler in `torrentmanager.py`.

DjLegolas:
Ok, managed to fix the `url` and `name()` issues as well.
2 points I think I should bring up:
* `resume_data` can be fixed even further by skipping it when coping the `alert`, just like all of the `__x__` methods.
* when fixing the `url`, I learned that there is a function to parse magnet uris - [`lt.parse_magnet_uri()`](https://libtorrent.org/upgrade_to_2.0-ref.html#magnet-links). The return value is an `add_torrent_params` object which can be manipulated just before passing it to the `lt.add_torrent()` method.
 I think we should consider using it at also in https://github.com/deluge-torrent/deluge/blob/e75ef7e31f3a8befd6365acbf8b7ebbb774bb5e5/deluge/common.py#L755

Lastly, I need your input, @cas--, about upgrading to lt1.2.14 as the minimum version as @an0n666 suggested.

cas--:
>     * `resume_data` can be fixed even further by skipping it when coping the `alert`, just like all of the `__x__` methods.

I'm not sure what you mean here, can you elaborate? 

>       I think we should consider using it at also in common.py#L755

We should not use any libtorrent method outside of core since we should not require it to be installed if using a thin client.

> about upgrading to lt1.2.14 as the minimum version 

As I said we should not put a high hard limit on the lt version so leave it as 1.2.0 unless there is a critical issue we know Deluge is not compatible with. It would be up to package managers to ensure using latest stable versions of libtorrent

cas--:
Could you also add to the commit message a few details about the changes

DjLegolas:
> I'm not sure what you mean here, can you elaborate?

Sure. In [`alertmanager.py`](https://github.com/deluge-torrent/deluge/blob/e75ef7e31f3a8befd6365acbf8b7ebbb774bb5e5/deluge/core/alertmanager.py#L126), we are coping the content of each alert as it arrives as it will be deleted by the caller.
When we receive an alert of type `save_resume_data_alert`, one of it's members is the deprecated `resume_data`.
As part of the copy, we access `resume_data` as well, which triggers the `DeprecatedWarning`.
A possible solution, it treat it as `__x__` (just an example):
```python
alert_copy = SimpleNamespace(
    **{
        attr: getattr(alert, attr)
        for attr in dir(alert)
        if not attr.startswith('__') or attr == 'resume_data`
    }
)
```
The "correct" way, according to libtorrent [1.2 upgrade guide](https://libtorrent.org/upgrade_to_1.2-ref.html#resume-data-handling) (last line), is stop compiling LT with deprecated functions enabled, but it's not part of this PR.
 
> We should not use any libtorrent method outside of core since we should not require it to be installed if using a thin client.

I agree. 
Also, passing it from the core to the thin client is not that important and efficient.
Dropping the suggestion.

> As I said we should not put a high hard limit on the lt version so leave it as 1.2.0 unless there is a critical issue we know Deluge is not compatible with. It would be up to package managers to ensure using latest stable versions of libtorrent

Keeping `1.2.0` it is.

> Could you also add to the commit message a few details about the changes
Thinking of a splitting it into 2 commits exactly for that. Give me a couple of minutes.

DjLegolas:
Everything is fixed.

doadin:
@DjLegolas 
I am also seeing.
`
13:59:22.436 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key num_peers, please use peer.num_peers_connected
13:59:22.436 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key dht_nodes, please use dht.dht_nodes
13:59:22.436 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key has_incoming_connections, please use net.has_incoming_connections
13:59:22.437 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key num_peers, please use peer.num_peers_connected
13:59:22.437 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key dht_nodes, please use dht.dht_nodes
13:59:22.437 [DEBUG   ][deluge.core.core                  :64  ] Deprecated session status key has_incoming_connections, please use net.has_incoming_connections
`

DjLegolas:
@doadin with which version of lt?

doadin:
@DjLegolas sorry, with 2.0.5. So idk if you want to cover that here or not.

doadin:
@DjLegolas googleing part of it like please use net.has_incoming_connections you can see people getting that with lt 1.1.5 https://forum.deluge-torrent.org/viewtopic.php?t=55670 so Id imagine at least that one is old when for each version im not sure.

doadin:
seems they are all as old as 1.1.5

doadin:
I guess this list should be looked at https://github.com/deluge-torrent/deluge/blob/540d557cb2163c41af894ee252cc677d01887376/deluge/core/core.py#L69

DjLegolas:
It's not a list, but a dict for deprecated session status keys, but seem we only use it when the key is not  available at all:
```python
for key in keys:
    try:
        status[key] = self.session_status[key]
    except KeyError:
        if key in DEPR_SESSION_STATUS_KEYS:
            new_key = DEPR_SESSION_STATUS_KEYS[key]
            log.debug(
                'Deprecated session status key %s, please use %s', key, new_key
            )
            status[key] = self.session_status[new_key]
        else:
            log.debug('Session status key not valid: %s', key)
```
I'll try to fix it also, thanks.
And also for lt2.0

doadin:
yea, sorry dict, but yea I just mean look over it see if theres anything that is'nt needed etc if we are going to require lt 1.2+. Anyways you got this :) thanks!

DjLegolas:
The real issue is finding the caller that uses old keys...
It might be even a plugin (like ltConfig)

doadin:
num_peers  dht_nodes  has_incoming_connections show up with no plugins enabled.(maybe the other too just did a quick run to check) Is is still possible that its a plugin?

doadin:
my understanding is ltconfig only loads its settings if enable and doesn't modify settings else where. Could these maybe just be set from an old version of deluge?

DjLegolas:
@doadin So the deprecated error was being just printed by `Deluge` itself in the `core`...
The source was in all of the UIs' status bar update code.

Thanks for the catch.

doadin:
> @doadin So the deprecated error was being just printed by `Deluge` itself in the `core`... The source was in all of the UIs' status bar update code.
> 
> Thanks for the catch.

Cool, now we have modern UI code and up-to-date libtorrent api usage, nice work!

doadin:
@DjLegolas unfortunatly I am not sure this is all set. Currently I am seeing the status bar in gtkui say download 1 G/s but down speed in the torrent list treeview shows 3.6M/s.

DjLegolas:
So it seem that I was using the total uploaded/downloaded bytes in the session...
Should be fix.
@doadin - can you verify? Thanks

doadin:
@DjLegolas looks like its fixed, thanks!

doadin (approved):

doadin:
@DjLegolas sorry don't want to be a bother but I am also seeing 

`
/home/runner/work/deluge/deluge/deluge/core/torrent.py:239: DeprecationWarning: get_torrent_info() is deprecated
  self.torrent_info = self.handle.get_torrent_info()

https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/torrent_handle.hpp#L927

/home/runner/work/deluge/deluge/deluge/core/torrent.py:437: DeprecationWarning: auto_managed() is deprecated
  self.handle.auto_managed(auto_managed)

https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/torrent_handle.hpp#L905

/home/runner/work/deluge/deluge/deluge/core/torrent.py:427: DeprecationWarning: set_sequential_download() is deprecated
  self.handle.set_sequential_download(sequential)

https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/torrent_handle.hpp#L909

/home/runner/work/deluge/deluge/deluge/core/torrent.py:447: DeprecationWarning: super_seeding() is deprecated
  self.handle.super_seeding(super_seeding)

https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/torrent_handle.hpp#L957

`

doadin:
@DjLegolas sorry ignore my last comment.

DjLegolas:
Found some places that needed to be fixed also.
In addition, now I have removed all new known deprecated functions that were introduced in `LT 2.0`.

PR #344, opened 2022-01-22
[Core] Return plugin keys with get_torrents_status

gazpachoking:
When requesting all keys, get_torrents_status was missing plugin added keys
This commit brings the behavior in line with get_torrent_status, and deluge 1.3

Closes: [3357](https://dev.deluge-torrent.org/ticket/3357)

Disclaimer: It's been a while since I've had a deluge development environment up, and got frustrated before I could get it working on windows. I haven't actually tested this code, but it's a pretty simple change, so I'm hoping somebody could quickly verify it does what I think, and merge it in.
🥺
🙏
👖   

(Happy to see the project get more active, and pumped for some Windows builds again! Good work everybody. I tried to get my environment set up by following the github action steps, but 🤷‍♂️ )

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -773,7 +773,7 @@ class Core(component.Component):
         def add_plugin_fields(args):
             status_dict, plugin_keys = args
             # Ask the plugin manager to fill in the plugin keys
-            if len(plugin_keys) > 0:
+            if len(plugin_keys) > 0 or not keys:
                 for key in status_dict:
                     status_dict[key].update(
                         self.pluginmanager.get_status(key, plugin_keys)


gazpachoking:
Did this seem okay @cas-- ?

cas--:
Sorry I'm not sure what happened here, I hadn't looked properly at it yet I closed it! 🤔

It seems okay but I do remember something about plugin keys not being in status without requesting them

gazpachoking:
> It seems okay but I do remember something about plugin keys not being in status without requesting them

Some sort of performance concern? I'd be fine with some other way to explicitly request all the fields, add a keyword argument maybe?

cas--:
Oh did you push this branch to this repo? That will be deleted since it's a mirror of git.deluge repo

cas--:
Looking at the code it was probably overlooked instead. Thanks for making the suggestion but it likely needs tweaked to pass a kwarg to the callback:
```diff
-        def add_plugin_fields(args):
+        def add_plugin_fields(args, all_keys=False):
             status_dict, plugin_keys = args
             # Ask the plugin manager to fill in the plugin keys
-            if len(plugin_keys) > 0:
+            if len(plugin_keys) > 0 or all_keys:
                 for key in status_dict:
                     status_dict[key].update(
                         self.pluginmanager.get_status(key, plugin_keys)
                     )
             return status_dict
 
-        d.addCallback(add_plugin_fields)
+        d.addCallback(add_plugin_fields, all_keys=not keys)
         return d
```
Would also be nice to see a test but not an easy method to test

It's great to see you around again, what are the issues with setting up your env?

gazpachoking:
> Oh did you push this branch to this repo? That will be deleted since it's a mirror of git.deluge repo

Doh, looks like I did! Didn't intend that. I'll recreate

> It's great to see you around again, what are the issues with setting up your env?

I didn't have the tools to compile libtorrent set up, and didn't find anything precompiled for windows. I'll get it set up eventually, just got impatient at the time it was taking to test a one line change. 😛 

> Looking at the code it was probably overlooked instead. Thanks for making the suggestion but it likely needs tweaked to pass a kwarg to the callback.

That callback is nested inside get_torrents_status, so I'm not sure how adding that kwarg makes any difference, it should still have access to the args passed in to get_torrents_status. (my kwarg suggestion was to add a kwarg directly to get_torrents_status, but I'm not sure how nice that is, since it could end up conflicting depending on what pas passed in 'keys' argument.) If it isn't a performance concern, I think my initial method is still valid.

PR #346, opened 2022-01-22
[GTK] Hide account password length in log

DjLegolas:
We should not let anyone know the account's password length,
as it can help to crack it.
Instead, we will print a constant amount (10) of asterisks.


DjLegolas:
Windows CI pased on [my repo](https://github.com/DjLegolas/deluge/actions/runs/1733846732)... strange...

cas-- (approved):

PR #347, opened 2022-01-23
[Core] Return plugin keys with get_torrents_status

gazpachoking:
When requesting all keys, get_torrents_status was missing plugin added keys
This commit brings the behavior in line with get_torrent_status, and deluge 1.3

Closes: [3357](https://dev.deluge-torrent.org/ticket/3357)

Re-created from #344 because I accidentally pushed to the wrong remote.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -763,25 +763,21 @@ class Core(component.Component):
         )
 
     @export
+    @defer.inlineCallbacks
     def get_torrents_status(self, filter_dict, keys, diff=False):
         """
         returns all torrents , optionally filtered by filter_dict.
         """
+        all_keys = not keys
         torrent_ids = self.filtermanager.filter_torrent_ids(filter_dict)
-        d = self.torrentmanager.torrents_status_update(torrent_ids, keys, diff=diff)
-
-        def add_plugin_fields(args):
-            status_dict, plugin_keys = args
-            # Ask the plugin manager to fill in the plugin keys
-            if len(plugin_keys) > 0:
-                for key in status_dict:
-                    status_dict[key].update(
-                        self.pluginmanager.get_status(key, plugin_keys)
-                    )
-            return status_dict
-
-        d.addCallback(add_plugin_fields)
-        return d
+        status_dict, plugin_keys = yield self.torrentmanager.torrents_status_update(
+            torrent_ids, keys, diff=diff
+        )
+        # Ask the plugin manager to fill in the plugin keys
+        if len(plugin_keys) > 0 or all_keys:
+            for key in status_dict:
+                status_dict[key].update(self.pluginmanager.get_status(key, plugin_keys))
+        return status_dict
 
     @export
     def get_filter_tree(self, show_zero_hits=True, hide_cat=None):


gazpachoking:
> Oh did you push this branch to this repo? That will be deleted since it's a mirror of git.deluge repo

Doh, looks like I did! Didn't intend that. I'll recreate

> It's great to see you around again, what are the issues with setting up your env?

I didn't have the tools to compile libtorrent set up, and didn't find anything precompiled for windows. I'll get it set up eventually, just got impatient at the time it was taking to test a one line change. 😛 

> Looking at the code it was probably overlooked instead. Thanks for making the suggestion but it likely needs tweaked to pass a kwarg to the callback.

That callback is nested inside get_torrents_status, so I'm not sure how adding that kwarg makes any difference, it should still have access to the args passed in to get_torrents_status. (my kwarg suggestion was to add a kwarg directly to get_torrents_status, but I'm not sure how nice that is, since it could end up conflicting depending on what pas passed in 'keys' argument.) If it isn't a performance concern, I think my initial method is still valid.

gazpachoking:
> > It's great to see you around again, what are the issues with setting up your env?
> 
> I didn't have the tools to compile libtorrent set up, and didn't find anything precompiled for windows. I'll get it set up eventually, just got impatient at the time it was taking to test a one line change. 😛

Annnd I figured it out. The answer was in the github action, I forgot to handle libcrypto and libssl dlls. I'll do some testing later tonight.

DjLegolas:
There is an official release at [pypi](https://pypi.org/project/libtorrent/).
So just use `pip`.

gazpachoking:
> There is an official release at [pypi](https://pypi.org/project/libtorrent/).
> So just use `pip`.

Yep, that's what I ended up doing, but needed the libcrypto and libssl dlls manually copied in.

gazpachoking:
Okay, actually tested this now, and it works as expected.

cas--:
> That callback is nested inside get_torrents_status, so I'm not sure how adding that kwarg makes any difference, it should still have access to the args passed in to get_torrents_status.

I understand there isn't a scope issue here but it better to be explicit and using `all_keys` denotes that

cas--:
> > There is an official release at [pypi](https://pypi.org/project/libtorrent/).
> > So just use `pip`.
> 
> Yep, that's what I ended up doing, but needed the libcrypto and libssl dlls manually copied in.

Hmm that it interesting, I assume just having openssl installed wasn't enough. What steps did you take to resolve it?

mhertz:
Cas, sorry for butting in, but with py3.8 and up, "everything" is about os.add_dll_directory(), as PATH not used for dll searching anymore, atleast through normal calls. The py-GTK wrappers adds the GTK-containing dirs to that call, from searching your PATH, without that it wouldn't even load the GTK-UI(and that is added in through gvsbuild patches, so not available in other methods obtained, manual builds and whatnot), but we still have the issue of the "core-components"(non-GTK) like libcrypto and intl.dll etc(later tried loaded, before dir added from the py-wrappers(in GTK-dir), hence error's in log).

Either need copy it into ../site-packages dir, so next to the libtorrent pyd looking for it, or e.g. add e.g this as test into site-packages\deluge\__init__.py(and I use direct path here, just for example):

`import os; os.add_dll_directory("C:/OpenSSL-Win64")`

cas--:
Thanks @mhertz for details. So what I can see it that for development as long as OpenSSL libs are in the system it should work (different when bundling the app): 

A quick search to gather details about DLLs changes in 3.8+:  

https://bugs.python.org/issue43173
https://docs.python.org/3/library/os.html#os.add_dll_directory
https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

The concern here is that the libtorrent wheel has a hard-requirement on openssl that is tricky to install correctly on Windows. Should it be bundled by libtorrent then? I would consider raising an upstream issue for this :thinking:

gazpachoking:
This certainly isn't the right place to discuss, but, is IRC active anymore? Do we have a discord or something? (And do we perhaps [want one](https://discord.gg/BC9NspHU)?) If there is still activity in IRC we can bridge it to discord.

cas-- (approved):
Code looks good, if you could squash the commits and put the full url to ticket in the message that would be useful. I can merge it later in week, do you have access to deluge git repo?

> is IRC active anymore? Do we have a discord or something? (And do we perhaps want one?) If there is still activity in IRC we can bridge it to discord.

I setup a libera.chat IRC channel but I don't have the bandwidth to be on IRC. I know that discord is popular, if you think it might help then can try it out.

gazpachoking:
> Code looks good, if you could squash the commits and put the full url to ticket in the message that would be useful. I can merge it later in week, do you have access to deluge git repo?

I'll squash it and fix up the commit message. Not sure if I still have access to git, been a long time.

> I setup a libera.chat IRC channel but I don't have the bandwidth to be on IRC. I know that discord is popular, if you think it might help then can try it out.

I made it, we can try it out. I don't know how much action there is in IRC, I can enable a bridge for now while we are testing. I did the same for FlexGet, and it seems like over 99% of our action moved over to discord now.

gazpachoking:
Okay, squashed and updated message. Looks like I don't have access to the git repo anymore. Or I lost my ssh key.

PR #348, opened 2022-01-25
Enable Windows tests

gazpachoking:
Working on fixing up some tests that are being unnecessarily skipped on Windows
- Uses the RPC to shutdown the daemon process started for tests. On windows sending the SIGINT caused the process to exit with a 1 exit code, which would fail the tests.
- Fixed up some handling of paths, either escaping backslashes from windows paths, or just using forward slashes on windows.
- The torrentview tests were never updated after a new default column was added, fixed those tests for both platforms. (though they aren't run by github actions)
- Apparently the sort order in listviews is different on linux vs windows, (whether punctuation comes before or after alphanum,) renamed the example filenames so that this doesn't come in to play.
- Some tests involving temp files weren't closing the file descriptor before deleting the file, which causes an exception on Windows. Just cleaned these up to use more modern tempfile context manager handlers where we don't have to worry about that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 .gitignore export-ignore
 *.py diff=python
 ext-all.js diff=minjs
+*.state -merge -text
--- a/deluge/i18n/util.py
+++ b/deluge/i18n/util.py
@@ -119,7 +119,7 @@ def setup_translation():
                 try:
                     libintl = ctypes.cdll.LoadLibrary(library)
                 except OSError as ex:
-                    errors.append(ex)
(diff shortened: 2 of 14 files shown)


gazpachoking:
Okay, I have most of  the tests running on Windows now. Didn't even look at the security tests, as they are using bash scripts. Having a weird issue with gtk tests, where it says the mutex is already open when starting the gtkui, just left that disabled for now.

DjLegolas:
I'm pretty sure that the GTK tests are not being tested in the CI...

PR #349, opened 2022-01-25
Multli-Home Continued

doadin:


doadin:
note:

    for interface in interface.split(','):
        if is_ip(interface) or is_adapter(interface):
            log.info('Found Interface or IP.', interface)
        else:
            log.error('Unable to verify Interface or IP.', interface)
            return False
    return True

PR #351, opened 2022-01-27
Simplify methods using inlineCallbacks

gazpachoking:
Just trying to clean up some of the code and make it more readable and easier to reason about. Opening this PR up for comments while in draft form. Some of these bits I think are easy wins, and some are a bit more complicated to figure if it's worth it. Maybe I should split off some into separate PRs so it's easier to merge simple ones without letting more complicated bits tie things up. I've been keeping the commits self contained, so it's probably easier to evaluate per commit rather than the whole diff.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -499,6 +499,7 @@ class Core(component.Component):
         return task.deferLater(reactor, 0, add_torrents)
 
     @export
+    @defer.inlineCallbacks
     def add_torrent_url(self, url, options, headers=None):
         """
         Adds a torrent from a URL. Deluge will attempt to fetch the torrent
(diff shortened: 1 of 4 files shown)


aresch:
I haven't reviewed the changes closely, but I like this. I was thinking the same thing after reviewing some other PRs recently. I do like the idea of breaking this up into smaller PRs though because as you said, some of the changes could get a bit complicated to reason about.

gazpachoking:
I'm going to test using async/await, and see how well that mixes in without trying to convert the whole code base at once. Would be really nice if we could move toward that syntax a bit at a time.

aresch:
Yes, I was reading about that last night. It looks like it's possible and there are various blog posts about it. Twisted provides an asyncreactor to use. It seems like a reasonable first step would be to move to using inlineCallbacks which would make the move to async/await a lot easier.

gazpachoking:
I think we could move to the await syntax without actually moving to asyncio too. Using the asyncreactor would let us use asyncio alongside twisted, but I don't think that's necessary, and might complicate things at first. Going to do a bit of playing around.

gazpachoking:
Okay, just barely started testing async/await things in #352 and I think that could be nicer. I'll try to continue that branch down to some meaningful improvements and see how much nicer, or more of a hassle it seems.

gazpachoking:
I'll redo this on top of maybe_coroutine stuff, if/when that goes in.

PR #352, opened 2022-01-27
[Core] Test mixing in async/await syntax

gazpachoking:
This is maybe a bit more ambitious alternative to #351 

Instead of using inlineCallbacks we could switch over to the await syntax. In order to make it the most nice, we'd have to start at the top of the call chain and convert over to async/await as we go down, to not have to mix in a lot of ensureDeferred calls. We could also make a decorator for ensureDeferred that works more like inlineCallbacks, which would make the transition easier, but would remove some of the nicer benefits of await syntax like the improved tracebacks (I think?)

### DIFF
--- a/deluge/decorators.py
+++ b/deluge/decorators.py
@@ -11,7 +11,7 @@ import re
 import warnings
 from functools import wraps
 
-from twisted.internet import defer
+from twisted.internet import defer, reactor
 
 
 def proxy(proxy_func):
@@ -163,12 +163,34 @@ def deprecated(func):
     return depr_func
 
 
+class CoroDeferred(defer.Deferred):
+    """Deferred that will dynamically act like a coroutine instead of a Deferred if it's awaited."""
+
+    def __init__(self, coro):
+        super().__init__()
+        self.coro = coro
+        self.awaited = None
+        self.activate_deferred = reactor.callLater(0, self.activate)
+
+    def __await__(self):
+        if self.awaited in [None, True]:
+            self.awaited = True
+            return self.coro.__await__()
+        return self.__iter__()
+
+    def activate(self):
+        """If the result wasn't awaited before the next context switch, we turn it into a deferred."""
+        if self.awaited is None:
+            self.awaited = False
+            d = defer.Deferred.fromCoroutine(self.coro)
+            d.chainDeferred(self)
+
+
 def ensure_deferred(f):
     """Wraps an async function to make it usable as a normal function that returns a Deferred."""
 
     @wraps(f)
     def wrapper(*args, **kwargs):
-        result = f(*args, **kwargs)
-        return defer.ensureDeferred(result)
+        return CoroDeferred(f(*args, **kwargs))
 
     return wrapper
--- a/deluge/decorators.py
+++ b/deluge/decorators.py
@@ -11,6 +11,8 @@ import re
 import warnings
 from functools import wraps
 
+from twisted.internet import defer
+
 
 def proxy(proxy_func):
     """
@@ -159,3 +161,14 @@ def deprecated(func):
         return func(*args, **kwargs)
 
     return depr_func
+
+
+def ensure_deferred(f):
+    """Wraps an async function to make it usable as a normal function that returns a Deferred."""
+
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        result = f(*args, **kwargs)
+        return defer.ensureDeferred(result)
+
+    return wrapper
--- a/deluge/component.py
+++ b/deluge/component.py
@@ -14,6 +14,8 @@ from twisted.internet import reactor
 from twisted.internet.defer import DeferredList, fail, maybeDeferred, succeed
 from twisted.internet.task import LoopingCall, deferLater
 
+from deluge.decorators import ensure_deferred
+
 log = logging.getLogger(__name__)
 
 
@@ -305,6 +307,7 @@ class ComponentRegistry:
         else:
             return succeed(None)
 
+    @ensure_deferred
     async def start(self, names=None):
         """Start Components, and their dependencies, that are currently in a Stopped state.
 
--- a/deluge/component.py
+++ b/deluge/component.py
@@ -305,7 +305,7 @@ class ComponentRegistry:
         else:
             return succeed(None)
 
-    def start(self, names=None):
+    async def start(self, names=None):
         """Start Components, and their dependencies, that are currently in a Stopped state.
 
         Note:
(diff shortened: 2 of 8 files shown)


aresch:
This would probably break plugins?

aresch:
By the way, here is some history when I went down a somewhat similar path in the past: https://dev.deluge-torrent.org/ticket/3207

gazpachoking:
> This would probably break plugins?

Hmm, probably right, I'll look closer at that. They can access more stuff than just the `@export`ed methods, right?



> By the way, here is some history when I went down a somewhat similar path in the past: https://dev.deluge-torrent.org/ticket/3207

I'll take a peek at that.

I'm thinking given these issues, maybe the nicest compromise is to make a decorator that calls ensureDeferred, and wrap any converted functions in it, just like we would with inlineCallbacks. It wouldn't give us all the advantages of async/await, but it would ensure full compatibility with everything existing, while letting us use that syntax within the wrapped functions. I don't think that method has any disadvantages compared with inlineCallbacks either.

gazpachoking:
Okay, I added the ensure_deferred decorator, which will let us keep using the wrapped async functions as normal functions returning deferred. I'm investigating whether there's something clever to do that will dynamically remove the wrapping if called directly from another async function, but it might be a bit too magicy, (or not work at all,) we'll see.

gazpachoking:
I may have just created a best of both worlds situation. I modified the `ensure_deferred` decorator to only call `ensureDeferred` in situations where the function was not being used like a coroutine function. This means in works exactly like an `inlineCallbacks` function, _unless_ it is called from within another async function, in which case it acts like a plain coroutine. This means in areas where we've converted the whole chain to async functions we'll get nice tracebacks, without breaking compatibility with places still using legacy twisted deferred syntax.

I'm really quite excited with this solution, but it probably needs a bit more testing before I can tell if there are any weird idiosyncrasies that make it sub-optimal.

@aresch @cas-- How do you feel?

example:
```python
@ensure_deferred
async def test():
    print('something')

@ensure_deferred
async def outer():
    await test()  # This works
    test()  # This also works  (☉_☉)

d = test()  # yep, works here too
d.addCallback(someotherfunc)  # totally fine
```

gazpachoking:
Hmm. Might be some issues still. Gtkui doesn't seem to be working.

gazpachoking:
I 'm pretty excited about this method, I'm hoping there are no caveats I'm failing to see at the moment. A couple questions though:
- Location: is decorators the right spot to put it? I guess so..
- Naming: It isn't really ensuring the result is a deferred anymore, I'm not sure if there is a better name.

deluge/decorators.py
```diff
@@ -159,3 +161,36 @@ def depr_func(*args, **kwargs):
...
+
+    def activate(self):
+        """If the result wasn't awaited before the next context switch, we turn it into a deferred."""
+        if self.awaited is None:
+            self.awaited = False
+            d = defer.Deferred.fromCoroutine(self.coro)
+            d.chainDeferred(self)
+
+
+def ensure_deferred(f):
```
cas--, line 189: Twisted already has [maybeDeferred](https://twistedmatrix.com/documents/current/api/twisted.internet.defer.html#maybeDeferred) so it might be confusing to name the decorator `maybe_deferred` so what if we went the other way with `maybe_coroutine`?

aresch:
This PR kind of scares me. I feel like it's warping the definitions of both coroutines and deferreds into a bit of a mess that might not be easy to reason about. Being able to call a "coroutine" without await seems weird to me. There is also no async loop created, so how would a regular coroutine even work? Are you planning to install the asyncreactor?

gazpachoking:
> This PR kind of scares me. I feel like it's warping the definitions of both coroutines and deferreds into a bit of a mess that might not be easy to reason about. Being able to call a "coroutine" without await seems weird to me. There is also no async loop created, so how would a regular coroutine even work? Are you planning to install the asyncreactor?

async/await is just a language feature separate from asyncio. I'm not involving asyncio at all in this PR, this is all still pure twisted. Twisted allows using the async/await syntax, which under the hood works almost exactly the same as using inlineCallbacks on a generator function. https://twistedmatrix.com/documents/16.4.1/core/howto/defer-intro.html#coroutines-with-async-await

Once an async function is wrapped in ensureDeferred, it can feel free to call other coroutines that aren't wrapped in ensureDeferred, as the outer most wrapping will handle the whole thing. This is desirable, because wrapping every level of coroutines in ensureDeferred mangles tracebacks, (not any more than they are now, but that is one of the nice benefits of async/await vs inlineCallbacks.) For us though, since we have a lot of places calling in to our functions, including plugins we might not control, leaving unwrapped coroutines would break any place that called it without the await keyword. The magic I've added here allows coroutines to be left unwrapped by ensureDeferred _only_ in the case where it's being called from another function that already has that wrapping. When used outside of an `await` statement, the ensureDeferrred wrapper is automatically added, making the coroutine into the excact same thing as an `inlineCallbacks` wrapped generator.

I'm still not 100% sure there aren't side effects, I'm trying to trace down the CI failures now to see what's going on with them, but at the moment I suspect it's because I've subtly changed the logic of component.start, and nothing to do with the new syntax.

gazpachoking:
> Being able to call a "coroutine" without await seems weird to me.

In this case, it turns into the same thing as running a function that returns a deferred.
```python
@inlineCallbacks
def a():
    yield defer.succeed()
    print('doing something')

a()

@ensure_deferred
async def b():
    await defer.succeed()
    print('doing something')

b()  # acts exactly the same as a()

aresch:
Thanks for the clarification. I guess I don't understand why the example with b() is preferable to a()?

gazpachoking:
> Thanks for the clarification. I guess I don't understand why the example with b() is preferable to a()?

A couple things I was shooting for:
1. get closer to a syntax where we could one day switch from twisted to asyncio
2. await is a bit nicer than yield for a few things, for example you can't:
    ```return yield d``` or ```callafunc(yield d)```, but you can with await
2. Improve tracebacks. With inlineCallbacks, (or regular callbacks,) the traceback doesn't show you the line that called `yield a_deferred` as the caller. When chaining awaits, it does.

Of these, 1 and 2 can be achieved by just creating a decorator that calls twisted's ensureDeferred which we wrap any functions we convert to async in. 3 is only achievable when the chained awaits only have the outermost coroutine wrapped. For this, we either need to convert everything at once, including plugins, or use something like the CoroDeferred that I made.

gazpachoking:
Man, I'm really struggling to figure out what's up with the CI failures. It looks like something happens to the daemon where it doesn't write the localclient account to the auth file. It doesn't happen every time though, and I can't figure why. Also can't figure how my current changes could cause this.

aresch:
Perhaps you could write some tests for CoroDeferred and the wrapper?

gazpachoking:
> Perhaps you could write some tests for CoroDeferred and the wrapper?

https://github.com/gazpachoking/deluge/blob/maybe_coroutine/deluge/tests/test_maybe_coroutine.py

I should probably stop making new branches. 😝  Perhaps the usage in those tests makes it more clear what I'm going for, and how it works? I'm sure there is other stuff to be tested, but can't figure it right now.

aresch:
Yeah, I was just suggesting it as a way to hunt down any edge cases that may be causing the CI failures.

gazpachoking:
Man, the converting componentregistry.start is really stumping me. I tried just regular inlinecallbacks and it has the same issue of sometimes not working.

gazpachoking:
I finally figured it out. The components need to be loaded in a certain order which isn't encoded fully in their 'depends'. I don't know exactly which order breaks things, but when I forced them to go in the same order as before I made any changes to the start function (by making each one depend on the next) it works every time.

gazpachoking:
And now I think I actually understand _why_ the order is important. The tests wait for the message 'Finished loading' to consider the daemon started. This means only that the TorrentManager component has loaded. It doesn't guarantee that the AuthManager and RPCServer has started, so if the client tries to connect before they have started there are issues.

gazpachoking:
Closing this. I'll do similar refactoring, but on top of maybe_coroutine PR if/when that goes in.

PR #353, opened 2022-01-29
Add maybe_coroutine decorator to easily mix in async/await syntax

gazpachoking:
Continuation of #351 and #352. Started fresh again, this time with some tests.

Goals:
- Clean up callback hell by making more code inline
- Use async/await syntax as it has more modern niceties than inlineCallbacks
  - Also gets us closer if we want to transition to asyncio in the future
  - `await` is usable in places that `yield` is not. e.g. `return await thething` `func(await thething, 'otherparam')`
  - IDEs know async semantics of async/await syntax to help more than with `inlineCallbacks`
- `maybe_coroutine` decorator has nice property (over `ensureDeferred`) that when used in a chain of other coroutines, they won't be wrapped in deferreds on each level, and so traceback will show each `await` call leading to the error.
- All async functions wrapped in `maybe_coroutine` are 100% backwards compatible with a regular Deferred returning function. Whether called from a coroutine or not.

### DIFF
--- a/deluge/decorators.py
+++ b/deluge/decorators.py
@@ -10,6 +10,9 @@ import inspect
 import re
 import warnings
 from functools import wraps
+from typing import Any, Callable, Coroutine, TypeVar
+
+from twisted.internet import defer
 
 
 def proxy(proxy_func):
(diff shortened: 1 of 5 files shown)


gazpachoking:
Okay, once more with feeling. Starting out at the very basics, with tests, and justifications, in hopes of getting something cleaner going here.

gazpachoking:
Man, what is going on with the tests. I didn't touch any deluge code yet in this branch.

deluge/decorators.py
```diff
@@ -159,3 +161,48 @@ def depr_func(*args, **kwargs):
...
+        assert not self.awaited, 'Cannot add callbacks to an already awaited coroutine.'
+        self.activate()
+        return super().addCallbacks(*args, **kwargs)
+
+
+def maybe_coroutine(f):
+    """Wraps an async function to make it usable as a normal function that returns a Deferred."""
+
+    @wraps(f)
+    def wrapper(*args, **kwargs):
```
aresch, line 202: Shouldn't this be an async def? Presumably you're using this decorator on async functions, so it should return one?

gazpachoking: Nope, it turns it into a regular function. That's why it's usable in situations outside of async functions. Works almost the same as the two commented lines below using 'ensureDeferred'.

gazpachoking: Sorta like this wrapper here. https://stackoverflow.com/a/44627140 `CoroutineDeferred` has _almost_ the same function as `ensureDeferred`, except it will pass the coroutine through directly where appropriate.

gazpachoking:
> Man, what is going on with the tests. I didn't touch any deluge code yet in this branch. Maybe those need to be refactored before any other work.

I think maybe the problem is that my understanding of how the reactor works when used in the test suite? Something my added test file does messes up subsequent tests, but I'm not really sure what I need to do differently.

gazpachoking:
Rebased this on top of the new test rewrites and everything it working fine now. Was some test weirdness with the mixed twisted.trial and unittests that must have been messing it up.

gazpachoking:
I don't want to put too much in to this, because some of the refactoring I want to do with it should probably be in their own PRs, so it's easier to review stuff. This one I've just added the decorator, tests and converted over a few existing inlineCallbacks.

Since this is on top of my test changes branch, here's the real diff for this branch:
https://github.com/gazpachoking/deluge/compare/no_twisted_trial...gazpachoking:maybe_coroutine

cas--:
@gazpachoking Thanks a lot for this and the tests PR! I'll try to have a look by the weekend

deluge/decorators.py
```diff
@@ -159,3 +162,54 @@ def depr_func(*args, **kwargs):
...
+        self.activate()
+        return super().addCallbacks(*args, **kwargs)
+
+
+_RetT = TypeVar('_RetT')
+
+
+def maybe_coroutine(
+    f: Callable[..., Coroutine[Any, Any, _RetT]]
+) -> Callable[..., defer.Deferred[_RetT]]:
```
cas--, line 205: I was about to merge this but this typing (I assume Twisted) is tripping me up

gazpachoking: I switched it to a comment now, that should fix things.

gazpachoking: And now I switched it to a string, which is a bit nicer.

gazpachoking:
Oops, found an issue. Wrapping those logging functions was a bad idea, because they get called before gtkui installs the gtkreactor.

gazpachoking:
Does anyone know why these are wrapped in inlineCallbacks to start with? https://github.com/deluge-torrent/deluge/blob/1089adb84420909c1e716329a2c419a63ee29b6f/deluge/log.py#L54-L86

cas--:
>Does anyone know why these are wrapped in inlineCallbacks to start with? My inclination is to just make them synchronous. 

There are some odd decisions in old code :shrug: I agree though

gazpachoking:
I made them synchronous. Doesn't seem to have any ill effects, and fixed the issue at hand.

cas--:
Excellent work @gazpachoking this should lay a good foundation for the async code!

PR #354, opened 2022-01-29
Remove twisted.trial tests

gazpachoking:
Everything I do seems to lead to other stuff that seems like it should be done. Trying to convert all the twisted.trial tests to pytest_twisted. Maybe also move off of unittest.TestCase as well. Seems there were several tests which weren't actually testing what they should, and also some code that wasn't doing what the broken test said it should. Started fixing that up as well. Everything is still quite messy and needs further refinement.

Goals:
- Remove twisted.trial tests
- Move to pytest fixtures, rather than many classess and subclasses with setup and teardown functions
- Move away from self.assertX to assert style tests
- FIx broken tests

With those in mind, I'd like to try to just get things working in a state where we aren't using twisted.trial anymore. Once we have a better framework and base, more tests can be modernized in smaller chunks to achieve more of the goals.

EDIT: After working on this longer, I think it is a good step. I believe I have (almost?) everything working again, and just have to do a bit more verification and cleanup. Going forward I think these should be the goals when adding/modifying tests:
- Don't use `BaseTest` or `set_up` `tear_down` methods any more. Fixtures should be used either in the test module/class, or make/improve the ones available in conftest.py
- For sure don't use unittest or twisted.trial, they mess up the pytest stuff.
- Prefer `pytest_twisted.ensureDeferred` with an async function over `inlineCallbacks`.
  - I think the async function syntax is nicer, and it helps catch silly mistakes, e.g. `await None` is invalid, but `yield None` isn't, so if some function returns an unexpected thing we try to await on, it will be caught earlier. (I struggled debugging a test for quite a while, then caught it immediately when switching to the new syntax)
  - Once the `maybe_coroutine` PR goes in, using the async syntax can also improve tracebacks when debugging tests.
 
Things that should probably be cleaned up going forward:
- Remove `BaseTestCase`
- Remove the subclasses like DaemonBase in favor of new fixtures.
  - I think there are some other utility subclasses that could be removed too
- Perhaps use parameterization in the ui_entry tests, rather that the weird combination of subclasses and the set_var fixture I mixed in.
- Convert some of the callback stuff to `pytest_twisted.ensureDeferred` tests, just for nicer readability

security_test


gazpachoking:
So, I guess I'd better ask, because this turned in to a much bigger thing than I thought. Are we in to this? @cas-- @aresch ?

I think it's probably worth it in the long run if I can get things cleaned up a bit more. Right now it's a mix of manually redone tests that are nice, some things that I change only just enough to work, and some in between. I have found multiple cases of tests that were broken, (passing, but not testing the thing they thought they were,) and even some where the thing they thought they were testing was broken. I think getting closer to pure pytest (with pytest-twisted) is much nicer, and converting some of the convoluted test subclasses into composable fixtures would be pretty nice.

aresch:
I don't have much of an opinion on this besides that I like pytest and moving to using it makes sense to me.

gazpachoking:
Looks like many of the pytest features don't work well with unittest test classes either. Some of the broken tests were because inlineCallbacks methods eat Exceptions if used in a unittests.TestCase class. Working on ripping that out completely.

deluge/tests/test_security.py
```diff
@@ -23,17 +23,10 @@
 SECURITY_TESTS = bool(os.getenv('SECURITY_TESTS', False))
 
 
-@pytest.mark.skipif(windows_check(), reason='windows cannot run .sh files')
-@pytest.mark.skipif(not SECURITY_TESTS, reason='skipping security tests')
+# TODO: This whole module has not been tested since migrating tests fully to pytest
```
cas--, line 26: We don't normally run this test as it's quite a long test but this module can be tested if you add the text security_test to PR description: https://github.com/deluge-torrent/deluge/blob/73394f1fc5da75ccf0bd466385c2141ff93bd604/.github/workflows/ci.yml#L45

deluge/tests/test_httpdownloader.py
```diff
@@ -154,140 +156,136 @@ def setUp(self):  # NOQA
...
-        return d
-
-    def test_page_not_found(self):
-        d = download_file(self.get_url('page/not/found'), fname('none'))
-        d.addCallback(self.fail)
-        d.addErrback(self.assertIsInstance, Failure)
-        return d
-
-    def test_page_not_modified(self):
+    @pytest.mark.xfail(reason="Doesn't seem like httpdownloader ever implemented this.")
```
gazpachoking, line 265: Not sure if we care about this or not, I didn't see any code that would actually handle this case.

deluge/tests/test_web_api.py
```diff
@@ -9,35 +9,34 @@
...
 
 
-class WebAPITestCase(WebServerTestBase):
-    def test_connect_invalid_host(self):
-        d = self.deluge_web.web_api.connect('id')
-        d.addCallback(self.fail)
-        d.addErrback(self.assertIsInstance, Failure)
-        return d
+class TestWebAPI(WebServerTestBase):
+    @pytest.mark.xfail(reason='This just logs an error at the moment.')
```
gazpachoking, line 28: The test claims it should raise an exception, but it just logs the error right now. Don't know if we want to update the behavior.

deluge/ui/console/__init__.py
```diff
@@ -13,4 +13,4 @@
 
 def start():
 
-    Console().start()
+    return Console().start()
```
gazpachoking, line 16: Was there a reason this didn't get returned before? It was being patched in the tests to return this, but I couldn't figure why we didn't want it to return in the real code.

deluge/tests/twisted/plugins/delugereporter.py
```diff
@@ -5,43 +5,44 @@
...
+#
+#
+# class DelugeReporter(TreeReporter):
+#     def __init__(self, *args, **kwargs):
+#         os.environ['DELUGE_REPORTER'] = 'true'
+#         TreeReporter.__init__(self, *args, **kwargs)
+#
+#     def addExpectedFailure(self, *args):  # NOQA: N802
+#         # super(TreeReporter, self).addExpectedFailure(*args)
+#         self.endLine('[TODO]', self.TODO)
```
gazpachoking, line 48: Anyone remember what this is for? Can it just be ripped out?

deluge/conftest.py
```diff
@@ -0,0 +1,158 @@
+#
```
gazpachoking, line 1: Not wild about this file not being in the `tests` subfolder, but I didn't know how to make it apply to plugin tests as well without moving it up to the root.

gazpachoking (commented):
I know this is huge and hard  to review, but I'm pretty sure everything is running and passing again on this version (including some extra stuff that wasn't working before.) It'd be nice to remove all the BaseTestCase tests, but I got tired of the really tedious work for now, and it isn't terrible. I left some comments on a few bits that I remembered might need discussion.

deluge/tests/twisted/plugins/delugereporter.py
```diff
@@ -5,43 +5,44 @@
 # See LICENSE for more details.
 #
```
aresch, line 7: Just delete this file?

gazpachoking: That was my thought, was just seeing if anyone remembered what it was for first.

aresch:
I've skimmed through most of this (it's too large to review too deeply) and it looks fine to me. Using pytest and removing some dependency on Twisted seem like good things to me. I'll defer to @cas-- if he has any issues with this.

LGTM

deluge/tests/test_security.py
gazpachoking: Thanks. Yep, those are looking good too.

gazpachoking:
My commits were not in meaningful chunks, and mostly had bad messages, so I squashed this down.

cas--:
@gazpachoking You know I recently watched a [talk by Raymond Hettinger](https://youtu.be/voXVTjwnn-U?t=1861) where he described a new Python core developer arriving and refactoring the entire test suite, causing unintended headaches...

I'm not saying this is not a good idea, it's just a notable coincidence  :stuck_out_tongue_closed_eyes:

deluge/conftest.py
cas--: This should be excluded from the packaging if it's in this dir. 

Is this [pytest issue](https://github.com/pytest-dev/pytest/issues/5822#issuecomment-697331920) the problem? Can we try the suggested solution?

deluge/ui/console/__init__.py
cas--: I'm not sure that there was any need to return it so seems fine to include it

deluge/conftest.py
gazpachoking: > The simplest solution is to just move the conftest.py up one directory (repo1/conftest.py).

This suggested solution is the one I went with.

Just tried the second solution of putting a conftest in both of our plugins that have tests test folders, looks like that suggested method isn't supported anymore.

```
Defining 'pytest_plugins' in a non-top-level conftest is no longer supported:
It affects the entire test suite instead of just below the conftest as expected.
  C:\Users\chase\PycharmProjects\deluge\deluge\plugins\Stats\deluge_stats\tests\conftest.py
Please move it to a top level conftest file at the rootdir:
  C:\Users\chase\PycharmProjects\deluge
For more information, visit:
  https://docs.pytest.org/en/latest/deprecations.html#pytest-plugins-in-non-top-level-conftest-files
collected 0 items / 1 error
```

gazpachoking: To exclude this from the package... that'd be MANIFEST.in?

gazpachoking:
> @gazpachoking You know I recently watched a [talk by Raymond Hettinger](https://youtu.be/voXVTjwnn-U?t=1861) where he described a new Python core developer arriving and refactoring the entire test suite, causing unintended headaches...
> 
> I'm not saying this is not a good idea, it's just a notable coincidence 😝

I'm happy to step back on any of these things if I'm stepping on toes or it's likely to break stuff. I'm exactly that guy that's exited to contribute that Hettinger was talking about, and I do think I'm making improvements, but let me know when I'm trampling on your flower garden. 😁 I have noticed several bits where the test/docstring doesn't do what it says on the tin, but that are delicate because there might be places depending on them working the way they do, even if the current way is 'wrong.'

deluge/conftest.py
cas--: Actually ignore my muttering about excluding it, must have been thinking of something else

cas--:
@gazpachoking This is fantastic work, thanks. I included a modified version of your PR description in the commit message, it's really useful!

PR #355, opened 2022-01-30
[Config] Don't log users passwords in plain text

doadin:


cas-- (changes requested):
This seems to replace the hosts line entirely instead of just hiding the password

doadin:
@cas-- If it runs into an error replacing that line yes.

doadin:
@cas-- If it works then the output looks like
`
Setting key "hosts" to: ('da57dedsrijvgbh38475ty78303c58', '127.0.0.1', 58846, 'localclient', '**********') (of type: <class 'tuple'>)
`

doadin:
I am also seeing on line 434 we also print passwords. Just noticed so that should be changed too.

doadin:
@cas-- Would you like it to do something different in the case of fail to replace?

cas--:
Okay I see. So the problem is that the code is clear enough that we are working around each host being a tuple. However it also doesn't account for multiple hosts so would need to loop the hosts list.

doadin:
@cas-- 
@line 179 in standalone or thinclient I am only seeing one host. @line line 434 though yes thats a whole another beast its a dict with lists of host in it I think? That one idk

cas--:
> line 179 in standalone or thinclient I am only seeing one host

The config key `hosts` is a list of hosts, your current change will overwrite them with a single host

doadin:
@cas-- Um interesting, I am currently seeing:

`
07:08:49.315 [DEBUG   ][deluge.config                     :179 ] Setting key "hosts" to: ('7b7ecb1276sdgkjvhn342571ee189f6cb1e533', '127.0.0.1', 58846, 'localclient', '**********') (of type: <class 'tuple'>)
07:08:49.315 [DEBUG   ][deluge.config                     :434 ] Config C:\Users\joshb\AppData\Roaming\deluge\hostlist.conf version: 1.3 loaded: {'hosts': [['18ebad274d29540sgjkn234758td20e8ffd0', '127.0.0.1', 58846, 'localclient', 'SECRET'], ['bf2c65eebb269a7sdklgjbns3473478c7a5dae8144a5', '127.0.0.1', 58845, 'localclient', 'SECRET'], ['b1e348c24sdjkfvgszjnkb34785387bb65c0b', '10.0.0.4', 58846, 'localclient', 'SECRET']]}
`

With my change and without:

`
07:12:48.900 [DEBUG   ][deluge.config                     :174 ] Setting key "hosts" to: [('1c09de15asjfvkasnjdk2345782345784111e6ef693e', '127.0.0.1', 58846, 'localclient', 'SECRET')] (of type: <class 'list'>)
07:12:48.901 [DEBUG   ][deluge.config                     :429 ] Config C:\Users\joshb\AppData\Roaming\deluge\hostlist.conf version: 1.3 loaded: {'hosts': [['18ebad274d295sjksfndvnjkl347834587940e8ffd0', '127.0.0.1', 58846, 'localclient', 'SECRET'], ['bf2c65eebsdjksdffjkn4784334782c7a5dae8144a5', '127.0.0.1', 58845, 'localclient', 'SECRET'], ['b1e348c2sdjsdjksd3453478478bb65c0b', '10.0.0.4', 58846, 'localclient', 'SECRET']]}
`

So idk it appears to only log one host at that line, weather thats intended or not idk.

doadin:
if key == "hosts":
    print(self.__config[key])

[('c6a33585sdjhkvfdjk3478437886d4f5e1a8', '127.0.0.1', 58846, 'localclient', 'SECRET')]

its only one host as far as I can tell. Idk maybe I am just doing something wrong. Sorry.

cas--:
It's not about what you are seeing just now, it is about what the code is capable of doing. Currently Deluge starts by setting only one host however there is no restriction to set multiple hosts in the config for this key and therefore the code changes are not correct since it doesn't account for a list of hosts. Add to the fact that the log.debug output for the `hosts` key should be a list of tuples not a tuple. Therefore this would result in a misleading debug log for the `hosts` key.

Likely want to do something like this:

```
if key == 'hosts' and value:
   # Scrub host password
   value = [host[:-1] + ('**********',) for host in hosts]
```

cas--:
If this is needed in multiple places within config code then might want to consider creating a scrub password function for the hosts list.

doadin:
@cas-- Current version of this pr is the best I got.

DjLegolas:
I would suggest the following changes.
* Use a scrab method to do all the work.
* Can handle both an item and the entire config
* Changed the default host list structure (from a list of tuples, to list of lists)
```diff
diff --git a/deluge/config.py b/deluge/config.py
--- a/deluge/config.py	(revision 8b0c8392b63f4af976c3046093f3ead96eddddc8)
+++ b/deluge/config.py	(date 1644059027928)
@@ -125,18 +125,31 @@
         # is set.
         self._save_timer = None
 
-        if defaults:
-            for key, value in defaults.items():
-                self.set_item(key, value)
-
         # Load the config from file in the config_dir
         if config_dir:
             self.__config_file = os.path.join(config_dir, filename)
         else:
             self.__config_file = get_default_config_dir(filename)
 
+        if defaults:
+            for key, value in defaults.items():
+                self.set_item(key, value)
+
         self.load()
 
+    def _scrab_passwords(self, value):
+        if self.__config_file.endswith('hostlist.conf'):
+            try:
+                if isinstance(value, dict):
+                    no_pass_value = [host[:-1] + ['**********'] for host in value['hosts']]
+                else:
+                    no_pass_value = [host[:-1] + ['**********'] for host in value]
+            except IndexError:
+                return ''
+            else:
+                return no_pass_value
+        return value
+
     def __contains__(self, item):
         return item in self.__config
 
@@ -171,7 +184,7 @@
         """
         if key not in self.__config:
             self.__config[key] = value
-            log.debug('Setting key "%s" to: %s (of type: %s)', key, value, type(value))
+            log.debug('Setting key "%s" to: %s (of type: %s)', key, self._scrab_passwords(value), type(value))
             return
 
         if self.__config[key] == value:
@@ -193,7 +206,7 @@
         if isinstance(value, bytes):
             value = value.decode('utf8')
 
-        log.debug('Setting key "%s" to: %s (of type: %s)', key, value, type(value))
+        log.debug('Setting key "%s" to: %s (of type: %s)', key, self._scrab_passwords(value), type(value))
         self.__config[key] = value
 
         global callLater
@@ -431,7 +444,7 @@
             filename,
             self.__version['format'],
             self.__version['file'],
-            self.__config,
+            self._scrab_passwords(self.__config),
         )
 
     def save(self, filename=None):

diff --git a/deluge/ui/hostlist.py b/deluge/ui/hostlist.py
--- a/deluge/ui/hostlist.py	(revision 8b0c8392b63f4af976c3046093f3ead96eddddc8)
+++ b/deluge/ui/hostlist.py	(date 1644059050119)
@@ -29,7 +29,7 @@
     """Create a new hosts key for hostlist with a localhost entry"""
     host_id = uuid.uuid4().hex
     username, password = get_localhost_auth()
-    return {'hosts': [(host_id, DEFAULT_HOST, DEFAULT_PORT, username, password)]}
+    return {'hosts': [[host_id, DEFAULT_HOST, DEFAULT_PORT, username, password]]}
 
 
 def validate_host_info(hostname, port):
```

cas--:
@DjLegolas I was just working on this myself including adding tests, I'll incorporate a few of your suggestions, thanks!

PR #356, opened 2022-02-02
Add the torrent label to the console info command

kingamajick:
Outputs the torrent's label when the detailed (or verbose) info is requested from deluge-console. This makes is easily accessible in scripts (such as those used by the execute plugin)

Closes: https://dev.deluge-torrent.org/ticket/1556

### DIFF
--- a/deluge/ui/console/cmdline/commands/info.py
+++ b/deluge/ui/console/cmdline/commands/info.py
@@ -67,6 +67,7 @@ STATUS_KEYS = [
     'total_payload_download',
     'total_payload_upload',
     'time_added',
+    'label',
 ]
 
 # Add filter specific state to torrent states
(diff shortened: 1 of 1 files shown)


gazpachoking:
Hmm. The labels come from a plugin, it feels weird to reference them directly in the core of the client. OTOH, I don't think we have any plugin support for console client... I think if it's even to be considered this has to not fail when the label plugin is not enabled. Also, it looks like the download folder output is overwritten by the label in this current incarnation.

cas-- (changes requested):
Needs work as @gazpachoking said

kingamajick:
Thanks for the feedback, i've updated it such that it will work both in the case that the label plugin is either enabled or disabled. I've also corrected the mistake where it would overwrite the download folder :)

cas-- (approved):

PR #358, opened 2022-02-02
[#3502]Fix Deluge won't start when logging to windows protected folder

doadin:


deluge/log.py
```diff
@@ -155,7 +155,10 @@ def setup_logger(
             handler_cls = getattr(
                 logging.handlers, 'WatchedFileHandler', logging.FileHandler
             )
-        handler = handler_cls(filename, mode=filemode, encoding='utf-8')
+        try:
+            handler = handler_cls(filename, mode=filemode, encoding='utf-8')
+        except FileNotFoundError:
+            handler = logging.StreamHandler(stream=output_stream)
```
gazpachoking, line 161: Should we log an error here so the user knows what happened?

doadin:
@gazpachoking look good now?

deluge/log.py
```diff
@@ -155,7 +155,12 @@ def setup_logger(
...
+            log = logging.getLogger(__name__)
+            log.error('Unable to write to log file')
```
gazpachoking, line 163: Maybe put the filename in the log message? Otherwise, looks good to me!
```suggestion
            log.error(f'Unable to write to log file `{filename}`')
```

doadin:
@gazpachoking Changed, thanks!

PR #359, opened 2022-02-03
[Core] Document all exported core methods with type hints

gazpachoking:
Started adding some type hints. This one does all the exported methods from core.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -14,6 +14,7 @@ import shutil
 import tempfile
 import threading
 from base64 import b64decode, b64encode
+from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.request import URLError, urlopen
 
 from twisted.internet import defer, reactor, task
@@ -401,7 +402,9 @@ class Core(component.Component):
 
     # Exported Methods
     @export
-    def add_torrent_file_async(self, filename, filedump, options, save_state=True):
+    def add_torrent_file_async(
+        self, filename: str, filedump: str, options: dict, save_state: bool = True
+    ) -> defer.Deferred[Optional[str]]:
         """Adds a torrent file to the session asynchronously.
 
         Args:
@@ -433,7 +436,9 @@ class Core(component.Component):
             return d
 
     @export
-    def prefetch_magnet_metadata(self, magnet, timeout=30):
+    def prefetch_magnet_metadata(
+        self, magnet: str, timeout: int = 30
+    ) -> defer.Deferred[Tuple[str, bytes]]:
         """Download magnet metadata without adding to Deluge session.
 
         Used by UIs to get magnet files for selection before adding to session.
@@ -461,7 +466,9 @@ class Core(component.Component):
         return result_d
 
     @export
-    def add_torrent_file(self, filename, filedump, options):
+    def add_torrent_file(
+        self, filename: str, filedump: Union[str, bytes], options: dict
+    ) -> Optional[str]:
         """Adds a torrent file to the session.
 
         Args:
@@ -486,7 +493,9 @@ class Core(component.Component):
             raise
 
     @export
-    def add_torrent_files(self, torrent_files):
+    def add_torrent_files(
+        self, torrent_files: List[Tuple[str, Union[str, bytes], dict]]
+    ) -> defer.Deferred[List[AddTorrentError]]:
         """Adds multiple torrent files to the session asynchronously.
 
         Args:
@@ -515,7 +524,9 @@ class Core(component.Component):
         return task.deferLater(reactor, 0, add_torrents)
 
     @export
-    def add_torrent_url(self, url, options, headers=None):
+    def add_torrent_url(
+        self, url: str, options: dict, headers: dict = None
+    ) -> defer.Deferred[Optional[str]]:
         """
         Adds a torrent from a URL. Deluge will attempt to fetch the torrent
         from the URL prior to adding it to the session.
@@ -553,7 +564,7 @@ class Core(component.Component):
         return d
 
     @export
-    def add_torrent_magnet(self, uri, options):
+    def add_torrent_magnet(self, uri: str, options: dict) -> str:
         """
         Adds a torrent from a magnet link.
 
@@ -571,7 +582,7 @@ class Core(component.Component):
         return self.torrentmanager.add(magnet=uri, options=options)
 
     @export
-    def remove_torrent(self, torrent_id, remove_data):
+    def remove_torrent(self, torrent_id: str, remove_data: bool) -> bool:
         """Removes a single torrent from the session.
 
         Args:
@@ -589,7 +600,9 @@ class Core(component.Component):
         return self.torrentmanager.remove(torrent_id, remove_data)
 
     @export
-    def remove_torrents(self, torrent_ids, remove_data):
+    def remove_torrents(
+        self, torrent_ids: List[str], remove_data: bool
+    ) -> defer.Deferred[List[Tuple[str, str]]]:
         """Remove multiple torrents from the session.
 
         Args:
@@ -625,7 +638,7 @@ class Core(component.Component):
         return task.deferLater(reactor, 0, do_remove_torrents)
 
     @export
-    def get_session_status(self, keys):
+    def get_session_status(self, keys: List[str]) -> Dict[str, Union[int, float]]:
         """Gets the session status values for 'keys', these keys are taking
         from libtorrent's session status.
 
@@ -656,13 +669,13 @@ class Core(component.Component):
         return status
 
     @export
-    def force_reannounce(self, torrent_ids):
+    def force_reannounce(self, torrent_ids: List[str]) -> None:
         log.debug('Forcing reannouncment to: %s', torrent_ids)
         for torrent_id in torrent_ids:
             self.torrentmanager[torrent_id].force_reannounce()
 
     @export
-    def pause_torrent(self, torrent_id):
+    def pause_torrent(self, torrent_id: str) -> None:
         """Pauses a torrent"""
         log.debug('Pausing: %s', torrent_id)
         if not isinstance(torrent_id, str):
@@ -671,7 +684,7 @@ class Core(component.Component):
             self.torrentmanager[torrent_id].pause()
 
     @export
-    def pause_torrents(self, torrent_ids=None):
+    def pause_torrents(self, torrent_ids: List[str] = None) -> None:
         """Pauses a list of torrents"""
         if not torrent_ids:
             torrent_ids = self.torrentmanager.get_torrent_list()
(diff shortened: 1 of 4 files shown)


cas--:
Unless we are going to include a type checker, I am not sure the value here? Also the type is included in the docstring and there should be only one source of truth

gazpachoking:
Hmm, that's a good point and I hadn't fully thought it out. Here were my goals:
1. Have the IDE know what stuff was when I'm refactoring. (I have several branches where I'm trying different stuff, and the IDE rarely has good type info at the moment.)
2. Define the contract a bit better for exported methods, so it's more apparent if they get changed. (Once again, I was thinking more how pycharm automatically highlights wrong usage/returns more than running an explicit type checker.)
3. I don't think it would be terrible to enable mypy checking in the loosest mode, where it only cares about the things we have already documented, such that type hints could be added bits at a time.

That being said, point 1 is probably not so applicable to the exported methods, because they normally get used via client, or via component.get(), both of which the IDE loses track of the type hints.
And 2 is probably better served by the docstrings, so it gets exported to the RPC docs. I do like the expressiveness of the type hints compared to just a simple type in the docstring, (e.g. List[Tuple[int, str]]) but in cases where the requirement is more complex it needs to be spelled out in words for the generated documentation anyway. If we did want to move to this style type hints, we could make it the single source of truth, and just make the docstrings for people https://pypi.org/project/sphinx-autodoc-typehints/ (this would be my vote.)

I still think the type hints are quite useful in the IDE for autocompletion, and I'd be inclined to add them to any bits I touch when doing any fixing/refactoring, unless you are against them completely. For more internal methods the disadvantages above are moot.

cas--:
I had forgotten about IDE usefulness so happy to try to accommodate there, just didn't want to have the types to become outdated or spend a lot of effort for little gain. 

Personally I find type hints can reduce code readability and unless constantly encountering type issues not worth the added overhead, especially if never fully verified by mypy.

Since we already have a form of typing in the code, I am happy if we experiment with moving types from docstrings. As you suggest starting with core exported methods and use [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/).

cas--:
@gazpachoking You know I recently watched a [talk by Raymond Hettinger](https://youtu.be/voXVTjwnn-U?t=1861) where he described a new Python core developer arriving and refactoring the entire test suite, causing unintended headaches...

I'm not saying this is not a good idea, it's just a notable coincidence  :stuck_out_tongue_closed_eyes:

DjLegolas:
I think that we should also start converting the docstrings to googles's format, probably in another PR.

cas--:
Can use https://github.com/dadadel/pyment/ to convert them

gazpachoking:
I did some work in core.py standardizing the docstrings, removing the types, and enabling sphinx generating them from the function signatures. Seems to be working pretty well.
image
The docstring updates were sorta tedious. I tried pyment, couldn't get it to work very well. I'd be more inclined to fix stuff as it gets touched going forward rather than do a big dump.

aresch:
Can we add a mypy check as a github action? We really should have typing everywhere that's feasible.

gazpachoking:
> Can we add a mypy check as a github action? We really should have typing everywhere that's feasible.

Yeah, I was just experimenting with mypy. Trying to figure the easiest way to introduce it without a shitload of comments on stuff to ignore before we get it all in.

deluge/core/core.py
```diff
@@ -54,6 +58,8 @@
 )
 from deluge.httpdownloader import download_file
 
+reactor = typing.cast(ReactorBase, twisted.internet.reactor)
```
gazpachoking, line 61: This isn't so nice. We import reactor quite a bit. If we enable mypy I think we should figure out some better way than doing this in every file.

deluge/core/core.py
```diff
@@ -456,7 +462,7 @@ def on_metadata(result, result_d):
 
         d = self.torrentmanager.prefetch_metadata(magnet, timeout)
         # Use a separate callback chain to handle existing prefetching magnet.
-        result_d = defer.Deferred()
+        result_d: defer.Deferred = defer.Deferred()
```
gazpachoking, line 465: I can't really fathom why mypy doesn't infer the type here?

deluge/core/core.py
```diff
@@ -673,7 +679,7 @@ def pause_torrent(self, torrent_id: str) -> None:
     @export
     def pause_torrents(self, torrent_ids: List[str] = None) -> None:
         """Pauses a list of torrents"""
-        if not torrent_ids:
+        if torrent_ids is None:
```
gazpachoking, line 682: This change was made so that mypy type narrowing worked and knew that torrent_ids couldn't be None any longer. For some reason `not torrent_ids` doesn't have the same result as far as mypy is concerned.

deluge/core/core.py
```diff
@@ -724,7 +730,7 @@ def resume_torrent(self, torrent_id: str) -> None:
     @export
     def resume_torrents(self, torrent_ids: List[str] = None) -> None:
         """Resumes a list of torrents"""
-        if not torrent_ids:
+        if torrent_ids is None:
```
cas--, line 733: This changes the functionality, an empty list would return nothing

deluge/core/core.py
```diff
@@ -770,9 +776,9 @@ def get_torrent_status(
             all_keys=not keys,
         )
 
-    @export
+    @export  # type: ignore
     @defer.inlineCallbacks
-    def get_torrents_status(
+    def get_torrents_status(  # type: ignore
```
gazpachoking, line 781: Ignored a bunch of stuff on inlineCallbacks for now. The generator would have to be properly type hinted as a generator, which is weird. This is one thing that would be a _lot_ easier with the `maybe_coroutine` branch. Type hinting an async function is the same as doing a normal function, no need to do fancy gen typing stuff.

deluge/core/core.py
```diff
@@ -770,9 +776,9 @@ def get_torrent_status(
             all_keys=not keys,
         )
 
-    @export
+    @export  # type: ignore
```
cas--, line 779: Why are we ignoring type here and below?

gazpachoking (commented):
I haven't used mypy a lot, but I think it could be a good idea to enable it. My only hesitation would be if we start having to add a lot of extra cruft to make mypy understand stuff. (by cruft, I mean stuff that is useless to a human reader, not just making things slightly more verbose, which could be useful to humans) I added a few notes on stuff I changed to get mypy to pass on core.py. My biggest concern is how it complains about twisted reactor. If there was some way we could solve that for the whole project at once it would be great.

cas--:
Can we leave the mypy to another PR and get this one merged

gazpachoking:
> Can we leave the mypy to another PR and get this one merged

Sure can. I'll back that stuff out, was just doing some experiments.

deluge/core/core.py
gazpachoking: Oh yeah, good point.

deluge/core/core.py
gazpachoking: I think I figured this out. Made some stubs for twisted that I told mypy to load. Defined reactor in the twisted.internet.__init__.pyi stub. I _think_ that still lets mypy use the real definitions for all the other twisted modules.

gazpachoking:
Restored this to just the doc changes and type hints.

deluge/core/core.py
aresch: You have to make torrent_ids an Optional[List[str]] since the default is None.

https://docs.python.org/3/library/typing.html#typing.Optional

gazpachoking: Optional is implied when the default is none.

gazpachoking: Hmm, I guess current recommendations are to make the `Optional` explicit, but the default mypy behavior still allows inferring it. https://mypy.readthedocs.io/en/stable/config_file.html#confval-no_implicit_optional

aresch: Maybe it only makes it Optional if you provide no typing?

gazpachoking: It for sure was getting the Optional type. The issue I was actually trying to solve was: `Item "None" of "Optional[List[str]]" has no attribute "__iter__" (not iterable)`. When the if was just `if not torrent_ids` it didn't narrow the type down to List[str], but when the check changed to `if torrent_ids is not None` it did narrow the type down, and allow iterating on it without complaint.

deluge/core/core.py
gazpachoking: I did a test merging my mypy stuff with the maybe_coroutine branch, and type hinting is muuuch nicer (than inlineCallbacks.) No need to wrap stuff in `Deferred[returntypehere]` no need to document generators with weird generator type hints where you have to specify the yield, send, and return types, all just works.

deluge/core/core.py
aresch: Huh, I guess it's because of the for loop below the conditional. Mypy doesn't know that you're changing the None into a List[str] with the call to get the torrents list. I suppose there are a few ways to change the code to work without reassigning torrent_ids which would likely fix the issue. I also wonder if setting the return value of TorrentManager.get_torrent_list() to List[str] would help?

gazpachoking: I actually figured it out. It wasn't the `not torrent_ids` that was messing up, it was that I hadn't typed `self.torrentmanager`, so it didn't actually know the return type when torrent_ids got reassigned. Did some more testing in another branch and got it working fine.

cas--:
@gazpachoking There are breaking compatibility changes with both older Twisted and Python versions in the PR, I have reverted the code from develop

Perhaps we concentrate on fixing other areas of the code and Windows release as I don't see the expenditure just now worth it?

gazpachoking:
> There are breaking compatibility changes with both older Twisted and Python versions in the PR

Ahh. Didn't consider older versions of twisted not having type hinting capability. Shouldn't be any issues with older python though, I didn't use the stdlib types as generics, so that should be fine. The twisted issue would also be resolved if we drop python 3.6 support, was questioning that in my mypy PR. https://github.com/deluge-torrent/deluge/pull/362

> Perhaps we concentrate on fixing other areas of the code and Windows release as I don't see the expenditure just now worth it?

Yeah, that's fine. I just prefer adding the type hints as I go fixing other stuff, so it's a bummer.

cas--:
> Shouldn't be any issues with older python though, I didn't use the stdlib types as generics, so that should be fine.

I thought I saw usage but seems I was mistaken

> re python 3.6

Certainly can consider dropping it for 2.1.x since it would simplify our support :thinking: 

If you can see a quick way to get typing in without too much overhead certainly update this PR. I just have limited time to investigate any resulting errors

I should also see about running a CI that has our minimum requirements and current requirements

gazpachoking:
> If you can see a quick way to get typing in without too much overhead certainly update this PR. I just have limited time to investigate any resulting errors

We could do comment type hints, or, probably nicer, just type Deferreds as Deferreds without typing their return value. If we do maybe_coroutine branch that won't be bad, since we don't return a Deferred if the function is wrapped in that.

gazpachoking:
I think there shouldn't be problems without the use of generic Deferred return types...

> I should also see about running a CI that has our minimum requirements and current requirements

Yeah, that would be useful.

gazpachoking:
Totally forgot type hints could just be strings. Made that change, it's a much better option.

gazpachoking:
Rebased to resolve the conflicts, should be good again.

PR #360, opened 2022-02-03
[Common] Fix deluge starting on Windows

mhertz:


mhertz:
Previous commit 897955f, removing all py2 support, breaks deluge.exe and deluge-gtk.exe from starting on windows, likewise deluge-script.pyw and deluge-gtk-script.pyw running through pythonw.exe.

I narrowed it down to said commit, and further disected down to this specific part to be the culprit, here readded back. Not sure if propper fix honestly, or you guys prefer other solution, in which case please close and implement better yourself, but just thought to make you aware atleast.

Granted, freezing the install with pyinstaller fixes this issue I found, and also mentioned this some time ago in other PR as a side-point, though thought it was related to something else there. Likewise can get working unfreezed with redirecting the empty stdout, maybe also by enabling debug-logging, didn't check that, but anyway, I guess you also want to support unfreezed installs regardless, for people installing themselves from source on windows.

Thanks in advance.

gazpachoking:
Hmm, what exactly was the issue @mhertz ? I haven't had trouble starting the gtkui on the develop branch.

mhertz:
The issue being that pythonw.exe running *.pyw scripts e.g. through generated exe's fails starting. I tested this on py 3.9.x and 3.10.x - the debug-exe works, as not through pythonw.exe, and as said, workarounded by adding '> foo' after e.g. deluge.exe from cmd.

Freezing fixes it.

Are you stating to be able to doubleclick deluge.exe and have it start, unfreezed on develop? I have ran some 20 tests now, with several commits and cloned VMs and whatnot, so would surprice me indeed :)

Thanks for your reply.

gazpachoking:
Ahh, no I didn't quite understand. I wasn't using the exes. I'll try to investigate what's going on. It's curious that running the same script with regular python works fine...

gazpachoking:
Figured out a nicer fix and sent a PR.

mhertz:
I much appreciate it, thanks alot :)

PR #361, opened 2022-02-03
Remove legacy PY2 sys.argv unicode handling

gazpachoking:
When using pythonw on windows, sys.stdout and stdin are None. We had a legacy `unicode_argv` handler that was crashing in this instance. Just removed that, since sys.argv is always unicode on python 3 to fix the crash.

This is an alternate fix for #360

On a side note, the logging (to a file) also doesn't work when using pythonw. I haven't figured out the reason for that yet, but I'll open up a separate issue/PR

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -11,7 +11,6 @@ import base64
 import binascii
 import functools
 import glob
-import locale
 import logging
 import numbers
 import os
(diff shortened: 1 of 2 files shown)


cas--:
:+1: Can you put the further description about why into the git commit message as useful reference

PR #363, opened 2022-02-05
[Config] Add mask_funcs to help mask passwords in logs

cas--:
Added a new Config class parameter `mask_funcs` to enable config
instances hide sensitive information that would appear in config debug
logs.

### DIFF
--- a/deluge/config.py
+++ b/deluge/config.py
@@ -107,13 +107,23 @@ class Config:
         file_version (int): The file format for the default config values when creating
             a fresh config. This value should be increased whenever a new migration function is
             setup to convert old config files. (default: 1)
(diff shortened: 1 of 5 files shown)


cas--:
@doadin @DjLegolas Can you double-check this solution works for hiding the passwords from logs

DjLegolas:
@cas-- looks good, can be merged as far as I can see.

BTW, I niticed there is no a direct test for `hostlist.py` at all. I guess it is being tested indirectly, but don't we want some specific tests?

DjLegolas (approved):

cas--:
@DjLegolas Ideally yes but I spent far too long messing with config code today :roll_eyes: :grin:

doadin (approved):

PR #364, opened 2022-02-05
[CI] Added CI run on minimum and current python requirements

DjLegolas:
CI will run on both python 3.6 and 3.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
   test-linux:
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.7", "3.10"]
 
@@ -34,15 +35,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
             ${{ runner.os }}-
-
-      - name: Add libtorrent deb repository
-        uses: myci-actions/add-deb-repo@8
-        with:
-          repo: deb http://ppa.launchpad.net/libtorrent.org/1.2-daily/ubuntu focal main
-          repo-name: libtorrent
-          keys: 58E5430D9667FAEFFCA0B93F32309D6B9E009EDB
-          key-server: keyserver.ubuntu.com
-          install: python3-libtorrent-dbg
 
       - name: Sets env var for security
         if: (github.event_name == 'pull_request' && contains(github.event.pull_request.body, 'security_test')) || (github.event_name == 'push' && contains(github.event.head_commit.message, 'security_test'))
@@ -71,7 +63,6 @@ jobs:
       - name: Test with pytest
         run: |
           ulimit -c unlimited  # Enable core dumps to be captured
-          cp /usr/lib/python3/dist-packages/libtorrent*.so $GITHUB_WORKSPACE/deluge
           python -c 'from deluge._libtorrent import lt; print(lt.__version__)';
           catchsegv python -X dev -m pytest -v -m "not (todo or gtkui)" deluge
 
@@ -85,6 +76,7 @@ jobs:
   test-windows:
     runs-on: windows-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.7", "3.10"]
 
@@ -111,8 +103,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip wheel
-          python -m pip install libtorrent==1.2.*
+          pip install --upgrade pip wheel
           pip install -r requirements.txt -r requirements-tests.txt
           pip install -e .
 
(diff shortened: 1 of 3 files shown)


DjLegolas:
Seem as there is no way to test on python3.6 because there is no libtorrent binding to 3.6 for windows or linux...

doadin:
imo 3.10 has been working nicely, I think we could change this from 3.7-3.10.

DjLegolas:
I'm thinking about upgrading the minimum python version, again, to 3.7 and run the CI on all the supported versions (3.7-3.10)...
@cas-- @gazpachoking @aresch what do you guys think?

gazpachoking:
I'm 👍 on dropping 3.7 support. Not sure when we'd want to fit that in in terms of versioning and releases. In terms of this PR, I'd say testing more than one version is an improvement, regardless of whether it hits every currently supported version. I'm also 👍 on adding in other versions up to 3.10

doadin:
should also maybe add fail-fast: false

DjLegolas:
While testing, it helps having everything fail fast.
But will add that flag later, thanks @doadin

doadin:
3.10 needs pytest >= 6.2.5 I think.

doadin:
https://github.com/deluge-torrent/deluge/blob/develop/requirements-tests.txt#L1
pytest == 6.2.5 ; python_version >= '3.10'
pytest != 5.2.3, < 5.4
maybe

DjLegolas:
No, I'll just drop the limits from `pytest`...
It worked on my local Windows and Ubuntu, both with python 3.8.

doadin (approved):

DjLegolas:
To sums this up:
* Testing python versions 3.7-3.10
* Switched `libtorrent`'s installation source to `pypi`
* Dropped `pytest` version limits, current version is `7.0.0` for python 3.10 (thanks @doadin for the heads up)

doadin:
I am noticing a few things we use like twisted in a coming version is droping python3.6 and pyinstaller, Pillow already has, could probabbly drop 3.6 our selves as well. if it was only pyinstaller then id say no but deps are starting to drop as well. At least on the windows side requiring 3.7+ only excluded windows xp I think which nobody should still be using so not really a big deal imo.

.github/workflows/ci.yml
```diff
@@ -10,6 +10,10 @@ on:
 jobs:
   test-linux:
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
```
cas--, line 14: Let's keep fail-fast on

.github/workflows/ci.yml
```diff
@@ -10,6 +10,10 @@ on:
...
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
```
cas--, line 16: can we test just 3.7 and 3.10

.github/workflows/ci.yml
```diff
@@ -32,22 +36,14 @@ jobs:
...
-          install: python3-libtorrent-dbg
-
       - name: Sets env var for security
         if: (github.event_name == 'pull_request' && contains(github.event.pull_request.body, 'security_test')) || (github.event_name == 'push' && contains(github.event.head_commit.message, 'security_test'))
         run: echo "SECURITY_TESTS=True" >> $GITHUB_ENV
 
       - name: Install dependencies
         run: |
           pip install --upgrade pip wheel
+          pip install libtorrent==1.2.*
```
cas--, line 46: let's move this to the requirements.txt and remove the version requirement

.github/workflows/ci.yml
```diff
@@ -68,7 +64,7 @@ jobs:
       - name: Test with pytest
         run: |
           ulimit -c unlimited  # Enable core dumps to be captured
-          cp /usr/lib/python3/dist-packages/libtorrent*.so $GITHUB_WORKSPACE/deluge
+          cp $LD_LIBRARY_PATH/python3*/site-packages/libtorrent*.so $GITHUB_WORKSPACE/deluge
```
cas--, line 67: this lib copy should not be needed if using pip install

cas-- (commented):
A few suggestions to make to both linux and win runs

.github/workflows/ci.yml
DjLegolas: Just a reminder, removing the version requirement will install libtorrent 2

.github/workflows/ci.yml
DjLegolas: Done

.github/workflows/ci.yml
DjLegolas: Removed.

.github/workflows/ci.yml
cas--: What I meant was, can we remove this line

PR #366, opened 2022-02-06
[#3501] Fix gdk.pixbuf crash loading tracker icons

doadin:


cas-- (commented):
A few things, this only applies to Windows so must be a check for that and should not be converting to png in this function, if anything in TrackerIcons code.

DjLegolas:
I think this should be a last resort solution...
Try to use the image, abd if we fail, convert.
Unless we cannot catch the error/exception at all.

doadin:
@cas--  go for it.

cas--:
@DjLegolas Unfortunately this is hard crash in GdkPixbuf

PR #368, opened 2022-02-08
Refactor prefetch_metadata for (hopefully) more clarity

gazpachoking:
Just trying to clean up some of the more complicated callback logic. Hopefully this implementation is a bit more clear. Okay if not, only one real bug is fixed, and that was just with the test. Notable changes:
- The test was awaiting a DeferredList. By default that will eat exceptions and just add them to the result list (including test assertion exceptions.) Added `fireOnOneErrback=True` to make sure that wasn't happening.
- Moved the logic for multiple calls to await the same response into torrentmanager from core, so no matter where the prefetch is called from it will wait for the original call.
- Implemented the multiple calls with an explicit queue of waiting callbacks, rather than a callback callback chain.
- Moved to one inline async function rather than split into a main and callback after alert function.
- Added some more type hints to the stuff I changed.

Planning on doing a few more cleanups, just trying to keep it in easy chunks.

### DIFF
--- a/deluge/tests/test_torrentmanager.py
+++ b/deluge/tests/test_torrentmanager.py
@@ -12,7 +12,7 @@ from unittest import mock
 
 import pytest
 import pytest_twisted
-from twisted.internet import task
+from twisted.internet import reactor, task
 
 from deluge import component
 from deluge.bencode import bencode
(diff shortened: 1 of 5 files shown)


deluge/tests/test_torrentmanager.py
```diff
@@ -78,7 +78,7 @@ async def test_prefetch_metadata(self):
 
         magnet = 'magnet:?xt=urn:btih:ab570cdd5a17ea1b61e970bb72047de141bce173'
         d = self.tm.prefetch_metadata(magnet, 30)
-        self.tm.on_alert_metadata_received(mock_alert)
+        reactor.callLater(0, self.tm.on_alert_metadata_received, mock_alert)
```
gazpachoking, line 81: Here's something I hadn't considered with the maybe_coroutine stuff. I don't think it's necessarily bad, since it really only affects tests rather than real usage. Since the prefetech_metadata call is actually async now (deferred returning functions are a bit different, in that they run synchronously until a deferred gets returned,) it hasn't run at all at this point, and just firing the metadata received alert happened before the torrent even got added to the session. Scheduling it to run means it happens after the initial prefetch_metadata call is started, and the test works properly again.

cas--: Could we add  this detail to the commit message and perhaps even a short comment summary above this line

cas--:
Thanks @gazpachoking I modified the test to use `inlineCallback` with `gatherResults` for a cleaner refactor since the for loop affronted my sensibilities ;)

PR #369, opened 2022-02-08
Automatically refresh and expire the torrent status cache

gazpachoking:
Stop at ratio was not working when no clients were connected, because
it was using a cached version of the torrent status, and never calling
for a refresh. When a client connected, it called for the refresh and
started working properly.

This fix turns `Torrent.status` into a property, which checks when the last refresh was, and automatically refreshes before returning if it was not recent (currently 5 seconds.) When setting the property it automatically stores the update time for the cache. Any place that wants `torrent.handle.status()` should instead use `torrent.status`, or `torrent.update_status()`, (which also returns said updated status,) after this change.

Closes: https://dev.deluge-torrent.org/ticket/3497

Questions:
- Is this method reasonable?
  _Seems like it_
- How long should it be before the status cache is considered expired?
  _Chose 5 seconds for now_
- There are a few methods in `Torrent` which grab `self.handle.status()` directly. Should those be converted to use the cache? If not, they should be changed to update the cache, to avoid extra cache refreshes.
  _Converted any of these to use `torrent.update_status()` which refreshes the cache (and returns the new status)_

### DIFF
--- a/deluge/core/torrent.py
+++ b/deluge/core/torrent.py
@@ -16,6 +16,7 @@ Attributes:
 import logging
 import os
 import socket
+import time
 from urllib.parse import urlparse
 
 from twisted.internet.defer import Deferred, DeferredList
@@ -234,7 +235,8 @@ class Torrent:
         self.handle = handle
 
         self.magnet = magnet
-        self.status = self.handle.status()
+        self._status = None
+        self._status_last_update = 0
 
         self.torrent_info = self.handle.torrent_file()
         self.has_metadata = self.status.has_metadata
@@ -267,7 +269,6 @@ class Torrent:
         self.prev_status = {}
         self.waiting_on_folder_rename = []
 
-        self.update_status(self.handle.status())
         self._create_status_funcs()
         self.set_options(self.options)
         self.update_state()
@@ -641,7 +642,7 @@ class Torrent:
 
     def update_state(self):
         """Updates the state, based on libtorrent's torrent state"""
-        status = self.handle.status()
+        status = self.update_status()
         session_paused = component.get('Core').session.is_paused()
         old_state = self.state
         self.set_status_message()
@@ -709,7 +710,7 @@ class Torrent:
             restart_to_resume (bool, optional): Prevent resuming clearing the error, only restarting
                 session can resume.
         """
-        status = self.handle.status()
+        status = self.update_status()
         self._set_handle_flags(
             flag=lt.torrent_flags.auto_managed,
             set_flag=False,
@@ -1024,7 +1025,7 @@ class Torrent:
             dict: a dictionary of the status keys and their values
         """
         if update:
-            self.update_status(self.handle.status())
+            self.update_status()
 
         if all_keys:
             keys = list(self.status_funcs)
@@ -1054,13 +1055,35 @@ class Torrent:
 
         return status_dict
 
-    def update_status(self, status):
+    def update_status(self):
+        """Get the torrent status fresh, not from cache.
+
+        This should be used when a guaranteed fresh status is needed rather than
+        `torrent.handle.status()` because it will update the cache as well.
+        """
+        self.status = self.handle.status()
+        return self.status
+
+    @property
+    def status(self):
+        """Cached copy of the libtorrent status for this torrent.
+
+        If it has not been updated within the last five seconds, it will be
+        automatically refreshed.
+        """
+        if self._status_last_update < (time.time() - 5):
+            self.status = self.handle.status()
+        return self._status
+
+    @status.setter
+    def status(self, status):
         """Updates the cached status.
 
         Args:
             status (libtorrent.torrent_status): a libtorrent torrent status
         """
-        self.status = status
+        self._status = status
+        self._status_last_update = time.time()
 
     def _create_status_funcs(self):
         """Creates the functions for getting torrent status"""
--- a/deluge/core/torrent.py
+++ b/deluge/core/torrent.py
@@ -17,6 +17,7 @@ import logging
 import os
 import socket
 import time
+from typing import Optional
 from urllib.parse import urlparse
 
 from twisted.internet.defer import Deferred, DeferredList
@@ -235,8 +236,8 @@ class Torrent:
         self.handle = handle
 
         self.magnet = magnet
-        self._status = None
-        self._status_last_update = 0
+        self._status: 'Optional[lt.torrent_status]' = None
+        self._status_last_update: float = 0.0
 
         self.torrent_info = self.handle.torrent_file()
         self.has_metadata = self.status.has_metadata
@@ -1055,7 +1056,7 @@ class Torrent:
 
         return status_dict
 
-    def update_status(self):
+    def update_status(self) -> 'lt.torrent_status':
         """Get the torrent status fresh, not from cache.
 
         This should be used when a guaranteed fresh status is needed rather than
@@ -1065,7 +1066,7 @@ class Torrent:
         return self.status
 
     @property
-    def status(self):
+    def status(self) -> 'lt.torrent_status':
         """Cached copy of the libtorrent status for this torrent.
 
         If it has not been updated within the last five seconds, it will be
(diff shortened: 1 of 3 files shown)


cas--:
Nice catch, I have been meaning to investigate that one properly

gazpachoking:
I'm looking at all the other bits that call the live handle.status(), and having trouble deciding whether they really need a fresh status or the cached one is good now that it's guaranteed to be relatively recent. If we do want to pull guaranteed fresh status, maybe we should add something like `torrent.get_status(update=False)` That way we are guaranteed to update the cache every time `handle.status()` is called, rather than calling it sometimes and not storing it to the cache.

gazpachoking:
I had a go at answering my own questions raised in the OP, and implemented them as such. Happy to reconsider if anyone has any better plans.

gazpachoking:
Another thing that could be done to fix this, potentially alongside this change, is to call get_torrents_status on all of the torrents we need to check rather than looping over them one by one. Presumably the bulk call to libtorrent is more efficient that getting the status on each torrent individually. Maybe leave the current cache expiry changes in as backup, but also add the get_torrents_status call to do the bulk update for the stop ratio check?

deluge/core/torrentmanager.py
```diff
@@ -279,19 +279,14 @@ def update(self):
                 'Paused',
                 'Queued',
             ):
-                # If the global setting is set, but the per-torrent isn't...
-                # Just skip to the next torrent.
-                # This is so that a user can turn-off the stop at ratio option on a per-torrent basis
-                if not torrent.options['stop_at_ratio']:
```
cas--, line 285: Why was this removed?

gazpachoking: It was never activated, because we are inside an if block that already checks it:
https://github.com/deluge-torrent/deluge/blob/b59549eaf2a222679a7cc53d8a95dbb63c0daf36/deluge/core/torrentmanager.py#L276

deluge/core/torrent.py
```diff
@@ -685,7 +687,7 @@ def force_error_state(self, message, restart_to_resume=True):
             restart_to_resume (bool, optional): Prevent resuming clearing the error, only restarting
                 session can resume.
         """
-        status = self.handle.status()
+        status = self.update_status()
```
cas--, line 690: Naming of `update_status` seems misleading, `get_status` or `fetch_status` perhaps? Maybe with a `cached=False` arg?

deluge/core/torrent.py
```diff
@@ -1024,13 +1026,35 @@ def get_status(self, keys, diff=False, update=False, all_keys=False):
...
+        """Get the torrent status fresh, not from cache.
+
+        This should be used when a guaranteed fresh status is needed rather than
+        `torrent.handle.status()` because it will update the cache as well.
+        """
+        self.status = self.handle.status()
+        return self.status
+
+    @property
+    def status(self) -> 'lt.torrent_status':
```
cas--, line 1039: I'm not sure if it would make it more or less complex to use `cached_property` but moving to a status property looks good :+1:

deluge/core/torrent.py
gazpachoking: Yeah, I did `get_status` at first, until I realized there already is a `get_status` which is a combined options and status dict. Maybe rename `torrent.status` cache to `torrent.lt_status`, and have `torrent.get_lt_status`?

deluge/core/torrent.py
gazpachoking: Took a peek at cached_property, doesn't look like it helps us much.

deluge/core/torrent.py
gazpachoking: So I did the rename to `torrent.lt_status`, with the refresh option now being `torrent.get_lt_status()`. I think this helps clarify the difference between the libtorrent object and the deluge status dict, which is grabbed by `torrent.get_status()`. Happy to do it a different way though if anyone has better ideas.

cas--: Yeah I did want to think about this some more as I am not sure that we should be exposing torrent handle status outside of the Torrent class

deluge/core/torrent.py
```diff
@@ -234,10 +236,11 @@ def __init__(self, handle, options, state=None, filename=None, magnet=None):
         self.handle = handle
 
         self.magnet = magnet
-        self.status = self.handle.status()
+        self._lt_status: 'Optional[lt.torrent_status]' = None
+        self._lt_status_last_update: float = 0.0
 
         self.torrent_info = self.handle.get_torrent_info()
-        self.has_metadata = self.status.has_metadata
+        self.has_metadata = self.lt_status.has_metadata
```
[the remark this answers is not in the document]
gazpachoking: It's the `@property` I added.

deluge/core/torrent.py
gazpachoking: Yeah, I was thinking that as well. It's only used in a few spots in torrentmanager, which could probably be updated.

deluge/core/torrent.py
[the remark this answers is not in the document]
cas--: doh missed that

deluge/core/torrent.py
gazpachoking: The one spot in torrentmanager where it seems sorta important is here https://github.com/deluge-torrent/deluge/pull/369/files#diff-3416e08006d0ef78d79619b83bf6d42c888dc1e5b20ea03801a71e357d5484c9R1615

Which is where it handles the bulk update alert and updates the status cache of all the torrents. I think this was one of the things that improved performance on large numbers of torrents.

cas--: I am not sure about wholesale renaming here, perhaps we keep `status` attribute for now and keep new `get_lt_status`?

gazpachoking: That seems good to me too. I think perhaps `status` could/should be renamed and become private in the future, (maybe `self._lt_status`,) but I don't think that needs to be tied up in this bugfix. I'll fix the PR up when I get a chance.

gazpachoking: Reverted the `.status` rename, but kept the `.get_lt_status()` now.

gazpachoking:
Hmm. How did I mess up the docs building?

deluge/core/torrent.py
```diff
@@ -234,7 +236,8 @@ def __init__(self, handle, options, state=None, filename=None, magnet=None):
         self.handle = handle
 
         self.magnet = magnet
-        self.status = self.handle.status()
+        self._status: 'Optional[lt.torrent_status]' = None
```
cas--, line 239: Should it only be the lt.torrent_status that is the string?

cas--:
My quick testing suggests it's related to typing added in commit [Add type hints for the torrent status methods](https://github.com/deluge-torrent/deluge/pull/369/commits/eac93ca35884ef95c37b055d12427eb0c1a40307)

I'm testing locally to see what is causing it

deluge/core/torrent.py
cas--, line 239: Imported Optional but is string here, overzealous stringifying?

deluge/core/torrent.py
```diff
@@ -1054,13 +1056,35 @@ def get_status(self, keys, diff=False, update=False, all_keys=False):
...
+        """Get the torrent status fresh, not from cache.
+
+        This should be used when a guaranteed fresh status is needed rather than
+        `torrent.handle.status()` because it will update the cache as well.
+        """
+        self.status = self.handle.status()
+        return self.status
+
+    @property
+    def status(self) -> 'lt.torrent_status':
```
cas--, line 1069: It seems that using `lt.torrent_status` here and for above method is causing the autodoc errors. Using the full `libtorrent.torrent_status` works though :shrug:

cas--: Actually using `lt` module seems to be the issue here so any other string would work

cas--: I added some debug to sphinx_autodoc_typehints.process_docstring and I can see the issue is that it is encountering a `Mock object` which seems to be tripping it up. This is likely due to mocking libtorrent and other 3rd party modules in sphinx config.

cas--: Fixed: 62a40521789eab27341

PR #370, opened 2022-02-09
Enable file_completed_alert handling

gazpachoking:
Without adding file_progress to the alert mask, the
TorrentFileCompletedEvent would never fire.

Closes: https://dev.deluge-torrent.org/ticket/3421


cas--:
Thanks! Though it is curious since it must have worked at some point and that alert mask was only added to 1.2

gazpachoking:
Yeah, I'm not super sure. I feel like when I started making a plugin to rename incomplete files this wasn't working, or maybe the event hadn't even been added. We don't use this event anywhere in the code, so 🤷‍♂️

PR #371, opened 2022-02-10
[GtkUI] Fix ETA being copied to neighboring empty cells

gazpachoking:
An optimization that avoided re-rendering treeview cells sometimes
went wrong, and rendered a value from the wrong row when moving
the mouse around the torrentview window.

Closes: https://dev.deluge-torrent.org/ticket/3500


gazpachoking:
@bendikro Only you can save this optimization... or even tell me if it was actually important.

PR #372, opened 2022-02-11
[Config] fixed callLater wrapper func

DjLegolas:
In a6840296, a refactor to the `config` class was introduced.
The change included an internal wrapper for `reactor.callLater`, for lazy
import, but didn't wrap it correctly and therefor, no args/kwargs were
passed to the wrapped method.
Furthermore, the exception was silently ignored.
This caused changes to be ignored and not applied, including
`preferencesmanager._on_config_value_change` callback.


deluge/tests/test_config.py
```diff
@@ -83,6 +84,55 @@ def test_set_get_item_none(self):
...
+
+        def check_called():
+            assert called
+
+        called = False
+        config = Config('test.conf', config_dir=self.config_dir)
+        config.register_change_callback(changed_cb)
+        config['foo'] = 1
+        assert config['foo'] == 1
+        yield task.deferLater(reactor, 0.5, check_called)
```
gazpachoking, line 103: I think we could use `unittest.mock.Mock` to simplify these callbacks a bit, as well as eliminate the problem of having to wait longer than needed for them.
```python
def mock_callback(timeout=0.5):
    """Returns a `Mock` object which can be registered as a callback to test against.

    If callback was not called within `timeout` seconds, it will raise a TimeoutError.
    The returned Mock instance will have a `deferred` attribute which will complete when the callback has been called.
    """
    deferred = defer.Deferred()
    deferred.addTimeout(timeout, reactor)
    mock = unittest.mock.Mock(side_effect=deferred.callback)
    mock.deferred = deferred
    return mock

cb = mock_callback()
config.register_change_callback(cb)
config['foo'] = 1
await cb.deferred
cb.assert_called_once_with('foo', 1)
```

If we need to test registering callbacks like this in any other tests, maybe the mock_callback functions goes somewhere common to import.

DjLegolas: I have created a fixture from this code and changed the tests

deluge/conftest.py
```diff
@@ -31,6 +33,29 @@ def listen_port(request):
...
+    def reset():
+        if mock.called:
+            org_reset_mock()
+        deferred = Deferred()
+        deferred.addTimeout(timeout, reactor)
+        mock.side_effect = lambda *args, **kw: deferred.callback((args, kw))
+        mock.deferred = deferred
+
+    mock = unittest.mock.Mock()
+    org_reset_mock = mock.reset_mock
```
gazpachoking, line 53: I might just spell out the whole `orig` here, not saving much space, and it seems more clear.

deluge/conftest.py
```diff
@@ -31,6 +34,28 @@ def listen_port(request):
     return DEFAULT_LISTEN_PORT
 
 
+@pytest.fixture
+def mock_callback(timeout=0.5):
```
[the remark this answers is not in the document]
DjLegolas: Done

deluge/conftest.py
DjLegolas: Done

cas--:
Nice set of tests :+1:

PR #374, opened 2022-02-12
ui logic/bug of checked move_completed

piengeng:
if move_completed is checked/True, options should be updated, not the other way round

### DIFF
--- a/deluge/ui/gtk3/options_tab.py
+++ b/deluge/ui/gtk3/options_tab.py
@@ -188,7 +188,7 @@ class OptionsTab(Tab):
             ):
                 options[status_key] = widget_value
 
-        if options.get('move_completed', False):
+        if options.get('move_completed', True):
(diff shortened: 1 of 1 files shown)


gazpachoking:
I agree there is a bug here, but I don't think this is quite the right fix. This will keep updating the option even when it hasn't changed. I think the better fix looks something like this:
```python
        move_completed_path = self.move_completed_path_chooser.get_text()
        if move_completed_path != self.prev_status['move_completed_path']:
            options['move_completed_path'] = move_completed_path
```

cas--:
@piengeng Are you going to test and update the PR with that suggestion?

piengeng:
@cas-- test and updated commit with suggestion.

gazpachoking:
Btw, looks like we have a ticket this would fix. https://dev.deluge-torrent.org/ticket/3393

PR #376, opened 2022-02-15
Replace gethostbyname

doadin:
socket.gethostbyname, Translate a host name to IPv4 address format. does not support IPv6 name resolution, and getaddrinfo() should be used instead for IPv4/v6 dual stack support.


doadin:
@cas-- @DjLegolas  This might help/be related to  [#3505](https://dev.deluge-torrent.org/ticket/3505) . Either way it is an improvement since ipv6 was not supported before. If you guys could review thanks!

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,42 @@ def on_connect_failed(reason, host_id):
             return defer.succeed(status_offline)
 
         try:
-            ip = gethostbyname(host)
+            # getaddrinfo returns a list of 5-tuples save this to list
+            addrinfo = getaddrinfo(host, None)
+            ip = []
+            for i in range(0, len(addrinfo)):
+                ip.append(addrinfo[i][4][0])
```
DjLegolas, line 223: change the `i in range` to `i in addrinfo` (and change the names to something even better)

DjLegolas (changes requested):
Looks good.
I looked into that ticket, but the issue is on the console side (getting the port as str, when the func expects int).

deluge/ui/hostlist.py
aresch: Why not something like:
```python
ips = [addrinfo[4][0] for addrinfo in getaddrinfo(host, None)]
```

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,42 @@ def on_connect_failed(reason, host_id):
...
-            return client.daemon.info().addCallback(on_info, host_id)
-        else:
-            # Attempt to connect to daemon with host_id details.
-            c = Client()
-            d = c.connect(host, port, skip_authentication=True)
-            d.addCallback(on_connect, c, host_id)
-            d.addErrback(on_connect_failed, host_id)
-            return d
+        host_conn_list = []
+        for host in ip:
```
aresch, line 232: Use plural named variable (`ips`) since it's a list of IPs.

doadin:
> Looks good. I looked into that ticket, but the issue is on the console side (getting the port as str, when the func expects int).

@DjLegolas Ah, ok I thought maybe it had to do with ipv6 having letters possibly or something like that. I'm not very good with naming but updated for part of the change you and  @aresch suggested. Thanks guys!

deluge/ui/hostlist.py
DjLegolas: @aresch yeah, even better

doadin: @aresch I guess that is possible, my python skills are very basic so, I often miss things like this. If that works I'll update the PR.

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,39 @@ def on_connect_failed(reason, host_id):
...
-            return d
+        host_conn_list = []
+        for host in ips:
+            host_conn_list.append(
+                (
+                    host,
+                    port,
+                    'localclient' if not user and host in LOCALHOST else user,
+                )
+            )
```
cas--, line 236: This can be simplified to a list comprehension 

Also use of `host in ips` is not correct as it overwrites the host check `host in LOCALHOST`. Naming is important to prevent confusion or errors so a better option is `ip` or `host_ip`.

Although it appears that LOCALHOST contains both hostname and ip, although is missing ipv6 `::1`. So the lookup could be replaced with `host_ip in LOCALHOST`

doadin:
@cas-- Honestly I don't follow half of what you said in reviews.

cas--:
@doadin It was a quick review but I can clarify where you need

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,39 @@ def on_connect_failed(reason, host_id):
             return defer.succeed(status_offline)
 
         try:
-            ip = gethostbyname(host)
+            # getaddrinfo returns a list of 5-tuples save this to list
```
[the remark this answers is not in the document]
doadin: um....sure.

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,39 @@ def on_connect_failed(reason, host_id):
...
+            ips = [addrinfo[4][0] for addrinfo in getaddrinfo(host, None)]
         except gaierror as ex:
             log.warning('Unable to resolve host %s to IP: %s', host, ex.args[1])
             return defer.succeed(status_offline)
+        except IndexError as ex:
```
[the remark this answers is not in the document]
doadin: The log message seems less accurate in the case of IndexError which is why I had it separate but ok.

deluge/ui/hostlist.py
```diff
@@ -22,7 +22,7 @@
 
 DEFAULT_HOST = '127.0.0.1'
 DEFAULT_PORT = 58846
-LOCALHOST = ('127.0.0.1', 'localhost')
+LOCALHOST = (('127.0.0.1', 'localhost'), ('::1', 'localhost'))
```
cas--, line 25: This change would break the lookup for an ip or host here. The `::1` ip just need appended to the tuple so not a tuple of tuples

deluge/ui/hostlist.py
```diff
@@ -216,30 +216,35 @@ def on_connect_failed(reason, host_id):
...
-            c = Client()
-            d = c.connect(host, port, skip_authentication=True)
-            d.addCallback(on_connect, c, host_id)
-            d.addErrback(on_connect_failed, host_id)
-            return d
+        host_conn_list = [
+            (
+                host_ip,
+                port,
+                'localclient' if not user and host in LOCALHOST else user,
```
cas--, line 228: I think it is more robust to lookup with `host_ip` instead of `host`

cas-- (changes requested):
Looks great but one breaking change made that needs fixed

deluge/ui/hostlist.py
DjLegolas: Just change it to `('127.0.0.1', 'localhost', '::1')`

doadin: oh wow, oops I was thinking of it was like connection pairs, thats localhost not localclient. ok that makes sense.

cas-- (approved):

doadin:
@cas-- Thanks!

PR #378, opened 2022-02-19
Increase the width of port input box in connection manager

an0n666:
Before:

deluge

After:

deluge2


cas--:
Thank you very much for the contribution but as I was about to merge it I realised that it must be GTK bug which I found and fixed: e15731fcd4f25d66d81224bb5c4812128d542de7

PR #379, opened 2022-02-21
[1015] improve trackers tab

DjLegolas:
closes: https://dev.deluge-torrent.org/ticket/1015


DjLegolas:
So a little update:
I have managed to create the tab in both GTK and Web UIs.
Both with the same information, which is:
1. Tracker URL
2. Status
3. Number of peers in that tracker
4. tracker message(?)

Need to figure out how to do it for DHT, PeX and LSD...

doadin:
@DjLegolas 
Just some thoughts, seeds added? :) Also maybe double click option for opening edit trackers.

As far as dht pex lsd goes, seems we need to track it ourselves, libtorrent doesn't provide it.
https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/peer_info.hpp#L219
   static constexpr peer_source_flags_t tracker
   static constexpr peer_source_flags_t dht
   static constexpr peer_source_flags_t pex
   static constexpr peer_source_flags_t lsd
Seems we can use dht_outgoing_get_peers_alert to run an update, to count,  per info_hash. https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/alert_types.hpp#L2307

for lsd, lsd_peer_alert
lsd https://github.com/arvidn/libtorrent/blob/RC_2_0/include/libtorrent/alert_types.hpp#L1914

flags?
session status example
https://github.com/deluge-torrent/deluge/blob/8ff4683780921111f26fe051e0274aac8afe8bf3/deluge/core/torrent.py#L1076
self.session = component.get('Core').session.status().get_peer_info()
something like that should get peer_info which holds peer flags dht and so on.
to get flags https://gist.github.com/ardhipoetra/457b35882d16289b4c50#file-lt_test-py-L59

Hopefully some of this makes sense/is helpful.

doadin:
you can see how qbittorent handles it here:
https://github.com/qbittorrent/qBittorrent/blob/0cc318664d97697d64a48955aaeb360b6f4d32ab/src/gui/properties/trackerlistwidget.cpp#L322
https://github.com/qbittorrent/qBittorrent/blob/7a910a8cc11a60bfbf294dbfb1bbf0514ba91db7/src/base/bittorrent/peerinfo.cpp#L48

cas--:
Why is there a second tab instead of updating existing trackers tab in GTKUI?

I am seeing changes to tracker_status type but we need to maintain version compatibility and this would be a breaking change. What I would probably suggest here is a new `trackers_status` status property and deprecate `tracker_status`. Be aware from libtorrent docs that num_peers in alert:  _"These are not necessarily all new peers, some of them may already be connected"_.

DjLegolas:
> Why is there a second tab instead of updating existing trackers tab in GTKUI?

It helped me while implementing the new one. I'll replace the original tab when I will rebase everything with better commits.

> I am seeing changes to tracker_status type but we need to maintain version compatibility and this would be a breaking change. What I would probably suggest here is a new `trackers_status` status property and deprecate `tracker_status`.

Hmm, good idea. Will do.

> Be aware from libtorrent docs that num_peers in alert: _"These are not necessarily all new peers, some of them may already be connected"_.

Yeah, will probably try to do something like `qBittorrent` (as @doadin suggested)...

doadin:
@DjLegolas idk if you tried a windows build yet, or I cought you mid changes or anything, but I just did and got:
```
  File "deluge-script.pyw", line 33, in <module>
  File "deluge\ui\ui_entry.py", line 140, in start_ui
  File "deluge\ui\gtk3\__init__.py", line 55, in start
  File "deluge\common.py", line 1353, in run_profiled
  File "deluge\ui\gtk3\__init__.py", line 49, in run
  File "deluge\ui\gtk3\gtkui.py", line 200, in __init__
  File "deluge\ui\gtk3\torrentdetails.py", line 128, in __init__
ImportError: cannot import name 'TrackersTab' from 'deluge.ui.gtk3.trackers_tab' (E:\Program Files\Deluge\deluge\ui\gtk3\trackers_tab.pyc)
```
https://github.com/doadin/deluge/actions/runs/1904315730

DjLegolas:
@doadin whoops... forgot to edit some of the names when copied everything back to the original `tracker_tab.py` file...
Thanks!

doadin:
@DjLegolas np, just trying to help, make sure its working, ill make another run and leave another review. Thanks for doing this nice to finally have this!

DjLegolas:
Found another issue so fixed it too.
In addition, I have added back the `private tracker` info to the details tab.

@doadin check now please.

doadin:
@DjLegolas hmmm, DHT LSX and PEX seem to say "Working" but 0 peers. the one tracker in this torrent says 100 peers, and the status tab says Seeds: 0(939) Peers: 0(10).

doadin:
@DjLegolas I just double checked and its working the same way qbittorent does, except they seems to have a "leeches" that counts connected peers while seeding. so you can see that there are peers coming from DHT etc while seeding since while seeding DHT etc are 0 since we are not connecting to them.

doadin:
It would be my vote to have DHT data displayed the same as trackers where it doesnt matter if they are connected or not but I think if we keep it the same as qbittorent it would be best to also adopt a leeches column like they have.

cas--:
@DjLegolas Remember that clients have to be able to handle connecting to an older version of Deluge daemon so there is no guarantee that the new torrent status keys will be available

cas--:
It's looking good :+1: 

Some further points:

* Private tracker should be placed below Pieces in the Details tab, otherwise it leaves no space for download path
* If any of DHT, PEX or LSD are disabled just don't show them. However I don't like that they are included in the trackers list, should they not be included at the top of Peers tab instead? Can we perhaps change the scope of the this PR to just the trackers tab and leave the peers for another PR?
* For editing how about if a tracker in the list is double-clicked it opens the edit tracker dialog? This would be a simple first-step and more complete solution added later.
* What is the difference between Status and Message column?

DjLegolas:
> @DjLegolas Remember that clients have to be able to handle connecting to an older version of Deluge daemon so there is no guarantee that the new torrent status keys will be available

Added support for using the old status keys and showing only that one entry.


> * Private tracker should be placed below Pieces in the Details tab, otherwise it leaves no space for download path

Changed.
 
> * If any of DHT, PEX or LSD are disabled just don't show them. However I don't like that they are included in the trackers list, should they not be included at the top of Peers tab instead? Can we perhaps change the scope of the this PR to just the trackers tab and leave the peers for another PR?

I'll move all of the DHT, PEX and LSD related things to a new PR when squashing the commits into one.

> * For editing how about if a tracker in the list is double-clicked it opens the edit tracker dialog? This would be a simple first-step and more complete solution added later.

Added support for double-click in trackers tab, in addition to right click.

> * What is the difference between Status and Message column?

Good question. Status is the short string which shows "Announce OK" or some other error.
Message suppose to be the error message for that tracker, when received after sending an announce message. But it looks as it does not being changed.
In addition, it seem as I'm using the wrong one: https://libtorrent.org/reference-Trackers.html#announce-infohash
But it seem that both are not being updated in python binding.

cas--:
> Message suppose to be the error message for that tracker, when received after sending an announce message. But it looks as it does not being changed.

I don't think `tracker.get('message')` would return anything since as you linked `message` is in the `announce_infohash` and each tracker has multiple endpoints so would need to parse tracker['endpoints'] for the message. However we already set the tracker status using that message in on_alert_tracker_error so I don't think we need to implement that just now. Of course we can look make further improvements in the future but let's keep changes simpler just now.

Again thanks for working on this :slightly_smiling_face:

DjLegolas:
> I don't think `tracker.get('message')` would return anything since as you linked `message` is in the `announce_infohash` and each tracker has multiple endpoints so would need to parse tracker['endpoints'] for the message. However we already set the tracker status using that message in on_alert_tracker_error so I don't think we need to implement that just now. Of course we can look make further improvements in the future but let's keep changes simpler just now.

From what I have seen, `tracker['endpoints']` is not being populated with any info.
This is why I wrote it is not being updated...

> Again thanks for working on this 🙂

No problem🙂

DjLegolas:
So, after almost 2 months, back to work on this.
First, I have removed DHT, PEX and LSD from this MR. Will be added in a later one...
Second, I have added cache for `Torrent.trackers` property. The issue with `tracker['endpoints']` was that we are pulling the trackers info from the handler when we call `set_trackers` with the `trackers` argument being None.
Now, each time there is a call for the trackers, we will use a cache and update it every 5 seconds (just like the `status` property).

I'll continue investigate the windows 3.7 crash.
Lint fails because of an internal issue in `black`. Fixed in https://github.com/deluge-torrent/deluge/pull/382

DjLegolas:
I hope this is the last rebase with updates.
I have changed `torrent.set_tracker_status` method so it will also get a `message` param.
This new param will get the error/warning message from the alert (as this method is only being called when handling alerts). In turn, it will be printed if it have any string in it but the tracker message is empty.
Here is how it's going to look:
image

cas--:
Just a quick note to say I had a look at this a few weeks ago and I have concerns about handling of the lt tracker data in the core, I feel it should be parsed to not expose lt data structures and only include required fields. 

Let's try to understand how to correctly expose tracker data in core and built out the UIs from there

DjLegolas:
@cas-- I have refactored the `trackers` property so it will return a dict with the needed data only.
This way, we will both not expose the underling `lt` tracker data and still have it cached.

DjLegolas:
depends on: #427

DjLegolas:
@cas-- I have rebased onto `develop` branch.

In addition, I just noticed something.
Part of the change is the validation of the daemon version using the new `daemon_version_check_min` method.
This means that we need to put some hard-coded version number as parameter, in which the new feature was introduced.
This is something that might be handled in a proper way.
What do you think?

doadin:
@DjLegolas idk if it is just an error on my end but in standalone this works but but in thinclient the trackers tab is empty. It has the headers but the list of trackers is empty.

doadin:
idk if it helps but i pulled in 
1ff9031f3a26c33283b0d96bc808822b16fc8946 
976bee6dfa345644baaefcfa10d2f7aafa6867a1
6a1ed5999fcb6d51d8944b9776f5d4f4f8b4c7f9
and it looks like 
1989d0de73403dc4569b02478aec333252d2cea8
was already merged

doadin:
@DjLegolas 
nvm I think its because in dev build number is still 2.1.1 and you added check for 2.1.2 in daemon version
https://github.com/deluge-torrent/deluge/pull/379/commits/f10a7992784de603851e64ace42decb1d189eb3e#diff-a1606d34bd19014f7edf6355016871765589790d9fcf99cc34d2e507aaac6b97R99

doadin:
@DjLegolas sorry for the multiple messages, and it was because of the daemon version check, now that I updated that though, it is only showing the first tier tracker while in gtk thinclient mode. GTK Standalone and webui seem fine.

zakkarry:
I was discussing the trackers tab a week or two ago with @cas-- and mentioned something I would like to see that I thought would be in the scope of this PR, so rather than opt to open an additional, perhaps mention it here first and see what you think.

When a tracker is in the error state and eventually goes to "skipping" - if I wanted to see _WHAT_ the error was when it was actively announcing in failing, currently I have to force a reannounce and hope that libtorrent honors that request to get the error from the tracker.

It would be nice if there was some way to see the last response from the tracker even when it is in a skipping state.

Potential ways to do this:

1.  Have two separate columns, one for `Message` that shows the last response given from a tracker and a column for the `Status` which would show "Working" or "Error" or "Skipping" messaging.
2. Have it so that when you mouse over the skipping announce text, a tip pops up with the last response announce message. I'm not particularly fond of this, but it does prevent additional columns. Perhaps there is something similar to make it more obvious?
3.  Do not overwrite the tracker response with skipping at all, this seems misleading and probably not at all appropriate, but would accomplish the goal.

Overall, I would opt for solution 1. But I am open to discuss other means of this and give my thoughts if you'd like.

Let me know what you think of this.

DjLegolas:
> @DjLegolas sorry for the multiple messages, and it was because of the daemon version check, now that I updated that though, it is only showing the first tier tracker while in gtk thinclient mode. GTK Standalone and webui seem fine.

does both the client and the daemon taken from this MR? because i have a backward compatibility functionality
if the daemon does not support it, the client will use the original method which will return only 1 tracker.
also, sorry for the late response

DjLegolas:
Rebased and added the ability to click on links when using GTK.

@zakkarry there are both a `status` and `message` columns. i don't remember what each one means, but we have this separation.

zakkarry:
> Rebased and added the ability to click on links when using GTK.
> 
> @zakkarry there are both a `status` and `message` columns. i don't remember what each one means, but we have this separation.

Status is whether the torrent is error'd or announcing (Announce Sent) or OK...

Message is the response, so "Unregistered Torrent" or the like...

DjLegolas:
@cas-- so i think it's time to merge this 😅 
i have rebased on latest `develop` and added a new menu item to update the trackers from the tracker tab in GTK only for now (as i forgot about the WebUI 🤦)

Edit: I have also updated the WebUI

PR #380, opened 2022-02-22
    Add a systemd user service

WhyNotHugo:
This file should be installed into /usr/lib/systemd/user/deluged.service

Unlike the existing service file, this one configures deluge to run as a desktop session user. It does not require root privileges to be controlled either.

### DIFF
--- a/packaging/systemd/deluged.service
+++ b/packaging/systemd/deluged.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Deluge Bittorrent Client Daemon
-Documentation=man:deluged
+Documentation=man:deluged(1)
 After=network-online.target
 
 [Service]
--- a/packaging/systemd/user/deluged.service
+++ b/packaging/systemd/user/deluged.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Deluge Bittorrent Client Daemon
+Documentation=man:deluged(1)
+
+[Service]
+UMask=007
+ExecStart=/usr/bin/deluged -d
+Restart=on-failure
+TimeoutStopSec=300
+Slice=background.slice
+
+[Install]
+WantedBy=default.target


cas--:
When adding the service file to the code I discussed that there was no need for a separate user service file since they are identical service files: https://dev.deluge-torrent.org/ticket/2034#comment:13

I'm not seeing anything here that would change that? Maybe what is missing is additional documentation for installing a user service file?

WhyNotHugo:
They have some differences identical: the user service includes a `Slice` and depends on a different target (the user service manager and system service manager have different targets). Copying one into the others location won't work.

cas--:
Ah I was unaware that user service doesn't include `multi-user.target`. Would setting the service to use `default.target` be a solution as I see it is usually symlinked to multi-user anyway? I know that the `wantedby` can be overridden but could that be more complicated that a separate service file :thinking: 

Why include `Slice`, I not sure I see the benefit here for the single deluged service? And why only for the user service?

WhyNotHugo:
Including the Slice indicates to the service manager that this is a background service. This can be used to handle OOM situations, or prioritising foreground processes. There's no equivalent for system services.

I'm not sure about pulling the system service for `default.target`; I don't know how that interoperates when booting into rescue mode or other special modes. It might be fine, but I can't guarantee it.

Ultimately though, shipping the two distinct files makes life easier for packagers of systemd-based distros since it signals that they can just copy both of them to their respective locations. I would also like to include a `sysusers` file for the system service to automatically handle that user (mostly to de-duplicate work that being repeated downstream). I'd rather tackle that last item separately though to scope the PRa better.

cas--:
Okay this change makes sense, thanks for details.

We would need the same for web service though

packaging/systemd/deluged.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Deluge Bittorrent Client Daemon
-Documentation=man:deluged
+Documentation=man:deluged(1)
```
cas--, line 3: We don't need the section number, it's optional and would open at 1 anyway

cas-- (changes requested):
If we are adding a deluged user service then we should also add a deluge-web user service

WhyNotHugo:
Thanks for wrapping this one up. :+1:

PR #386, opened 2022-06-05
[Packaging] Pinned Pyinstaller to v4.10 and readme update

mhertz:
Pin Pyinstaller to latest v4.x until issue of aborting upon missing typelibs for various unbuilt gst-modules can be properly investigated and resolved. Specific error for one of the modules being:

    36738 INFO: Loading module hook 'hook-gi.repository.Gst.py' from 'C:\\hostedtoolcache\\windows\\Python\\3.9.13\\x64\\lib\\site-packages\\PyInstaller\\hooks'...
      File "<string>", line 7, in <module>
    gi.repository.GLib.GError: g-irepository-error-quark: Typelib file for namespace 'Gst', version '1.0' not found (0)
    36870 ERROR: gi repository 'GIRepository 2.0' not found. Please make sure corresponding package is installed.
      File "<string>", line 4, in <module>
      File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\gi\__init__.py", line 139, in require_version
        raise ValueError('Namespace %s not available' % namespace)
    ValueError: Namespace Gst not available

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -71,12 +71,12 @@ jobs:
 
       - name: Install Python dependencies
         run: >
-          python -m pip install --find-links="C:\GTK\release\python" pycairo PyGObject
+          python -m pip install --no-index --find-links="C:\GTK\release\python" pycairo PyGObject
 
           python -m pip install
           twisted[tls]==22.2.0rc1
           libtorrent==${{ matrix.libtorrent }}
-          pyinstaller
+          pyinstaller==4.10
           pygame
           -r requirements.txt
 
--- a/packaging/win/README.md
+++ b/packaging/win/README.md
@@ -11,13 +11,15 @@ Instructions for building the Deluge NSIS Installer for Windows Vista/7/8/8.1/10
 == Build Steps ==
 
 1.  Build and Install Deluge on Windows.
-2.  Run the pyinstaller from the deluge\packaging\win directory.spec:
+2.  Run pyinstaller from the deluge\packaging\win directory:
 
-    `pyinstaller --clean delugewin.spec --distpath .\packaging\win\freeze`
+    `pyinstaller --clean delugewin.spec --distpath freeze`
 
     The result is a PyInstaller version of Deluge in `packaging\win\freeze`.
 
-3.  Run the NSIS script:
+3.  Run the NSIS scripts:
+
+    `python setup_nsis.py`
 
     64-bit python:
 


packaging/win/delugewin.spec
```diff
@@ -149,7 +149,7 @@ for e, d in executables.items():
         hookspath=[],
         hooksconfig={},
         runtime_hooks=runtime_hooks,
-        excludes=[],
+        excludes=['gi.repository.Gst', 'gi.repository.GstAudio', 'gi.repository.GstVideo', 'gi.repository.GstBase', 'gi.repository.GstPbutils'],
```
cas--, line 152: I don't see `gstreamer` as part of gvsbuild, why do we need this? If it is included then likely better to not build it in the first place by adding `gstreamer` to the [skip step](
https://github.com/deluge-torrent/gvsbuild-release/blob/5c3100a8726af2945915029673853b70aa616373/.github/workflows/release.yml#L50)

cas-- (commented):
Could you clarify what has changed with Pyinstaller 5?

packaging/win/delugewin.spec
mhertz: Gstreamer project isn't built, but for some reason some references to it is picked up by pyinstaller through gi.repository. This was same in pyinstaller 4.x and displayed a "traceback" error in cmd-output during freeze, but the freezing continued till end succesfully regardless - this was one of the things they changed/fixed in pyinstaller 5.x i.e. implememnted a new child/parent method which would now correctly block/quiet the run(main-loop I gather) whenever some sub-process failed.

This fix/workaround is fixing a "mssing typelib" error for various gst components for both pyinstaller 4.x and 5.x, but just now doesn't quit the process in error litle after start.

I have been trying to see if could find the culprit specifically so could report to gvsbuild devs, but haven't yet though.

When you asked if I could clarify, then did you refer to the commit-message, or to you(like here)? Sorry for being dense :)

I understand of-course if you close this, but was also just to note(as you have no "issues" section here) that either need something like this, or, need stipulate 'pip install pyinstaller==4.10' to produce a freezed build currently.

Thanks fo taking a look :)

cas--: It is always good to include the reason why something has changed in the commit message, that way in the future we can use `git blame` to see why the change was made.

As for the error, I don't see any issues with the CI builds, is this a local build issue? What are your steps to reproduce? 

image

I am happy to merge the Readme changes but without more details on the error I am reluctant to include that change

mhertz:
Thanks for hints, appreciated :)

Sorry I don't really understand much CI yet, so always run local, though same as your yaml, except newest python and few less gtk3 packages etc, but have tested py3.9.12/13 and your gtk3 build instead of mine, before making this PR.

I just did again now too.

I am thinking that the new gvsbuild setup(not running build.py directly anymore), broke that test-run and hence gtk3 and more specifically gi.repository not built, hence pyinstaller couldn't fail with that error. I checked a few jobs at your own repo and in one the deluge exe's was half'ish size, "tell-tailing" this(just scanned the log quick and when saw exe sizes, then didn't check more though, yet atleast), and in another run then it stated no log for freeze, though could be retention-rules was stated.

Anyway, i'll look more into it later, and if find I have messed up then will correct of-course, but just was preliminary thinking out loud :)

mhertz:
I'm pretty sure this is correct, and I just got confused in my last post/edit as forgot gvsbuild was used from prebuilt etc - The logs showing pyinstaller 5.1 are producing too small exe artifacts and the runs producing propper size are having no logs though are from a date before pyinstaller 5.0 released, plus I can consistently reproduce here, done countless times in testing e.g. in what to do/exclude etc.

Let me know if need clarification on something, and please change my wording where needed thanks, or ask me to do it, whatever is easiest for you, and same with if you want readme fix splitted off, e.g. if unsure about the other part, blahblah you know what I mean :) 

Thanks!

cas--:
I enabled the CI run for this PR and the package size is still lower, likely due to differences between Pyinstaller 4 vs 5. 

In meantime I would suggest we pin Pyinstaller to latest 4 version until issues can be properly investigated and resolved.

mhertz:
Gotcha. 

I'm very surpiced by that CI run, and will post back if finding something, but I don't really know how troubleshhot CI unfortunetly, so is like doing it in "blind" for me, but will post back if finding something.

I'll ammend-comit/force-push later to trigger CI-rereun and hopefully resolve those errors currently showing - ussually helps but not this time, so will wait some more time.

Thanks again, appreciated you looking into this as said :)

cas--:
Ah gotcha, excellent :+1: 

We'll go with 4.10 and look to resolve for 5 in future.

For reference try to detail the precise error so it easier to understand and search for

``` 
36738 INFO: Loading module hook 'hook-gi.repository.Gst.py' from 'C:\\hostedtoolcache\\windows\\Python\\3.9.13\\x64\\lib\\site-packages\\PyInstaller\\hooks'...
  File "<string>", line 7, in <module>
gi.repository.GLib.GError: g-irepository-error-quark: Typelib file for namespace 'Gst', version '1.0' not found (0)
36870 ERROR: gi repository 'GIRepository 2.0' not found. Please make sure corresponding package is installed.
  File "<string>", line 4, in <module>
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\gi\__init__.py", line 139, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace Gst not available
```

https://github.com/mhertz/deluge/runs/6863381267?check_suite_focus=true#step:10:275

cas-- (approved):

doadin:
idk if it will matter but you can see the branch i use for the builds I post to the forums which the latest got good feedback here: https://github.com/doadin/deluge/tree/dev-mine and you can see here: https://github.com/doadin/deluge/runs/6747148127?check_suite_focus=true it ran with pyinstaller 5.1.

mhertz:
Not Cas, but thanks mate :) I was actually gonna ask Cas if we shouldn't ask you how you proceded, though in end didn't need anyway - I did look at a point earlier though, but didn't found your branch :)

Puzzling me that you can run without --no-index for the py-bindings, as here, both locally and CI, tries to use tar.gz for pyobject and build from source and then failing. I haden't catched that before or else would have posted also that in your thread on forum previously, but for some reason you're not hitting it.

Anyway, thanks for the help, and Cas can ask you further if need more info.

mhertz:
Cas, forgive me for being complete moron here, but should I just reference 

`gi.repository.GLib.GError: g-irepository-error-quark: Typelib file for namespace 'Gst', version '1.0' not found (0)`

and/or

`36870 ERROR: gi repository 'GIRepository 2.0' not found. Please make sure corresponding package is installed.`

and/or

`ValueError: Namespace Gst not available`

or full message like you quoted, for one of the modules.

PR #388, opened 2022-07-09
[Packaging] Bump PyInstaller to 5.2

mhertz:
With release of PyInstaller 5.2 no longer need to pin back to 4.10 to resolve issue of gi-hooks making freeze error out.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -66,7 +66,7 @@ jobs:
           python -m pip install
           twisted[tls]==22.4.0
           libtorrent==${{ matrix.libtorrent }}
-          pyinstaller==4.10
+          pyinstaller==5.2
           pygame
           -r requirements.txt
 


mhertz:
Sorry for little lame PR, but was just to inform that whenever updating pin-holds of build-components next, then don't need to keep hold PyInstaller back to 4.x anymore, thanks to the helpfull PyInstaller devs. 

Also removes the confusing Traceback errors flashing repeatedly over screen during the freeze, which despite not stopping freeze then is confusing/worrying to end-user.

Thank you!

mhertz:
Closing because PyInstaller v5.3 just released today, and this PR was mostly ment as a FYI kinda thing and I don't wanna clutter.

Thanks!

cas--:
That's fine, we don't need to chase the versions unless there is a pressing need e.g. known bug or significant improvement

mhertz:
Thanks mate, do appreciate your reply! :)

PR #389, opened 2022-07-12
Update metainfo install path

tobbez:
The metainfo file was being installed into /usr/share/appdata, but that
path has been deprecated. Metainfo files should instead be installed
into /usr/share/metainfo.

Closes: https://dev.deluge-torrent.org/ticket/3394
Ref: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent


cas--:
Should appdata also be renamed to metainfo?

tobbez:
If you're asking about the file name in the install path, then no. The file should still be named `%{id}.metainfo.xml`.

On the other hand, if you're asking about the variable names in setup.py and the path within the repository, then that's up to you. I can update the PR to change those too if you prefer.

tobbez:
Updated.

There are now no remaining `appdata` references.

PR #390, opened 2022-07-16
Fix serving from custom base path

mueslo:
Twisted requires a child as bytes, not str to work as expected

See also https://stackoverflow.com/questions/42346264/python-twisted-putchild-not-forwarding-expectedly

Possibly same issue encountered here:
https://dev.deluge-torrent.org/ticket/3105
https://dev.deluge-torrent.org/ticket/2677
https://forum.deluge-torrent.org/viewtopic.php?t=56022
https://forum.deluge-torrent.org/viewtopic.php?t=54520
https://forum.deluge-torrent.org/viewtopic.php?t=54668

### DIFF
--- a/deluge/ui/web/server.py
+++ b/deluge/ui/web/server.py
@@ -682,7 +682,7 @@ class DelugeWeb(component.Component):
 
         if self.base != '/':
             # Strip away slashes and serve on the base path as well as root path
-            self.top_level.putChild(self.base.strip('/'), self.top_level)
+            self.top_level.putChild(self.base.strip('/').encode('utf8'), self.top_level)
 
         setup_translation()
 


DjLegolas:
Was in the middle of testing my fix (was thinking of adding a test) when I saw you have opened a PR 😢😅 
This, of course, works.

But, TBH, I think this should be handled somehow different.
The current solution adds the base url as a child of the root, with the root as it's resource. This add the ability to infinite call the base as part of the url, i.e. if the base is `tmp`, this is possible `127.0.0.1:8112/tmp/tmp/tmp/tmp/tmp/...`

mueslo:
Yeah, I also wanted to fix it properly when I saw that... I actually already fixed this particular issue last year (see patch-1 branch) but was also appalled by the weird double resource thing lol. So I never submitted the PR since i thought the web UI needs a general overhaul. But if course, I never got around to it. But I've been using my particular fix for a while now and it's a good stopgap fix.

Theres also the absolute weirdness of the base url being sent as part of the request. I have no idea if there is some particular reason this was done, but it's just seems like a very bad design choice.

cas--:
Thanks for the fix! Realised i forgot to merge this one and was actually fixed in 3565a9a by DjLegolas

PR #391, opened 2022-07-16
[WebUI] Fix base path not being honored

DjLegolas:
Twisted requires a child as `bytes`, not `str` to work as expected.
In addition, it adds the ability to call base in an infinite loop of
calls, i.e. `hostname/base/base/base/...`.

The solution is to build the tree as served.

This is an alternative PR to #390.

### DIFF
--- a/deluge/ui/web/server.py
+++ b/deluge/ui/web/server.py
@@ -453,8 +453,14 @@ class TopLevel(resource.Resource):
         'css/deluge.css',
     ]
 
-    def __init__(self):
+    def __init__(self, children: dict = None):
         resource.Resource.__init__(self)
+
+        if children is not None:
+            for path, child in children.items():
+                self.putChild(path, child)
+            self.js = children[b'js']
+            return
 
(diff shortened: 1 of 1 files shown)


DjLegolas:
Just noticed that serving both root and base is by design for both local and reverse proxy usage...
So this PR is relevant if we want to keep the behavior... (🤦 on myself...)

DjLegolas:
Ok, so now there will be no more infinite loops

cas--:
Can you elaborate on what this is fixing?

DjLegolas:
Sure.
Currently, if you set a base path (`--base tmp`) for the `WebUI`, you will get the following URL `http://1270.0.1:8112/tmp`.
Now, if you will add another base (`http://127.0.0.1/tmp/tmp`), the page will be loaded as usual, even though there is nothing in that URL.
This can be used to enter unlimited amount of children to the URL.

cas--:
Ah I see! So I wasn't happy about adding a hack to a hack so crafted a simpler solution with tests: 4b6ac1f4c4

Let me know if I have missed anything with the implementation

PR #392, opened 2022-07-17
[3536][UI] Add a way to change themes

DjLegolas:
Currently, there is no normal way to change the theme for both UIs (Gtk and Web).
In WebUI, the only way to change the themes is by manually set a value
in the `web.conf` file itself.
In Gtk, the only way to change the themes is by manually set a value
in the command line or set it as env variable.

Closes: https://dev.deluge-torrent.org/ticket/3536


DjLegolas:
Ok, I have decided to add the option to GtkUI also.

PR #393, opened 2022-08-01
[3538][console]fix console connection manager

DjLegolas:
closes: https://dev.deluge-torrent.org/ticket/3538

### DIFF
--- a/deluge/ui/console/modes/connectionmanager.py
+++ b/deluge/ui/console/modes/connectionmanager.py
@@ -197,7 +197,8 @@ class ConnectionManager(BaseMode, PopupsHandler):
                 if chr(c) == 'q':
                     return
                 elif chr(c) == 'D':
-                    host_id = self.popup.current_selection()[1]
+                    host_index = self.popup.current_selection()
+                    host_id = self.popup.inputs[host_index].name
                     self.delete_host(host_id)
                     return
                 elif chr(c) == 'a':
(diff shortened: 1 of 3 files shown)


deluge/ui/console/modes/connectionmanager.py
```diff
@@ -127,12 +127,14 @@ def add_popup(self):
 
     def add_host(self, hostname, port, username, password):
         log.info('Adding host: %s', hostname)
+        if port.isdecimal():
+            port = int(port)
```
DjLegolas, line 131: This can be handled by replacing the `port`'s `add_text_input` in the `add_popup` method with:
```python
popup.add_int_spin_input('port', _('Port:'), min_val=0, max_val=65535, value=DEFAULT_PORT)
```

deluge/ui/console/widgets/popup.py
```diff
@@ -252,6 +252,9 @@ def current_selection(self):
 
     def set_selection(self, index):
         """Set a selected index"""
+        max_input = len(self.inputs)
+        if index >= max_input:
+            index = max_input - 1
         self.active_input = index
```
cas--, line 258: simpler with: 
```
self.active_input = min(index, len(self.inputs) - 1)
```

DjLegolas: changed

PR #394, opened 2022-08-02
[UI][core] Fix interface not being updated in thinclient

DjLegolas:
When changing the interface in both `GTK` and `ConsoleUI`, we call
directly to `deluge.common.is_interface`, which checks the interface on
the local machine and not on the daemon's machine.

The `WebUI` does not seem to have any validation on input.

closes: https://dev.deluge-torrent.org/ticket/3540


cas--:
Ah yes I encountered this issue!

Will need  to handle backward compatibility in the client for daemons that don't have that method exported

DjLegolas:
depends on: #427

deluge/core/core.py
```diff
@@ -817,6 +817,11 @@ def set_config(self, config: Dict[str, Any]):
                 continue
             self.config[key] = config[key]
 
+    @export
+    def is_valid_interface(self, interface: str) -> bool:
```
cas--, line 821: I have concerns about the naming of this method and the underlying problem we are solving. 

In the first instance `valid` seems redundant but can understand clarification as a core method. Even if we were to use it I would suggest using it as a suffix rather than prefix `is_interface_valid`. 

However I am leaning towards either `core.has_interface` or `core.has_interface_name` with perhaps more weight to the latter since `common.is_interface` seems to be my source of the contention. In addition the check if an IP is valid doesn't require a core method call and we don't actually check that the IP is assigned to one of the core host interfaces. 

Perhaps `core.has_interface` will suffice as a core method with a future check that IP is assigned to an interface.

My final concern is that should we prevent the user from assigning an 'invalid' interface, e.g. interface is down briefly, and instead show a UI warning... That does increase the scope of this change so will need to revisit that

cas-- (commented):
I have some thoughts on this change... :smile:

PR #395, opened 2022-08-06
[3541][GTK] Fix Add-torrent-dialog not respecting dirs

DjLegolas:
When adding a new torrent, there is a problem changing the path on
Windows machines, due to the difference between the way the files are
being read from the torrent and how we handle them in addtorrentdialog.
This effects both changing and showing the files in the UI.

Now, all path seperator is being considered and converted to slash '/'.

Closes: https://dev.deluge-torrent.org/ticket/3541

### DIFF
--- a/deluge/ui/gtk3/addtorrentdialog.py
+++ b/deluge/ui/gtk3/addtorrentdialog.py
@@ -379,7 +379,7 @@ class AddTorrentDialog(component.Component):
         self.listview_files.expand_row(root, False)
 
     def prepare_file(self, _file, file_name, file_num, download, files_storage):
-        first_slash_index = file_name.find(os.path.sep)
+        first_slash_index = file_name.find('/')
         if first_slash_index == -1:
             files_storage[file_name] = (file_num, _file, download)
         else:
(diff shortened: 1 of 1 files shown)


cas--:
How can this only just been noticed on Windows? Switching back to using the fixed path separator should be carefully considered so I would like to know more about this especially why it no longer applies. 

Especially since it reverts this original change: https://github.com/deluge-torrent/deluge/commit/14f894959fb59ef3634fe3d531563399288d6bcc

DjLegolas:
This is something that as changed in the `TorrentInfo` class inside the `deluge.ui.common` package in https://github.com/deluge-torrent/deluge/commit/3ec23ad96bfa205c8ae23b662c6da153efbfed3a.
As you can see, all paths in it are being converted to posix path, but in the `addtorrentdialog.py` file we are trying to split the paths according to `os.path.sep`, which in Windows is `\`.

zhen-huan-hu:
Not sure if this is related but in Windows, once a torrent is added, although its directory tree is correctly displayed in the file panel, when you rename a directory, it only affects one file under the directory so that you have to do it repeatedly for all files.

ER-95:
How is this still happening 2 years later?

zhen-huan-hu:
Because there hasn't been a new release where this commit is in.

PR #400, opened 2022-11-12
Import setuptools before distutils

stefanor:
setuptools 60 uses its own bundled version of distutils, by default. It injects this into sys.modules, at import time. So we need to make sure that it is imported, before anything else imports distutils, to ensure everything is using the same distutils version.

This change in setuptools is to prepare for Python 3.12, which will drop distutils.

Fixes: https://bugs.debian.org/1022491

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,13 +12,13 @@ import glob
 import os
 import platform
 import sys
-from distutils.command.build import build as _build
-from distutils.command.clean import clean as _clean
-from distutils.command.install_data import install_data as _install_data
 from shutil import rmtree, which
 
 from setuptools import Command, find_packages, setup
 from setuptools.command.test import test as _test
+from distutils.command.build import build as _build
+from distutils.command.clean import clean as _clean
+from distutils.command.install_data import install_data as _install_data
 
 import msgfmt
 from version import get_version


DjLegolas:
I don't think we need this fix, or at least not in the current state:
1. it breaks `linting` tests.
2. when I have imported `distutils` when I had an old version of `setuptools`, the imported package was from the `stdlib`.
after updating `setuptools` to latest, and doing the same, I got `distutils` from the `setuptools` package.

anyway, I might be mistaken.
and of curse, thanks for this 😄

stefanor:
>  after updating setuptools to latest, and doing the same, I got distutils from the setuptools package.

That depends on whether you have already imported setuptools or not. When you import it, it injects its distutils module into sys.modules.

stefanor:
Obviously the best change to make is to replace any distutils imports with something from pure setuptools.

DjLegolas:
> > after updating setuptools to latest, and doing the same, I got distutils from the setuptools package.
> 
> That depends on whether you have already imported setuptools or not. When you import it, it injects its distutils module into sys.modules.

Well, I don't see this behavior, (tested also on `ubuntu`):
```batch
C:\Users\dj_le>py
Python 3.10.6 (tags/v3.10.6:9c7b4bd, Aug  1 2022, 21:53:49) [MSC v.1932 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import distutils
>>> distutils.__file__
'C:\\Python310\\lib\\site-packages\\setuptools\\_distutils\\__init__.py'
>>>
```


> Obviously the best change to make is to replace any distutils imports with something from pure setuptools.

Agreed. But not all of the functionality is supported by `setuptools` yet, i.e. all of the `distutils.command`s we use in `deluge`

stefanor:
> Well, I don't see this behavior, (tested also on ubuntu):

Thanks, I had been surprised that I was in the vanguard of filing these issues. But I see now that Debian is missing `distutils-precedence.pth` that does the magic to hook distutils at interpreter startup time. Obviously at a small performance penalty.

So, you won't see it unless you have setuptools > 60 and don't have `distutils-precedence.pth`.

If you do want to reproduce the issue, use Debian unstable outside a virtualenv. Or  remove `distutils-precedence.pth` from the virtualenv.

I totally understand if you don't take the patch, because this is a bug in Debian, I'd say it  is, I just filed https://bugs.debian.org/1025216

cas--:
Thanks for providing a potential fix, I was surprised that a fragile import order change was required. Based on your latest findings I feel it probably is better that debian fixes the issue rather than attempt to fix every project importing distutils.

Decomposing the setup.py file and replacing distutils has been on my mind for a long time but not got around to it...

PR #401, opened 2022-11-30
Fixing alignment issues with the FilesTab progress bar

GetOutOfMyBakery:
# What
Two issues:
- Height offset for the white overlayed text was slightly off, causing the background darker text to be slightly visible when replaced with the foreground white text: 

- Progress offset caused the white foreground text to be 10 pixels ahead in progress, meaning the replacement text wasn't aligned with the background progress bar:

### DIFF
--- a/deluge/ui/web/css/deluge.css
+++ b/deluge/ui/web/css/deluge.css
@@ -283,7 +283,7 @@ dl.singleline dd {
 }
 
 .x-progress-renderered .x-progress-bar .x-progress-text {
-    margin-top: -1px;
+    margin-top: -2px;
     height: 18px;
 }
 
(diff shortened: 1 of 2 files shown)


deluge/ui/web/css/deluge.css
```diff
@@ -283,7 +283,7 @@ dl.singleline dd {
 }
 
 .x-progress-renderered .x-progress-bar .x-progress-text {
-    margin-top: -1px;
+    margin-top: -2px;
```
GetOutOfMyBakery, line 286: Actually, this causes all the other (non-FilesTab) progress bar's to be offset.

DjLegolas:
Looks nice!
Have you tested it with several browsers by any chance?

PR #403, opened 2022-12-02
[Common] Replace `pkg_resources` with `importlib`

DjLegolas:
Related to CI stalling/Timed out fixed in 2a945de

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,9 +7,6 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-env:
-  SETUPTOOLS_ENABLE_FEATURES: "legacy-editable"
-
 jobs:
   test-linux:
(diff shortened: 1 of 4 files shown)


cas--:
I am concerned about this breaking in subtle ways, perhaps for the plugins. Based on my previous attempts at using importlib and running into issues, along with requiring third-party lib (could be vendored if needed).

bendikro:
It would be nice with a commit message that describes why this should be done :)

cas--:
Yeah it's also good to give a quick reason in the commit. However I have refactored this to be cleaner so don't worry about updating it, thanks the input to resolve it. Once I've pushed commit I'll close this :+1: 

The only thing on note is that we still are using pkg_resources for plugins. However there is a big problem there since with plugins eggs importlib only creates temporary files when extracting data files we'll need to find a better solution, either return as a file stream or extract and cache files in the system ourselves...

PR #405, opened 2022-12-22
Losslessly compress PNG images

anthonyryan1:
Using Efficient-Compression-Tool, we managed to reduge image sizes by 9.13KB out of 241.88KB (3.7741%) without changing visual appearence.

It's a small improvement, but we get something for nothing and that's nice.


cas--:
What are the command arguments used for compression? 

I would prefer that this is integrated into a script or workflow so it's repeatable. For example there is an existing script used to generate the Deluge icons using rsvg-convert and zopflipng. Also I noticed there is no ECT Linux binary which is an inconvenience.

I think that using [pngquant](https://pngquant.org/) to reduce colour palette and [oxipng](https://github.com/shssoichiro/oxipng) would give equal or better gains.

Also note that changing the icon in test will break the tests so will need reverting.

PR #406, opened 2023-01-22
[GTK3UI] replace "utilizing" -> "using"

partev:
In this context the word "using" is more appropriate.


cas--:
Thanks for suggestion but changing this would also require translations to need updating and personally I don't see it being more appropriate therefore I am closing the PR.

Found an article that sheds light on usage: https://www.merriam-webster.com/words-at-play/is-utilize-a-word-worth-using

PR #412, opened 2023-02-14
Proposal about i2p

jiigen:
Proposal to fix the missing peers when using the I2P proxy.


cas--:
A quick review that this is a good start but we need to be aware of backward compatibility, both with libtorrent version and daemon version.

jiigen:
> A quick review that this is a good start but we need to be aware of backward compatibility, both with libtorrent version and daemon version.

What requires the last version of libtorrent on git (and doesn't work with other versions) is the check on i2p_socket flag which can be eventually subsituted by a check on ip (i2p peers have 0.0.0.0:0 address) the "pid" i've introduced is even in older versions so it is nit a problem. Anyway this code is just a matter of visualization everything works fine even without it but it's "ugly" to see an empty peer list :-)

deluge/core/torrent.py
```diff
@@ -821,6 +821,9 @@ def get_peers(self):
                     "progress": float,
                     "seed": bool,
                     "up_speed": int
+                    "pid": str
+                    "i2p_socket": bool
```
cas--, line 825: Please fix typo, should be `i2p`

deluge/core/torrent.py
```diff
@@ -838,9 +841,14 @@ def get_peers(self):
                 client = 'unknown'
 
             try:
-                country = component.get('Core').geoip_instance.country_code_by_addr(
-                    peer.ip[0]
-                )
+                #if it's an I2P peer use I2P logo as flag
+                #added ip.png to ui/data/pixmaps/flags
```
cas--, line 845: Should be removed since should not be describing changes in comments, those are for commit messages

deluge/core/torrent.py
```diff
@@ -838,9 +841,14 @@ def get_peers(self):
                 client = 'unknown'
 
             try:
-                country = component.get('Core').geoip_instance.country_code_by_addr(
-                    peer.ip[0]
-                )
+                #if it's an I2P peer use I2P logo as flag
```
cas--, line 844: Can simplify this comment to just say 

```suggestion
                # Use I2P logo as country flag
```

deluge/ui/gtk3/peers_tab.py
```diff
@@ -282,30 +281,17 @@ def _on_get_torrent_status(self, status):
...
-                    # This is an IPv4 address
-                    ip_int = sum(
-                        int(byte) << shift
-                        for byte, shift in zip(
-                            peer['ip'].split(':')[0].split('.'), (24, 16, 8, 0)
-                        )
-                    )
-                    peer_ip = peer['ip']
+                #requires:
+                #https://github.com/arvidn/libtorrent/commit/02884d367b5641136cf9938d27fdb8d685b32c6d
```
cas--, line 285: this comment should be removed as it's not relevant here and should be in commit message

cas--:
> What requires the last version of libtorrent on git (and doesn't work with other versions) is the check on i2p_socket

I can see the relevant changes but the code actually needs to handle when that attribute is missing for backwards compatibility.

For the same reason the UI client code needs to handle connecting to older daemons that do not have `pid` or `i2p` keys in the peer dict from get_peers.

We really should have tests associated with the core changes.

deluge/core/torrent.py
jiigen: where???

jiigen:
> > What requires the last version of libtorrent on git (and doesn't work with other versions) is the check on i2p_socket
> 
> I can see the relevant changes but the code actually needs to handle when that attribute is missing for backwards compatibility.
> 
> For the same reason the UI client code needs to handle connecting to older daemons that do not have `pid` or `i2p` keys in the peer dict from get_peers.
> 
> We really should have tests associated with the core changes.

In the latest libtorrent version (2.0.9.0) there have been some changes in managing i2p:
1 - It's no more needed (but according to devs it has never) to set a proxy for i2p but it's enough to set i2p_hostname and i2p_port parameters.
2 - there's now a function to get the i2p address (destination): peer_info::i2p_destination() (which under my request has a python binding too) so using the pid can be now dropped.
3 - There are new option to fine-tune the i2p SAM connection: tunnels outbound and inbound lenght and quantity.

So I think that my pull request is now outdated.

PR #414, opened 2023-03-07
[Tests] Autorun async tests with pytest_twisted

gazpachoking:
Since all of our async tests should be run with twisted, it's annoying to have to decorate them with pytest_twisted.ensureDeferred. Forgetting to do this will cause the test to pass without actually running. This changes the behavior to detect all coroutine function tests and mark them to be run by pytest_twisted the same way the decorator does.


cas--:
Perfect! :tada:

PR #415, opened 2023-03-13
Crush all the PNGs

stonedDiscord:
i ran 3 tools, pngcrush, ect and pinga on all PNG files to reduce their size without changing the quality


stonedDiscord:
oh wait actually this is the same thing as #405

PR #416, opened 2023-04-11
[WebUI] use setTimeout instead of setInterval

xuefer:
When server is busy or the request is slow for big file list, WebUI still requests for new update blindly. "Connection lost" is often triggerd.

Change to only ask for update 2s after reponse (either success or error)


SpikedCola:
Thank you for this fix! I was having this exact problem. I hope it is accepted.

PR #418, opened 2023-04-11
[WebUI] edit trackers for multiple torrents

xuefer:


cas--:
There is an existing Trac ticket for editing trackers on multiple torrents, does this handle the use-cases defined there? I am concerned about unexpected outcomes e.g. user not noticing multiple torrents selected etc..

xuefer:
trac ticket: https://dev.deluge-torrent.org/ticket/2047

your concern is not solved by this pull request. 'search and replace' is better way than this pull request, IF it's implemented in the UI (gtk/web)

cas--:
Yeah I think text or confirmation popup to warn that the action will replace all trackers for multiple torrents is needed

dethmourne:
FWIW "search/replace" doesn't resolve at least one use case I have - I swapped from a different torrent client, and ALL the torrents from a private tracker got their tracker list wiped out in the process. Since I wasn't keeping backup torrent files, I had just copied the 'backups' from that program's app data folder, but those backups didn't contain the tracker list. I have all of those torrents siloed into a specific label now, but now I need to add tracker data to each one by hand :(

XinfitX:
Any progress on having it in the next version?

PR #420, opened 2023-04-11
[WebUI] request always in utf8

xuefer:
assume request from browser is in UTF-8 regardless of whatever encoding env set for running server python


cas--:
Python decode method defaults to utf8 so this is unnecessary

PR #421, opened 2023-04-11
[Core] fix handling for unicode path

xuefer:


cas--:
I would like to know how these Unicode errors are being propagated. Can you provide examples or tests?

cas--:
I cannot accept this change without an actual root cause and the UnicodeError exception handler hides the problem.  It is surprising that a decoded byte string cannot be encoded again.

These are also two different fixes here, one is related file renaming and the other related to local download directory paths where there really should not be an issue.

xuefer:
> I cannot accept this change without an actual root cause and the UnicodeError exception handler hides the problem. It is surprising that a decoded byte string cannot be encoded again.
> 
> These are also two different fixes here, one is related file renaming and the other related to local download directory paths where there really should not be an issue.

You're right, back at the time I introduced this patch there're encoding bugs but stated fixed later by deluge-torrent officially. I may make some time to disable this patch and see if there is any new torrent trigger unicode error without my patch.

PR #424, opened 2023-05-25
[GTKUI] Add torrent dialog incorrectly removes dir

bendikro:
When adding a torrent in the add torrents dialog, containing only a single directory with a single file inside, the directory is not included as a prefix to the filename.
The prefix is added only if there are multiple files inside the directory.

Fix by adding the prefix also when there is only one file inside a dir.

Closes: https://dev.deluge-torrent.org/ticket/3602


cas--:
These test failures might be due to a new release of libtorrent

PR #430, opened 2023-08-21
[Core] implement synchronous and v2 metafile creation

rcarpa:
Assuming an entity which has access to deluge daemons on multiple storage servers, 
this will allow to generate a torrent file on a source server, followed by triggering
the transfer from multiple source locations towards the destination storage server.

In particular, I'm interested to add support for deluge/bittorent as a wire transfer protocol in [rucio](https://github.com/rucio/rucio)

The individual commit messages contain additional information about the changes.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -17,7 +17,7 @@ from base64 import b64decode, b64encode
 from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.request import URLError, urlopen
 
-from twisted.internet import defer, reactor, task
+from twisted.internet import defer, reactor, task, threads
 from twisted.web.client import Agent, readBody
 
 import deluge.common
@@ -992,13 +992,13 @@ class Core(component.Component):
         path,
         tracker,
         piece_length,
-        comment,
-        target,
-        webseeds,
-        private,
-        created_by,
-        trackers,
-        add_to_session,
+        comment=None,
+        target=None,
+        webseeds=None,
+        private=False,
+        created_by=None,
+        trackers=None,
+        add_to_session=False,
     ):
         log.debug('creating torrent..')
         threading.Thread(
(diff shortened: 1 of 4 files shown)


deluge/core/core.py
```diff
@@ -1017,6 +1017,35 @@ def create_torrent(
             ),
         ).start()
 
+    @export
+    def create_torrent_sync(
```
cas--, line 1025: If I'm not mistaken we don't need a new method, create_torrent can return a deferred without affecting existing client usage

deluge/metafile.py
```diff
@@ -10,10 +10,14 @@
...
 #
 
+import copy
 import logging
 import os.path
 import time
 from hashlib import sha1 as sha
+from hashlib import sha256
+
+import libtorrent as lt
```
cas--, line 20: Cannot import libtorrent here, this burdens UI clients with libtorrent requirement

deluge/core/core.py
```diff
@@ -999,6 +999,8 @@ def create_torrent(
         created_by=None,
         trackers=None,
         add_to_session=False,
+        include_v1=True,
+        include_v2=False,
```
cas--, line 1003: I don't feel it is a good idea to have these libtorrent implementation details leaking into the functions and API. It would be better to use something like a TorrentFormat Enum and have a `format` param that can accept str or Enum.

A tentative implementation attempt:

```
class TorrentFormat(str, Enum):
     V1 = 'v1_only'
     V2 = 'v2_only'
     HYBRID = 'hybrid'

     @classmethod
     def _missing_(cls, value):
         if not value:
             return cls.V1

         value = value.lower()
         for member in cls:
             if member.value.startswith(value):
                return member

     def to_lt_flag(self):
         if self.value == 'v1_only':
             return 64
         if self.value == 'v2_only':
             return 32
         return 0
```

This can be better represent available options to users.

deluge/core/core.py
rcarpa: I updated the PR to return a Deferred directly from the existing function. Looks like you were right.

deluge/metafile.py
rcarpa: I imported libtorrent in the corresponding function only

deluge/core/core.py
rcarpa: Did it a bit differently. The called it torrent_format everywhere to avoid clash with the builtin `format` function. The format names are `v1`, `v2` and `hybrid` (without the _only suffix). And enforce a valid format string in _missing_. Instead of returning `v1` on invalid format.

rcarpa:
Sorry for taking so long. Was overwhelmed by a couple of more urgent tasks.

rcarpa:
@cas-- , would it be possible to create a release ? (could be an -dev or an -alpha release). It would be very useful to be able to use this new features by just installing deluge from pip.

PR #432, opened 2023-09-05
[Core] Implement SSL peers support

rcarpa:
This feature is interesting when multiple deluge instances are managed by the same administrator who uses it to transfer private data across a non-secure network.

A separate port has to be allocated for incoming SSL connections from peers. Libtorrent already supports this. It's enough to add the suffix 's' when configuring libtorrent's listen_interfaces. Implement a way to activate listening on an SSL port via the configuration.

To actually allow SSL connection between peers, one has to also configure a x509 certificate, private_key and diffie-hellman for each affected torrent. This is achieved by calling libtorrent's handle->set_ssl_certificate. The certificates are only kept in-memory, so they have to be explicitly re-added after each restart. Implement two ways to set these certificates:
- either by putting them in a directory with predefined names and letting deluge set them when receiving the corresponding alert;
- or by using a core api call.

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -13,6 +13,7 @@ import os
 import shutil
 import tempfile
 from base64 import b64decode, b64encode
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.request import URLError, urlopen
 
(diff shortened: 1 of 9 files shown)


rcarpa:
Some core-functions added in this PR rely on a pending PR in libtorrent to expose the needed function in the python bindings: https://github.com/arvidn/libtorrent/pull/7509 . So will be available in the next libtorrent release the earliest. Is there any convention on how to handle such cases in deluge?

deluge/core/core.py
```diff
@@ -822,6 +859,11 @@ def get_listen_port(self) -> int:
         """Returns the active listen port"""
         return self.session.listen_port()
 
+    @export
+    def get_ssl_listen_port(self) -> int:
+        """Returns the active SSL listen port"""
+        return self.session.ssl_listen_port()
```
cas--, line 865: This is a tricky one since we are both adding a new RPC method and relying on a new libtorrent exposed method...

Sometimes we check for specific libtorrent version but it likely is easier in this case to catch the AttributeError and log a warning. For the clients we could either re-raise the exception with a nice message or return an invalid port number, not sure which would be better long-term

```suggestion
        try:
            return self.session.ssl_listen_port()
       except AttributeError:
            log.warning("libtorrent version >=2.x required to get active SSL listen port")
            return -1
```

deluge/core/torrentmanager.py
```diff
@@ -1339,6 +1340,48 @@ def on_alert_torrent_checked(self, alert):
...
+                dh_params_path = params_path
+                break
+        if dh_params_path:
+            for file_name in [torrent_id, 'default']:
+                crt_path = os.path.join(base_path, file_name)
+                key_path = crt_path + '.key'
+                if os.path.isfile(crt_path) and os.path.isfile(key_path):
+                    certificate_path = crt_path
+                    private_key_path = private_key_path
+                    break
```
cas--, line 1369: Is there a consensus on file naming convention here? I notice that the cert file is missing an extension which doesn't seem quite right.

I feel that the files should end in `.pem` to indicate the encoding. The [libtorrent tests](https://github.com/arvidn/libtorrent/blob/5ce4c516479085eda9327d35891b7732d6d5037b/examples/client_test.cpp#L1067) use:
* dhparams.pem
* <torrent_id>.pem
* <torrent_id>_key.pem

Also these default files will need documented. 

I wonder if we can make the code cleaner with pathlib and for else:

 ```suggestion
        certs_dir = Path(self.config['ssl_torrents_certs'])
        dh_params_file = certs_dir / f'{torrent_id}.dh'
        if not dh_params.is_file():
             dh_params_file = certs_dir / 'default.dh'
   
        for filename in [torrent_id, 'default']:
            cert_file = certs_dir / f'{filename}.pem'
            priv_key_file = certs_dir / f'{filename}_key.pem'
            if dh_params.is_file() and cert.is_file() and priv_key.is_file():
                break
        else:
            log.error('<Certs not found...>')
            return
                   
```
In fact this feels like reusable code for a common helper function

cas-- (changes requested):
This should be a useful addition, I've added comments for changes and happy to hear your thoughts on how you expect the feature to be used.

If possible it would be quite nice to have some tests.

rcarpa:
Thank you for the review. I updated the pull request. I slightly deviated from the naming convention you proposed for the files on the disk, because it seemed strange to have a specific naming for `dh_params.pem`, while calling it `<torrent_id>.dh.pem` for per-torrent configuration. Files are now automatically saved on disk when set via RPC. I also added a test which creates a seeder and a leecher instance and performs an actual transfer between the 2 using SSL torrents. I did some minor adjustments to facilitate adding this test case. I also rebased on top of  #430 which allows to slightly simplify the test case (create and return the torrent on the seeder in one call) + avoids merge conflicts because both PRs change the signature of create_torrent.

rcarpa:
Just rebased. No functional changes in last force push.

cas--:
Thank you for implementing this feature, sorry it took me so long to get it merged

PR #433, opened 2023-09-09
Fix Errors Due to Deps

doadin:


doadin:
The pillow change fixes pip trying to compile pillow from source but is missing zlib headers(maybe others), you could set these up on your system/CI/CD or 9.5.0 works. They also recommend using the wheel if you don't want setup the headers.

The zope.interface change fixes a error on windows and linux in test where on python 3.7 you get pkg_resources.DistributionNotFound: The 'zope-interface>=5' distribution was not found and is required by twisted even though zope.interface v6 is installed.

The change to pyinstaller fixes a typelib error in freeze on windows saying freetype 2.0 is missing.

doadin:
I don't know if you can do both greater and less then in requirements but also not sure that it matters in this case as I think the versions of the deps are old enough that versions older are not likely to be used with the versions of python deluge supports, but idk.

doadin:
So I thought zope.interface change fixes python 3.7 but apparently not test still pass on 3.10, so I guess this is a python 3.7 issue?

mhertz:
It's same issue also for py3.9(zope.interface), and I just reverted pyopenssl and cryptography to earlier version for quick fix when I ran into it last week.

Anyway, I looked little into it now, and don't have more time now, but found seemingly for some reason an older setuptools(47.1.0) is installed for python's under 3.10(tested 3.7, 3.9 and 3.10), and when upgrading setuptools afterwards to newest, then works. Probably need pin setuptools to 68.0.0, as latter two versions are py3.8 and up.

Also, in requirements.txt, then twisted should be pinned to 22.10, like I believe in CI, as else: Nameerror fdesc is not defined, when newest twisted installed. Haven't looked into that though.

Thanks! :)

doadin:
seems related: https://github.com/zopefoundation/zope.interface/issues/255

doadin:
@mhertz the pyinstaller issue is when running deluge after packaging not in the packaging process sorry I was not clear.

mhertz:
> seems related: [zopefoundation/zope.interface#255](https://github.com/zopefoundation/zope.interface/issues/255)

Good catch, yeah must've been fixed here presumably then: https://github.com/pypa/setuptools/pull/3153 for setuptools 62. Just strange because first thing I tried myself was mess with that(change install name), to no avail, and see requirements.txt and setup.py has the canonical name regardless, but whatever.

I guess the setuptools change should be added little up too in CI file, as still old version in log.

Sorry didn't ment to butt in here again, unless called, but will adhere to now promise ;) Thin line to walk for guessing if beeing perceived helpful VS annoying you know :) Thanks as said.

doadin:
@mhertz @cas-- I think this is all the changes I am going to put forth. The team over at twisted are looking into the issue with v23 and looking to improve their CI to prevent issues in the future. For now v22.8 is latest working.

cas--:
Thanks for the work identifying the build issues, I prefer that we identify and resolve each issue separately to properly understand the issue that is being fixed, as such I referenced your changes in new commits. 

Just to note that the versions specified in requirements are usually there as a project minimum/maximum rather than for CI/CD and shouldn't be removed without checking the commit history. There are likely some improvements to how we handle the dependencies for project, packaging and CI/CD.

Also until we specifically drop support for Python 3.7 I still want the runners for it. The total number of Python versions to test against should also be small so we don't have to wait too long for them all to finish. There could be an improvement where we run a single Python version for the PR/tests and then run against all versions once merged to develop branch.

doadin:
@cas-- Thank you for making the required changes. I am sorry though I might have not been clear maybe? Or I just didn't notice a reference to it in your commits. However the change for pillow is not just for compatibility with python versions but there is also missing header files for compiling from source. So maybe a comment in a commit mentioning zlib source headers missing from pillow source code for compiling pillow from source? idk. Again sorry either way if you did or didn't. And big thanks for making the changes!

cas--:
Ah yeah I did find the other Pillow issue which is isolated to Windows 32-bit builds which as you say is due to missing wheels and trying build from source: 6c9b058d81639

PR #440, opened 2023-12-29
Fix the icon to fit within the row in the main window tracker column.

cross:
For me at least, Safari on Mac OS X, the tracker icon significantly overflows in the Tracker column of the torrent list.  (It does show the correct size in the Trackers filter, though.  Different CSS.)

This change causes it to constrain down to the height of the column and display correctly.


cross:
Deluge-web-overlarge-icons

Deluge-web-contained-icons

Before and after screen-shots.

cross:
Hey folks.  Anyone available to kick this through the needed workflows?

PR #441, opened 2024-01-16
Update main_window.tabs.ui

DanielHuey:
Some labels were showing wrong stats ("Down speed" was showing stats for "uploaded" and vice versa)  so i swapped them. 
Please do help me check to confirm whether I am correct.

torrent thingy
The image above shows an example of the current stats shown by "Uploaded:" and "Down Speed:" before this patch.


cas--:
I thought this was already fixed in develop at least? :thinking:

DanielHuey:
Oh, pardon me, you are right; that commit fixes it.
I had gotten the application through the Debian hosts on Bunsen Linux, and I thought it hadn't yet been patched. The Debian repositories are not frequently updated to ensure stability, so I believe many others will still come across the issue for a while.

Thanks for verifying.

PR #443, opened 2024-02-21
[3595][Core] add ability to download files from daemon

DjLegolas:
When we use thin client mode, we will download tracker's icon from the client endpoint.
This means we are leaking the IP of the client location, instead using the daemon, which is already connected to the tracker. Therefor, an ability to download files from the daemon is added.

Closes https://dev.deluge-torrent.org/ticket/3595

### DIFF
--- a/deluge/core/core.py
+++ b/deluge/core/core.py
@@ -10,7 +10,9 @@
 import glob
 import logging
 import os
+import random
 import shutil
+import string
 import tempfile
 from base64 import b64decode, b64encode
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


DjLegolas:
@cas-- as part of this fix, we need to change some (or even, all) of the calls to `deluge.httpdownloader.download_file`.
But, currently, I didn't managed to use `client.core.download_file` with `callback` param.
For example, it's being used in the `add_from_url` method in the `GtkUI`, when we show a progress bar for the download.

Do you think we need to support this functionality?
Or should we download files from the client? Which means an IP leak again.

cas--:
Hmm this is an interesting problem and you are likely correct with the long-term solution. However initially we will want to have an option to stop this happening client side :thinking:

DjLegolas:
@cas-- i managed to overcome the callback issue. can you take a look?
i'll add the support for other calls as well.

PR #445, opened 2024-03-19
[Console] Improve interactive-mode preferences saving

mhertz:
The bottom options for cancel/apply/ok where confusing for end-users as being checkboxes needing spacebar prepended to activate firstly, before return/enter to activate said previous selection, but changed now to omit. Also fixed not showing canceled options as sticking.

### DIFF
--- a/deluge/ui/console/modes/preferences/preferences.py
+++ b/deluge/ui/console/modes/preferences/preferences.py
@@ -139,7 +139,13 @@ class Preferences(BaseMode, PopupsHandler):
         ]
 
         self.action_input = SelectInput(
-            self, None, None, [_('Cancel'), _('Apply'), _('OK')], [0, 1, 2], 0
+            self,
+            None,
+            None,
+            [_('Cancel'), _('Apply'), _('OK')],
+            [0, 1, 2],
+            0,
+            require_select_action=False,
         )
 
     def load_config(self):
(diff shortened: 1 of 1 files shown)


cas--:
It is good to explain the problem that is being solved in the commit message because it is not always self-evident and it took a while for me to understand this issue.

I can see now that the preferences close dialog actions by default requires a two-step action, first select one-of `Cancel`, `Apply`, `Ok` with space then hit enter. This is probably counter-intuitive to what the user expects and since the default action is cancel their changes to preferences are lost. I can see that your fix would work but I feel the correct fix here is to set the `SelectInput` arg `require_select_action` to `False` thus explicitly denoting the type of action we require. 

There is a second more subtle issue where if the user hits cancel in preferences then the changes made are not discarded so the user might still think those preferences have been saved. Something a bit hacky like this should solve it: 

```diff
-    def _update_preferences(self, core_config):
+    def _update_preferences(self, core_config, console_config=None):
         self.core_config = core_config
         for pane in self.panes:
-            pane.update_values(core_config)
+            if isinstance(pane, InterfacePane) and console_config:
+                pane.update_values(console_config)
+            else:
+                pane.update_values(core_config)
 
     def _actions_read(self, c):
         self.action_input.handle_read(c)
         if c in [curses.KEY_ENTER, util.KEY_ENTER2]:
             # take action
             if self.action_input.selected_index == 0:  # Cancel
+                # Reload stored config for panes
+                self._update_preferences(self.core_config, self.console_config)
                 self.back_to_parent()
             elif self.action_input.selected_index == 1:  # Apply
                 self._apply_prefs()
```



This also raises the question about how the preference panes are refreshed when other interfaces update the core config since it appears core config is only loaded once per session (unless Apply selected).

cas-- (changes requested):
See above comment

mhertz:
I'm sorry Cas, I didn't actually understood those where checkboxes and thought regular buttons(so return/enter only), so thought actual bug and not just usability issue, thanks for explaining, appreciated and i'll try be better with verbosity, which agreed particularly bad here, despite my confusion.

As said before, if feel faster for you to just make yourself (here/elsewhere) then please just close and don't need explain, little anxious about being a nuisance honestly :) You understand hopefully. Thanks!

cas--:
Thanks for understanding, these are helpful commits and I am happy to have the discussion to ensure the changes are what we expect them to be :)

PR #447, opened 2024-03-29
[Console] Fix 'move' command hanging when done

mhertz:
Console.get_torrent_name() expects string, but was given list from console.match_torrent(), so NoneType breaking join() in move message. Simplified and fixed now.


deluge/ui/console/cmdline/commands/move.py
```diff
@@ -46,7 +46,8 @@ def handle(self, options):
         for t_id in options.torrent_ids:
             tid = self.console.match_torrent(t_id)
             ids.extend(tid)
-            names.append(self.console.get_torrent_name(tid))
+            for name in tid:
+                names.append(self.console.get_torrent_name(name))
```
cas--, line 50: This is a bit confusing with `for name in tid` and we could simplify this code a lot:

```suggestion
    ids = self.console.match_torrents(options.torrent_ids)
    names = [self.console.get_torrent_name(id) for id in ids]
```

Interestingly we are using match_torrent which supports any string, so it's an undocumented feature that you can move torrents by name too...

cas--:
You don't need to close this if you made a mistake, make the changes to your branch and force push it

mhertz:
Appreciate you Cas. I used my phone for this, never done before, and got confused, sorry. I thought to try revert branch back and make change again, so single commit for you to merge, but as you kindly said can just change/force-push, then did that. I would have amend'ed the last commit, but that was the "sync fork" commit I pressed on phone(thought would be same as rebase in cli), after accepting review, so hence made new commit. Sorry about that. I will look into the linting error later, and change if can. Thanks!

PR #451, opened 2024-05-12
[UI] Add flags, bytes uploaded/downloaded, RTT to Peers tab on webui

ppayne:
* add a flags string to each peer indicating the current state of the peer
* add the total bytes uploaded/downloaded by each peer
* add the estimated round-trip time to peers to which we connected

### DIFF
--- a/deluge/core/torrent.py
+++ b/deluge/core/torrent.py
@@ -115,6 +115,64 @@ def convert_lt_files(files):
     return filelist
 
 
+def peer_flags(peer):
+    """Returns a string representing flags about a peer.
+
+    Peer flags:
+        see also https://web.archive.org/web/20141111072948/http://www.utorrent.com/help/faq/misc#faq13
+        and http://libtorrent.org/reference-Core.html#peer_info
+        - S - snubbed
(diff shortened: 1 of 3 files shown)


ppayne:
Actually I created this patch 8 years ago (have had it in my own fork of this repository since 5 August 2016) and have found it invaluable - for some reason I like to know the details about the peers.

I've taken the latest pull from the `develop` branch, added this code, and tested it locally by building and installing and running.

ppayne:
I noticed <a href="https://dev.deluge-torrent.org/ticket/3499">#3499 Stop using deprecated LibTorrent features</a> in which it recommends using `torrent.status.flags && lt::torrent_flags::paused`.

So I'm wondering if I should try something like `peer.flags && lt::peer_flags::interesting` rather than `peer.flags && peer.interesting` (see <a href="https://www.libtorrent.org/reference-Core.html#peer_info">peer_info</a>).

ppayne:
Well I've tried every combo I can think of and none work, not `lt.peer_flags.`, not `lt.peer_flags_t.`, not `lt::peer_flags::`, not `lt::peer_flags_t::`.

PR #452, opened 2024-05-13
[WebUI]Fix Connection Manager stop daemon, return 5 elem from get_host_info()

ppayne:
Attempting to stop the daemon from Connection Manager in the WebUI would throw an error "An error occurred". On deeper inspection this was because this line threw a ValueError expecting 4 items but only getting 3:

```python
host, port, user, password = host[1:5]
```

This fix updates the `get_host_info()` function to also return password.

### DIFF
--- a/deluge/ui/hostlist.py
+++ b/deluge/ui/hostlist.py
@@ -163,12 +163,12 @@ class HostList:
             host_id (str): The host id to get info on.
 
         Returns:
-            list: A list of (host_id, hostname, port, username).
+            list: A list of (host_id, hostname, port, username, password).
 
         """
         for host_entry in self.config['hosts']:
             if host_entry[0] == host_id:
-                return host_entry[0:4]
+                return host_entry[0:5]
         else:
             return []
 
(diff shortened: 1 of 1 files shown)


ppayne:
Tested locally, now successfully stops and starts the daemon from Connection Manager.

deluge/ui/hostlist.py
```diff
@@ -163,12 +163,12 @@ def get_host_info(self, host_id):
             host_id (str): The host id to get info on.
 
         Returns:
-            list: A list of (host_id, hostname, port, username).
+            list: A list of (host_id, hostname, port, username, password).
```
cas--, line 166: It's a bit of a red flag to return a cleartext password to solve this issue. I think the correct solution is to use `hostlist.connect_host` in `json_api.stop_daemon`

cas-- (changes requested):

cas--:
Fixed via 1e5f248fb88804030df70cc511094888de4bd77f

PR #454, opened 2024-06-28
Ci fixes and updates

Lord-Kamina:
This branch updates the CI stack a bit in different ways.

1) It updates the workflow version of several actions that were crying due to Node 16 actions being deprecated.
2) It adds builds newer versions of python, so now it's built on 3.7, 3.9 and 3.10 (not using newer versions due to unavailability of newer libtorrent wheels on pypi)
3) Uses an updated gvsbuild-release for newer python versions.
4) Fixes console tests on `test_ui_entry.py` hanging, which allows us to unpin the pytest version used.

Notes:
- Point 3 above makes this PR dependent on https://github.com/deluge-torrent/gvsbuild-release/pull/3
- I updated libtorrent but we're still currently pinned to <2.0.9. This is because of point 2 above, as well as the fact that a change in libtorrent 2.0.9 makes one of the deluge tests fail (this should be addressed by fixing/removing that functionality or removing that test, I guess)


cas--:
I'm afraid there are too many changes here and not enough details in the commits about why the change is required. If possible, can you split these up into more concise PRs it will help reviewing and merging

Lord-Kamina:
> I'm afraid there are too many changes here and not enough details in the commits about why the change is required. If possible, can you split these up into more concise PRs it will help reviewing and merging

I hadn't seen this comment, sorry for not answering before.

I'd rather not split it (although it IS true that there are some not-strictly related things here) because then we'll never get it merged. However, I'm more than happy to amend commit messages and/or explain why I've made each change.

deluge/ui/gtk3/gtkui.py
```diff
@@ -21,13 +21,13 @@
...
 from gi.repository.Gtk import Builder, ResponseType
-from twisted.internet import defer, gtk3reactor
+from twisted.internet import defer, gireactor
 from twisted.internet.error import ReactorAlreadyInstalledError
 from twisted.internet.task import LoopingCall
 
 try:
     # Install twisted reactor, before any other modules import reactor.
-    reactor = gtk3reactor.install()
+    reactor = gireactor.install()
```
Lord-Kamina, line 30: This is just to avoid using a deprecated class name.

requirements-ci.txt
```diff
@@ -1,4 +1,4 @@
 -r requirements.txt
 -r requirements-tests.txt
-libtorrent==2.0.7
-pytest==7.4.2
+pytest
```
Lord-Kamina, line 3: Currently, pytest is pinned due to console tests failing, which is fixed by this PR.

requirements-ci.txt
```diff
@@ -1,4 +1,4 @@
...
+libtorrent<2.0.9
```
Lord-Kamina, line 4: libtorrent is pinned also due to a test failure, but the change was introduced in 2.0.9. We should eventually fix our test rather than keep pinning, but we can still update libtorrent even if just a bit.

deluge/tests/test_web_api.py
```diff
@@ -76,7 +76,7 @@ def test_set_config(self):
         assert config['pwd_sha1'] != web_config['pwd_sha1']
         assert config['sessions'] != web_config['sessions']
 
-    @defer.inlineCallbacks
+    @pytest_twisted.inlineCallbacks
```
Lord-Kamina, line 79: These changes are due to the pytest developers recommending that twisted not be used directly in tests with fixtures. From the pytest-twisted docs:

`Using twisted.internet.defer.inlineCallbacks as a decorator for test functions, which use fixtures, does not work. Please use pytest_twisted.inlineCallbacks instead.`

deluge/tests/test_ui_entry.py
```diff
@@ -391,19 +392,22 @@ async def test_console_command_status(self):
...
 
-    @defer.inlineCallbacks
+    @pytest_twisted.inlineCallbacks
     def test_console_command_config_set_download_location(self):
         fd = StringFileDescriptor(sys.stdout)
         self.patch_arg_command(['config --set download_location /downloads'])
         self.patch(sys, 'stdout', fd)
 
-        yield self.exec_command()
+        with mock.patch('deluge.ui.console.main.ConsoleUI.quit', autospec=True):
```
Lord-Kamina, line 408: This was calling the real `ConsoleUI.quit()`, which was stopping the reactor and thus making the tests bo boom.

deluge/core/alertmanager.py
```diff
@@ -174,7 +174,7 @@ def handle_alerts(self):
                 d = task.deferLater(reactor, 0, handler, alert)
                 on_handler_timeout = partial(
                     self.on_delayed_call_timeout,
-                    handler=handler.__qualname__,
+                    handler=handler.__class__.__qualname__,
```
Lord-Kamina, line 177: This flat-out produces an error, because __qualname__ is an attribute that classes (not their instances) have.

requirements-ci.txt
cas--: These are pinned to avoid breaking the CI over time so can be version bumped but must still be pinned

deluge/ui/gtk3/gtkui.py
cas--: We support a minimum version of Twisted and afaik this would break that so needs to be reverted, I don't think they are actually removing it anytime soon. We could add a Twisted version check for the import but would need a comment to explain why gireactor is being used since the Twisted docs are now contradictory about not using it for UI.

deluge/core/alertmanager.py
cas--: When does it raise an error? Handlers are functions or methods not class instances so `handle.__class__.__qualname__` does not make seem to make sense here?

requirements-ci.txt
Lord-Kamina: > These are pinned to avoid breaking the CI over time so can be version bumped but must still be pinned

For now, I'm happy to change it to `libtorrent==2.0.8`, but I think there's a discussion to be had about this (not in the context of this PR): I can understand the desire for the CI to not break easily, especially if you have little time as-is, but IMO a CI shouldn't easily break (if at all) due to a dependency publishing new bug-fix level releases. 

You're pinned to a two-year old version of what is quite probably your most important dependency. There has to be a better way to approach it (I have a few different ideas, but I don't really know where is the best place to start a discussion such as this?)

deluge/ui/gtk3/gtkui.py
Lord-Kamina: Does it though? the current requirement spec is: 
```twisted[tls]>=17.1; sys_platform != 'win32'
twisted[tls]<23,>=17.1; sys_platform == 'win32'```

And the Twisted api documentation for 16.4.1, a full major version earlier (https://twisted.org/documents/16.4.1/api/twisted.internet.gireactor.html), already has the `gireactor`, with an API that I think is identical?

deluge/core/alertmanager.py
Lord-Kamina: > When does it raise an error? Handlers are functions or methods not class instances so `handle.__class__.__qualname__` does not make seem to make sense here?

See this log, for an example:

https://github.com/Lord-Kamina/deluge/actions/runs/10050873417/job/27779525108

Lord-Kamina: Although now... reading that again, it might be an issue specific to running the tests.

Lord-Kamina:
@cas-- Using `libtorrent==2.0.8`fails because for some reason that version is missing from pip. I think that's why I had done `<2.0.9`

zakkarry:
> @cas-- Using `libtorrent==2.0.8`fails because for some reason that version is missing from pip. I think that's why I had done `<2.0.9`

The publishing to pip seems to be intermittent, I've made an issue on libtorrent a little over a week ago about pip builds, but I haven't seen any update. There is a open PR that seems to address some of the builds/CI for libtorrent, but I'm not sure if this directly address this concern.

You can see my issue and the linked PR [here](https://github.com/arvidn/libtorrent/issues/7729)

DjLegolas:
@zakkarry there were several issues with the `pypi` publish process for `libtorrent`...
i hope the fixes will work on the next release, as py3.13 is about to be released

@Lord-Kamina please revert the change to `Install GTK` step as it will always fail until we have a newer release for it.

zakkarry:
> @zakkarry there were several issues with the `pypi` publish process for `libtorrent`... i hope the fixes will work on the next release, as py3.13 is about to be released
> 
> @Lord-Kamina please revert the change to `Install GTK` step as it will always fail until we have a newer release for it.

As I've found out recently heh.

A user on the forums was asking about 3.13 support, and I told him to make an issue or PR, and he made https://github.com/deluge-torrent/deluge/pull/462 for it - it's failing some tests though it seems.

Lord-Kamina:
> > @cas-- Using `libtorrent==2.0.8`fails because for some reason that version is missing from pip. I think that's why I had done `<2.0.9`
> 
> 
> 
> The publishing to pip seems to be intermittent, I've made an issue on libtorrent a little over a week ago about pip builds, but I haven't seen any update. There is a open PR that seems to address some of the builds/CI for libtorrent, but I'm not sure if this directly address this concern.
> 
> 
> 
> You can see my issue and the linked PR [here](https://github.com/arvidn/libtorrent/issues/7729)

I know and I was inquiring about it as well a while ago. Thing is, even if that's fixed, I doubt we'll get retroactive releases for an old version, so the point stands in that we need to be able to 1) update libtorrent and 2) have at least some CI workflow that fails anf alerts us when newer versions break, because else we're putting ourselves in the same position the folks at twisted did, which is essentially forgetting it's an issue.

Of course, I'm willing to help however I can but I have essentially no experience working with libtorrent so I'm not sure that really amounts to much.

What do you think @DjLegolas?

deluge/core/alertmanager.py
Lord-Kamina: @cas-- I have reverted that change and implemented a proper fix in `conftest` now.

Lord-Kamina:
> > @zakkarry there were several issues with the `pypi` publish process for `libtorrent`... i hope the fixes will work on the next release, as py3.13 is about to be released
> > @Lord-Kamina please revert the change to `Install GTK` step as it will always fail until we have a newer release for it.
> 
> As I've found out recently heh.
> 
> A user on the forums was asking about 3.13 support, and I told him to make an issue or PR, and he made #462 for it - it's failing some tests though it seems.

I don't think that's at all related?
What was merged (and still is not really reflecting anywhere) is a change in the way gvsbuild is built for newer releases to be used in Deluge.

zakkarry:
I wasn't saying it was related necessarily, but you mentioned 3.13, so I informed you of a relevant PR to the version, that's all.

Lord-Kamina:
Closing this as per #466

PR #456, opened 2024-08-10
[Packaging] deluge-console Windows workaround

mhertz:
Deluge-console broke on Windows after commit d559f67, which before that was working because it's source files getting included with rest UI/plugin data copied into place. This workaround here, copies back it's source files into place. 

The underlying workarounded issues here is two-fold. It's not that anything is missing from the freezing, but rather that there at runtime is generated subparsers for the commands, using path_[0] (UI_PATH var in init.py, joined in console.py with rest path), so looking for files/paths not there anymore, as freezed now, and second, the use of 'overrides' decorator which breaks in freezed env since in decorators.py trying use inspect.stack()[2][4][0], and in freezed env inspect.stack()[2][4] is None, and hence any mention of former breaks with 'NoneType object is not subscriptable'. In non-freezed, it returns the call-context correctly, as source code readily available here.

### DIFF
--- a/packaging/win/delugewin.spec
+++ b/packaging/win/delugewin.spec
@@ -107,6 +107,8 @@ datas += copy_metadata('twisted', recursive=True)
 # Copy UI/Plugin and translation files to where pyinstaller expects
 package_data = collect_data_files('deluge')
 datas += package_data
+# Workaround for Console UI files not being included in package data
+datas += [('../../deluge/ui/console', 'deluge/ui/console')]
 
 icon = [src for src, dest in package_data if src.endswith('deluge.ico')][0]
 


mhertz:
To elaborate little, deluge-console broke on windows after commit d559f67, which before that, was working by happy coincidence from it's source files getting included with rest UI/plugin data copied into place. This fix, well workaround rather, copies back it's source files into place, to restore old ways, until proper fix can be made.

Anyway, I found the issue here is two-fold I believe. It's not that anything is missing from the freezing, but rather that there at runtime is generated subparsers for the commands, using __path___[0] (UI_PATH var in __init__.py, joined in console.py with rest path), so looking for files/paths not there anymore, as freezed now, and second , the use of @overrides which breaks in freezed env because of issues with inspect.stack() used in decorators.py - inspect issues is known with pyinstaller and a runtime hook for inspect already auto-added I see, but isn't enough apparently. 

Just though maybe little clarification warranted on this...

mhertz:
Thought better to wait for the good people able to actually fix this properly, instead of issuing interim fixes, hence closing this. 

Thank you :)

packaging/win/delugewin.spec
```diff
@@ -107,6 +107,7 @@ datas += copy_metadata('twisted', recursive=True)
 # Copy UI/Plugin and translation files to where pyinstaller expects
 package_data = collect_data_files('deluge')
 datas += package_data
+datas += [('../../deluge/ui/console', 'deluge/ui/console')]
```
cas--, line 110: Adding a quick comment above this would be nice

    # Workaround for Console UI files not being included in package data

mhertz:
Thanks buddy! :) 

I made a bug ticket before, describing first part i.e. subparsers generated for the console commands at run-time, but done with the path that would otherwise be available unfreezed, but pyinstaller has now packed the command py's with rest collected py's in one big archive of pyc's, without paths and seeks inside at will for file-data, so direct called command paths will not work when nothing there as a previous commit stopped copying the console files in place anymore : https://dev.deluge-torrent.org/ticket/3546

The last part is console-ui uses override decorator, which breaks in freezed env because decorators.py runs an inspect.stack command using call-context and source-code not available in freezed env - more info in second paragraph of commit message.

cas-- (approved):

PR #457, opened 2024-08-16
[Console] Block interactive-mode on Windows even with curses

mhertz:
Hangs after initial opening, so still needs looking into.


mhertz:
Nothing new btw, never worked on windows in interactive mode with windows-curses, so not me breaking this in the other PR fixing deluge-console again on windows.

PR #460, opened 2024-08-26
[UI][Common] Wrap torrent comment and tracker status URLs in HTML (clickable)

zakkarry:
This PR adds a function to common.py that uses a `re.sub` function to wrap supported URLs (taken from the is_url function) with HTML.

With this function, the torrent comments in `details_tab` and tracker status in `trackers_tab` will now display these URLs as clickable hyperlinks. 

This was previously only supported where the comment itself was only a URL. It will now support links within the comment. Also for tracker errors that give a link, in the case of removal and replacement for a new torrent on a tracker, the link reported in the status is now clickable too.

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -720,6 +720,16 @@ def parse_human_size(size):
     raise InvalidSize(msg % (size, tokens))
 
 
+def anchorify_urls(text: str) -> str:
+    """
+    Wrap all occurrences of text URLs with HTML
+    """
+    url_pattern = r'((htt)|(ft)|(ud))ps?://\S+'
+    html_href_pattern = r'<a href="\g<0>">\g<0></a>'
+
+    return re.sub(url_pattern, html_href_pattern, text)
+
+
 def is_url(url):
     """
     A simple test to check if the URL is valid
--- a/deluge/ui/gtk3/details_tab.py
+++ b/deluge/ui/gtk3/details_tab.py
@@ -10,7 +10,7 @@ import logging
 from xml.sax.saxutils import escape as xml_escape
 
 import deluge.component as component
-from deluge.common import decode_bytes, fdate, fsize, is_url
+from deluge.common import anchorify_urls, decode_bytes, fdate, fsize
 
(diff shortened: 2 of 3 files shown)


deluge/common.py
```diff
@@ -720,6 +720,23 @@ def parse_human_size(size):
     raise InvalidSize(msg % (size, tokens))
 
 
+def wrap_url_with_html(text):
```
cas--, line 723: Hmm I wonder if `anchorify_urls` might be a bit more succinct

deluge/common.py
```diff
@@ -720,6 +720,23 @@ def parse_human_size(size):
...
+    """
+    A simple regex sub to wrap all occurrences of URLs with HTML
```
cas--, line 725: Try to avoid describing the implementation detail, using regex, in docstring/comment. Also 'simple' is quite subjective so should be avoided.

deluge/common.py
```diff
@@ -720,6 +720,23 @@ def parse_human_size(size):
...
+
+    Args:
+        text (str): The string to replace urls in.
```
cas--, line 728: With Python 3 type hints can be moved to function definition

cas-- (changes requested):
This is a nice addition, just a few small changes

.gitignore
```diff
@@ -23,3 +23,4 @@ RELEASE-VERSION
 # used by setuptools to cache downloaded eggs
 /.eggs
 _pytest_temp/
+/.idea
```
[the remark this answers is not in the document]
zakkarry: Sure, I understand.

deluge/common.py
zakkarry: Sure, I'm just used to JSDocs in typescript and saw it used in other functions in common, so I just copy pasted it and did the small editing you see. 

Will move to type hints in the definition though.

deluge/common.py
zakkarry: sounds good to me

cas-- (approved):

PR #461, opened 2024-08-26
console-ui: endwin() not needed when using wrapper

Lord-Kamina:
Using it anyway produces a crash because it returns an error if called two times without any intervening updates.


cas--:
Thanks :+1:

PR #462, opened 2024-08-29
Replace deprecated cgi module with email

mtasaka:
As PEP 594 says, cgi module is marked as deprecated in python 3.11, and will be removed in 3.13
(actually removed at least in 3.13 rc1).

As suggested on PEP 594, replace cgi.parse_header
with email.message.Message .
email module is introduced in python 3.6 .

Ref: https://peps.python.org/pep-0594/#deprecated-modules
Ref: https://peps.python.org/pep-0594/#cgi

### DIFF
--- a/deluge/httpdownloader.py
+++ b/deluge/httpdownloader.py
@@ -6,7 +6,7 @@
 # See LICENSE for more details.
 #
 
-import cgi
+import email.message
 import logging
 import os.path
 import zlib
@@ -133,7 +133,9 @@ class HTTPDownloaderAgent:
                 content_disp = headers.getRawHeaders(b'content-disposition')[0].decode(
                     'utf-8'
                 )
-                content_disp_params = cgi.parse_header(content_disp)[1]
+                message = email.message.Message()
+                message['content-type'] = content_disp
+                content_disp_params = dict(message.get_params()[1:])
                 if 'filename' in content_disp_params:
                     new_file_name = content_disp_params['filename']
                     new_file_name = sanitise_filename(new_file_name)
@@ -152,7 +154,11 @@ class HTTPDownloaderAgent:
                     self.filename = new_file_name
 
             cont_type_header = headers.getRawHeaders(b'content-type')[0].decode()
-            cont_type, params = cgi.parse_header(cont_type_header)
+            message = email.message.Message()
+            message['content-type'] = cont_type_header
+            message_params = message.get_params()
+            cont_type = message_params[0][0]
+            params = dict(message_params[1:])
             # Only re-ecode text content types.
             encoding = None
             if cont_type.startswith('text/'):
--- a/deluge/httpdownloader.py
+++ b/deluge/httpdownloader.py
@@ -133,11 +133,10 @@ class HTTPDownloaderAgent:
                 content_disp = headers.getRawHeaders(b'content-disposition')[0].decode(
                     'utf-8'
                 )
-                message = email.message.Message()
-                message['content-type'] = content_disp
-                content_disp_params = dict(message.get_params()[1:])
-                if 'filename' in content_disp_params:
-                    new_file_name = content_disp_params['filename']
+                message = email.message.EmailMessage()
+                message['content-disposition'] = content_disp
+                new_file_name = message.get_filename()
+                if new_file_name:
                     new_file_name = sanitise_filename(new_file_name)
                     new_file_name = os.path.join(
                         os.path.split(self.filename)[0], new_file_name
@@ -154,11 +153,10 @@ class HTTPDownloaderAgent:
                     self.filename = new_file_name
 
             cont_type_header = headers.getRawHeaders(b'content-type')[0].decode()
-            message = email.message.Message()
+            message = email.message.EmailMessage()
             message['content-type'] = cont_type_header
-            message_params = message.get_params()
-            cont_type = message_params[0][0]
-            params = dict(message_params[1:])
+            cont_type = message.get_content_type()
+            params = message['content-type'].params
             # Only re-ecode text content types.
             encoding = None
             if cont_type.startswith('text/'):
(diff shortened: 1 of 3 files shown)


deluge/ui/web/json_api.py
```diff
@@ -191,7 +191,9 @@ def _on_json_request(self, request):
         Handler to take the json data as a string and pass it on to the
         _handle_request method for further processing.
         """
-        content_type, _ = cgi.parse_header(request.getHeader(b'content-type').decode())
+        message = email.message.Message()
+        message['content-type'] = request.getHeader(b'content-type').decode()
+        content_type = message.get_params()[0][0]
```
cas--, line 196: Using list indexes is rather opaque and indicates area for improvement. Did you consider using `get_content_type`?

deluge/httpdownloader.py
```diff
@@ -133,7 +133,9 @@ def request_callback(self, response):
                 content_disp = headers.getRawHeaders(b'content-disposition')[0].decode(
                     'utf-8'
                 )
-                content_disp_params = cgi.parse_header(content_disp)[1]
+                message = email.message.Message()
```
cas--, line 136: Why not use `EmailMessage` instead of legacy `Message` class?

deluge/httpdownloader.py
```diff
@@ -133,7 +133,9 @@ def request_callback(self, response):
...
+                message['content-type'] = content_disp
```
cas--, line 137: Why not use `content-disposition` and then `get_filename`

```
message['content-disposition'] = content_disp
new_file_name = message.get_filename()
if new_file_name:
    new_file_name = sanitise_filename(new_file_name)
```

deluge/httpdownloader.py
```diff
@@ -152,7 +154,11 @@ def request_callback(self, response):
                     self.filename = new_file_name
 
             cont_type_header = headers.getRawHeaders(b'content-type')[0].decode()
-            cont_type, params = cgi.parse_header(cont_type_header)
+            message = email.message.Message()
+            message['content-type'] = cont_type_header
+            message_params = message.get_params()
```
cas--, line 159: We can simplify with 
```
cont_type = message.get_content_type()
params = message['content-type'].params
```

cas-- (changes requested):
Thanks for the PR, just need some minor refactoring to simplify use of opaque list slice and indexes and use non-legacy `EmailMessage`

deluge/httpdownloader.py
mtasaka: Okay, thanks. will update.

deluge/httpdownloader.py
mtasaka: Thank you for comment. I did not notice this. Will reflect your review.

deluge/httpdownloader.py
mtasaka: Thank you . Actually https://docs.python.org/3/library/cgi.html#cgi.parse_header says your way is preferred. Will update.

mtasaka:
Thank you for review. I think I've reflected your comments.

cas-- (approved):
:+1:

mtasaka:
So:
`test_httpdownloader.py::TestDownloadFile::test_download_with_rename_sanitised` is failing..

I did some investigation on python3.12/email, and accoding to: 
https://www.rfc-editor.org/rfc/rfc2183#section-2.3
```
2.3 The Filename Parameter
Current RFC2045  grammar restricts parameter values (and hence
   Content-Disposition filenames) to US-ASCII
```
and https://www.rfc-editor.org/rfc/rfc2045 :
```
     parameter := attribute "=" value

     attribute := token
                  ; Matching of attributes
                  ; is ALWAYS case-insensitive.

     value := token / quoted-string

     token := 1*<any (US-ASCII) CHAR except SPACE, CTLs,
                 or tspecials>

     tspecials :=  "(" / ")" / "<" / ">" / "@" /
                   "," / ";" / ":" / "\" / <">
                   "/" / "[" / "]" / "?" / "="
                   ; Must be in quoted-string,
                   ; to use within parameter values```

So python `email.message.EmailMessage()` assumes  `filename=/etc/passwd` as invalid and rejects this to parameter.

As RFC2045 says quoted value is valid, and actually if I change

```
diff --git a/deluge/tests/test_httpdownloader.py b/deluge/tests/test_httpdownloader.py
index 1c2704560..0a4695b5c 100644
--- a/deluge/tests/test_httpdownloader.py
+++ b/deluge/tests/test_httpdownloader.py
@@ -206,10 +206,10 @@ async def test_download_with_rename_exists(self):
         self.assert_contains(filename, 'This file should be called renamed')
 
     async def test_download_with_rename_sanitised(self):
-        url = self.get_url('rename?filename=/etc/passwd')
+        url = self.get_url('rename?filename="/etc/passwd"')
         filename = await download_file(url, fname('original'))
         assert filename == fname('passwd')
-        self.assert_contains(filename, 'This file should be called /etc/passwd')
+        self.assert_contains(filename, 'This file should be called "/etc/passwd"')
 
     async def test_download_with_attachment_no_filename(self):
         url = self.get_url('attachment')
```

Then test_download_with_rename_sanitised test passes. Is this okay?

cas--:
Ah good to find the test is exercising that code. Yeah I agree that the quotes are optional unless using special characters so that change makes sense

mtasaka:
> Ah good to find the test is exercising that code. Yeah I agree that the quotes are optional unless using special characters so that change makes sense

Okay, thank you for confirmation. I've updated the testcase.

mtasaka:
Oops...  I've modified the last commit a bit (quoting the filename should be in 2 places)

mtasaka:
Now CI seems good, except that CI for windows with python3.7 is failing, but looking at the log, I don't think the failure is related to this PR.

PR #463, opened 2024-08-30
Fix bug in AuthenticationDialog

Jaywalker:
Specific bug is described here: https://dev.deluge-torrent.org/ticket/3643#ticket

The issue occured because the `AuthenticationDialog` was `destroy()`'d before the calls were made to get the text from their `GTK.Entry()`'s. It has been fixed by using the same method as `AccountDialog` and storing the information in an `Account` object prior to the `destroy()` call.

It may be worth considering giving the user an option to save/update the `hostlist.conf`  file with the working credentials as well, but that would merit discussion and a separate PR. This PR simply fixes the problem and leaves existing logic flow in place.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,6 +23,7 @@ Contributors (and Past Developers):
     * Pedro Algarvio ('s0undt3ch') <ufs@ufsoft.org>
     * Cristian Greco ('cgreco') <cristian@regolo.cc>
     * Chase Sterling ('gazpachoKing') <chase.sterling@gmail.com>
+    * Justin Williams ('Jaywalker')
 
 Plugin Developers:
     * Autoadd       : Chase Sterling
--- a/deluge/ui/gtk3/connectionmanager.py
+++ b/deluge/ui/gtk3/connectionmanager.py
@@ -316,7 +316,7 @@ class ConnectionManager(component.Component):
 
             def dialog_finished(response_id):
                 if response_id == Gtk.ResponseType.OK:
-                    self._connect(host_id, dialog.get_username(), dialog.get_password())
+                    self._connect(host_id, dialog.account.username, dialog.account.password)
 
             return dialog.run().addCallback(dialog_finished)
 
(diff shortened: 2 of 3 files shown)


cas-- (changes requested):
Thanks for this, I imagine this will also affect PasswordDialog so would be useful to fix there too

AUTHORS
```diff
@@ -23,6 +23,7 @@ Contributors (and Past Developers):
     * Pedro Algarvio ('s0undt3ch') <ufs@ufsoft.org>
     * Cristian Greco ('cgreco') <cristian@regolo.cc>
     * Chase Sterling ('gazpachoKing') <chase.sterling@gmail.com>
+    * Justin Williams ('Jaywalker')
```
cas--, line 26: This list is really for authors of significant contributions to Deluge

deluge/ui/gtk3/connectionmanager.py
```diff
@@ -316,7 +316,7 @@ def _on_connect_fail(self, reason, host_id, try_counter):
 
             def dialog_finished(response_id):
                 if response_id == Gtk.ResponseType.OK:
-                    self._connect(host_id, dialog.get_username(), dialog.get_password())
+                    self._connect(host_id, dialog.account.username, dialog.account.password)
```
zakkarry, line 319:
```suggestion
                    self._connect(
                        host_id, dialog.account.username, dialog.account.password
                    )
```

deluge/ui/gtk3/dialogs.py
```diff
@@ -245,6 +247,16 @@ def get_password(self):
     def on_password_activate(self, widget):
         self.response(Gtk.ResponseType.OK)
 
+    def _on_response(self, widget, response):
+        if response == Gtk.ResponseType.OK:
+            self.account = Account(
+                self.username_entry.get_text(),
+                self.password_entry.get_text(),
+                "",
```
zakkarry, line 255:
```suggestion
                '',
```

zakkarry (changes requested):
Two linting issues by the way, figured I'd point them out.

cas--:
Forgot about this PR and actually fixed it separately while resolving same issue with tray password: https://github.com/deluge-torrent/deluge/commit/88e6bbc1335c8d4c8ac2ee2152f3d23e931034f1

PR #464, opened 2024-09-03
[Console] Fix 'rm' command hanging when done

mhertz:
Commander.exec_command was await'ing returned deferreds as per commit 253eb22.


zakkarry:
I love seeing you active here bro! <3

cas-- (approved):

PR #466, opened 2024-09-18
Update GTK stack for use in CI/CD

Lord-Kamina:
Ok @cas--, you win; I'll split it.

I also opted to make a new PR  that will supersede #454, to make it easier to keep track of the discussion, were there to be any.

After this, I will make another PR for unpinning pytest (I _might_ include the change from `gtkreactor` to `gireactor` in this latter one because I genuinely think it would make everybody's lives easier, but you tell me)

This is just the bit updating the GTK stack. As you can see, I've reverted the changes to libtorrent and such (although I maintain it should be somewhat of a priority to address this sooner rather than later)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,6 +16,9 @@ on:
       ref:
         description: "Enter a tag or commit to package"
         default: ""
+      gvsbuild-tag:
+        description: "Use an alternative gvsbuild release for the windows build. Defaults to latest."
+        default: "latest"
 
 jobs:
   windows_package:
@@ -23,8 +26,8 @@ jobs:
     if: (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'package'))
     strategy:
       matrix:
-        arch: [x64, x86]
-        python: ["3.9"]
+        arch: [x64]
+        python: ["3.10"]
         libtorrent: [2.0.7, 1.2.19]
 
     steps:
@@ -49,12 +52,20 @@ jobs:
           cache: pip
 
       - name: Prepare pip
-        run: python -m pip install wheel setuptools==68.*
+        run: python -m pip install wheel setuptools==70.*
+
+      - name: Determine gvsbuild release URL
+        id: gvsbuild-url
+        shell: bash
+        run: |
+          test -z "${{ github.event.inputs.gvsbuild-tag }}" && tag=latest || tag="${{ github.event.inputs.gvsbuild-tag }}"
+          if [[ "$tag" == "latest" ]]; then URL="https://github.com/${{ github.repository_owner }}/gvsbuild-release/releases/$tag/download"; else URL="https://github.com/${{ github.repository_owner }}/gvsbuild-release/releases/download/$tag" ; fi
+          echo "gvsbuild-release-url=$URL" >> $GITHUB_OUTPUT
 
       - name: Install GTK
         run: |
           $WebClient = New-Object System.Net.WebClient
-          $WebClient.DownloadFile("https://github.com/deluge-torrent/gvsbuild-release/releases/download/latest/gvsbuild-py${{ matrix.python }}-vs16-${{ matrix.arch }}.zip","C:\GTK.zip")
+          $WebClient.DownloadFile("${{ steps.gvsbuild-url.outputs.gvsbuild-release-url }}/gvsbuild-py${{ matrix.python }}-vs17-${{ matrix.arch }}.zip","C:\GTK.zip")
           7z x C:\GTK.zip -oc:\GTK
           echo "C:\GTK\release\lib" | Out-File -FilePath $env:GITHUB_PATH -Append
           echo "C:\GTK\release\bin" | Out-File -FilePath $env:GITHUB_PATH -Append
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,13 +10,19 @@ on:
       core-dump:
         description: "Set to 1 to enable retrieving core dump from crashes"
         default: "0"
+      gvsbuild-tag:
+        description: "Use an alternative gvsbuild release for the windows build. Defaults to latest."
+        default: "latest"
 jobs:
   test-linux:
-    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
-
+        python-version: ["3.9", "3.10"]
+        os: ["ubuntu-24.04"]
+        include:
+          - os: ubuntu-22.04
+            python-version: 3.7
+    runs-on: ${{ matrix.os }}
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v4
@@ -74,7 +80,8 @@ jobs:
     runs-on: windows-2022
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
+        python-version: ["3.7", "3.9", "3.10"]
+        arch: ["x64"]
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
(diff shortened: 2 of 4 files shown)


Lord-Kamina:
@cas-- anything that needs to change here before this can be merged? (The CI fails should be resolved by triggering a manual run at the `gvsbuild-release` repo)

.github/workflows/ci.yml
```diff
@@ -36,6 +42,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          sudo apt-get install libcairo2-dev libgirepository1.0-dev
```
cas--, line 45: This doesn't seem right to be installing these dev dependencies, everything should have been built already?

.github/workflows/ci.yml
```diff
@@ -74,7 +81,8 @@ jobs:
     runs-on: windows-2022
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
+        python-version: ["3.7", "3.9", "3.10"]
```
cas--, line 84: Why are we including 3.9 here and removing it from package build?

.github/workflows/cd.yml
```diff
@@ -77,7 +88,6 @@ jobs:
         working-directory: deluge_src
         run: |
           python -m pip install .
-          python setup.py install_scripts
```
cas--, line 80: Are you sure removing this is correct?

.github/workflows/cd.yml
```diff
@@ -24,8 +27,8 @@ jobs:
     strategy:
       matrix:
         arch: [x64, x86]
-        python: ["3.9"]
-        libtorrent: [2.0.7, 1.2.19]
+        python: ["3.7", "3.10"]
```
cas--, line 30: Why do we need 3.7 build?

.github/workflows/ci.yml
```diff
@@ -10,13 +10,19 @@ on:
...
     strategy:
       matrix:
-        python-version: ["3.7", "3.10"]
-
+        python-version: ["3.9", "3.10"]
+        os: ["ubuntu-24.04"]
+        include:
+          - os: ubuntu-22.04
+            python-version: 3.7
+    runs-on: ${{ matrix.os }}
```
cas--, line 25: Let's keep these Linux CI changes separate from the Windows GTK changes

.github/workflows/ci.yml
```diff
@@ -89,6 +97,24 @@ jobs:
           cache: "pip"
           cache-dependency-path: "requirements*.txt"
 
+      - name: Determine gvsbuild release URL
```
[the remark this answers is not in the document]
Lord-Kamina: The idea is to enable the gtk tests though. I think that might have gotten cut onto another PR though.

.github/workflows/docs.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: |
           pip install --upgrade pip wheel
           pip install tox
-          sudo apt-get install enchant-2
+          sudo apt-get install enchant-2 libgirepository1.0-dev
```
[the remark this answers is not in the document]
Lord-Kamina: I honestly can't remember 100% but it is possible that it failed somewhere without it. Would have to look it up.

requirements-tests.txt
```diff
@@ -1,4 +1,4 @@
-libtorrent
+libtorrent==2.0.7
```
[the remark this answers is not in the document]
Lord-Kamina: Yes because things in newer libtorrent made tests fail, so until tests are updated, it has to be.

.github/workflows/cd.yml
```diff
@@ -24,8 +27,8 @@ jobs:
...
+        libtorrent: [2.0.8, 1.2.19]
```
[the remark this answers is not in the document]
Lord-Kamina: Actually I think this was a typo or possibly I missed this when reverting. I can't find it now so it should go back to 2.0.7 (for now)

.github/workflows/cd.yml
Lord-Kamina: We don't really but I had kept it because it's what deluge was using and/or officially targetting at the moment.

The point here was to have a 3.7 build to represent not breaking the current build, alongside building with newer versions.

cas--: The actual release only uses one version. There is a difference between CI tests that need to both test oldest and latest support dependencies versus this workflow that creates the Windows release package which should aim to be most stable recent release of dependencies.

Lord-Kamina: I just noticed that the packaging test was not running at all on my fork (probably because I never had an event there that would trigger it) so I'll need a bit to check that everything is actually ok on that front.

I'm in the process of looking into some of your other comments and corrections. Also, please take a look at what I said you over on discord.

.github/workflows/ci.yml
Lord-Kamina: That'll just prolong this whole thing. The linux changes are just: keep 22.04 (which needs an older python) and add a build on a newer version of ubuntu.

.github/workflows/cd.yml
Lord-Kamina: I'm leaving just x64-py3.10 but I put LT1.2 back in due to the outburst in Discord.

.github/workflows/cd.yml
Lord-Kamina: Nope. I think it might be with something I did on another branch but I'll cross that bridge when I get there.
Reverted for now.

.github/workflows/ci.yml
Lord-Kamina: I'm including it due to some differences in the importlib module. It's not relevant now but I think it will be to test against with some PR I've yet to submit.

.github/workflows/ci.yml
Lord-Kamina: I removed it for now.

.github/workflows/cd.yml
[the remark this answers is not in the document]
Lord-Kamina: Changed back to 2.0.7

.github/workflows/docs.yml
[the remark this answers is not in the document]
Lord-Kamina: Removed as well.

Lord-Kamina:
BTW @cas--  I have pinned twisted here because of the issue with the maybe_coroutine test, until we can figure out what changed/is going on.
Other than that, I think most comments have been addressed.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 libtorrent
-twisted[tls]>=17.1
+twisted[tls]>=17.1,<=24.7
```
cas--, line 2: I have resolved the problem with Twisted 24.10 so this can be removed

.github/workflows/cd.yml
```diff
@@ -49,12 +52,20 @@ jobs:
...
       - name: Prepare pip
-        run: python -m pip install wheel setuptools==68.*
+        run: python -m pip install wheel setuptools==70.*
+
+      - name: Determine gvsbuild release URL
+        id: gvsbuild-url
+        shell: bash
+        run: |
+          test -z "${{ github.event.inputs.gvsbuild-tag }}" && tag=latest || tag="${{ github.event.inputs.gvsbuild-tag }}"
+          if [[ "$tag" == "latest" ]]; then URL="https://github.com/${{ github.repository_owner }}/gvsbuild-release/releases/$tag/download"; else URL="https://github.com/${{ github.repository_owner }}/gvsbuild-release/releases/download/$tag" ; fi
```
cas--, line 62: This should be able to split over multiple lines to make it more readable?

.github/workflows/ci.yml
cas--: Can we keep the scope of this PR to Win GTK only please

.github/workflows/ci.yml
cas--: Let's leave that to the other PR then

.github/workflows/ci.yml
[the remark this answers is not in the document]
cas--: If we are not using it let's leave it out of this PR, the code is always going to be in CD.yml. Also it doesn't address the Linux GTK tests

requirements-tests.txt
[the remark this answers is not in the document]
cas--: So for CI we already have this pinned so this only affects local testing, perhaps there shouldn't be a difference but that is also a question about pinning the entire test/dev stack to known working dependencies. I really do want to fix the tests for libtorrent > 2.0.7 so keeping it unpinned serves as a reminder

cas-- (changes requested):
Can we please keep PRs to a single feature/fix as outlined here: https://testing.googleblog.com/2024/07/in-praise-of-small-pull-requests.html?m=1

Lord-Kamina:
> Can we please keep PRs to a single feature/fix as outlined here: https://testing.googleblog.com/2024/07/in-praise-of-small-pull-requests.html?m=1

So here's my problem with this...

I understand the concept and motivation for wanting small PRs, and IMO I've been nothing if not accommodating. But other issues with it aside, this "nothing" PR has been open for months. This is not a dig on you or anyone else, because I know everyone working on this project does so as a volunteer freely contributing their time.

My point is this: These commits were meant as the first-step in getting a more modern CI pipeline as the foundation of other rather large and potentially breaking changes to deluge that require using newer versions of python (this is, dropping `pkg_resources` entirely and moving the plug-in system to using wheels and importlib).
That code is also already written and tested (rebasing, cleaning-up and addressing other issues aside).

And, that itself is just a preparatory step for something I've not yet done, which is make a better and easily-reproducible process for creating macOS packages and eventually, adding macOS to the CD.

With this in mind, and considering the snail pace at which this project moves (again, not a dig, but rather a fact), at some point you maybe gotta start wondering why don't you make it easier for everyone, yourself included? This was originally a much larger PR that involved fixing tests and also updating the python runtimes, which is why the there is an update on the linux workflow as well. I already split it and excluded the fixes to the testing suite, but at this point it's gone well beyond "small PRs are good". The change to the linux workflow is absolutely immaterial to the working of deluge or the CI, and is directly related to the intent of this PR.

It's incredibly frustrating because it seems like we're gettint fixated with dividing what should have been a simple task into an ever-increasing number of steps ad nauseam.

So, to recap: I understand and fully support the spirit of advocating for small PRs, but I think you're being maybe a tad excessive on this matter.

PR #467, opened 2024-09-20
[3350][GTK] Fix a regression with quick search

DjLegolas:
When one tries to search in the TorrentView, then a little search window appears but does nothing.

Closes: https://dev.deluge-torrent.org/ticket/3350

### DIFF
--- a/deluge/ui/gtk3/listview.py
+++ b/deluge/ui/gtk3/listview.py
@@ -782,7 +782,7 @@ class ListView:
 
         return True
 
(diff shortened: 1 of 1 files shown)


cas--:
Nice catch

For reference it's useful to include a link to gtk docs so we know what the function signature looks like: https://lazka.github.io/pgi-docs/Gtk-3.0/callbacks.html#Gtk.TreeViewSearchEqualFunc

cas-- (approved):

DjLegolas:
Added reference in commit message 👍

PR #468, opened 2024-10-28
[Build] Prevent usage of twisted>=24

mhertz:
Breaking ability of adding multiple torrents at once in WebUI, with following error in web-console after pressing 'Add':
```
deluge-all.js:7 Uncaught TypeError: Cannot read properties of undefined (reading 'filename')
    at S.getFilename (deluge-all.js:7:1946)
    at S.<anonymous> (deluge-all.js:3:2013)
    at Ext.util.MixedCollection.each (ext-all.js:21:111850)
    at constructor.each (ext-all.js:21:321534)
    at S.onAddClick (deluge-all.js:3:1932)
    at S.onClick (ext-all.js:21:426947)
    at HTMLTableElement.I (ext-all.js:21:57750)
```
The issue stems from only one of the torrents getting uploaded to server, after cgi.parse_multipart was removed from twisted.web.http and replaced with email.message_from_bytes, in there commit [4579398](https://github.com/twisted/twisted/commit/4579398f6b089f93181ba2f912e2524bd1f43c5e). 

Note, there's currently a twisted draft [PR](https://github.com/twisted/twisted/pull/12352) replacing email.message* with 'multipart' lib, as seemingly better fitted/faster, and which coincidentally also is fixing our issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 libtorrent
-twisted[tls]>=17.1
+twisted[tls]<24,>=17.1
 twisted[tls]<23,>=17.1; sys_platform == 'win32'
 rencode
 pyopenssl


cas--:
Can you provide an example of the problem? Why can't we handle this in the web server code?

cas--:
Okay I can see the problem now. The content-disposition of the file upload form is using the same name for both files. 

Sample snippet of request
```
Content-Disposition: form-data; name="file"; filename="test1.torrent"
Content-Type: application/x-bittorrent
...
Content-Disposition: form-data; name="file"; filename="tes2.torrent"
Content-Type: application/x-bittorrent
```

From [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#name) there is a suggestion that this is a valid when using the `multiple` attribute:

> When dealing with multiple files in the same field (for example, the [multiple](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#multiple) attribute of an [<input type="file">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) element), there can be several subparts with the same name.

I think the only fix is to override the `_getMultiPartArgs` with our own version:

```diff
+from collections import defaultdict
 import fnmatch
 import json
 import logging
@@ -40,7 +41,7 @@ def _getMultiPartArgs(content: bytes, ctype: bytes) -> dict[bytes, list[bytes]]:
     """
     Parse the content of a multipart/form-data request.
     """
-    result = {}
+    result = defaultdict(list)
     multiPartHeaders = b'MIME-Version: 1.0\r\n' + b'Content-Type: ' + ctype + b'\r\n'
     msg = message_from_bytes(multiPartHeaders + content)
 
@@ -56,7 +57,7 @@ def _getMultiPartArgs(content: bytes, ctype: bytes) -> dict[bytes, list[bytes]]:
             continue
 
         payload: bytes = part.get_payload(decode=True)  # type:ignore[assignment]
-        result[name.encode('utf8')] = [payload]
+        result[name.encode('utf8')].append(payload)
     return result
```

mhertz:
Thank you my friend cas, appreciate you :)

PR #469, opened 2024-11-02
[UI] Split trackers by newline, not by "/n"

garret1317:
Fixes a bug where e.g. http://not-a-real-tracker.com was split into 
```
http:/
ot-a-real-tracker.com
```
when the add trackers dialogue was next shown

First adding the tracker (typing/pasting it in):
Screenshot from 2024-11-02 17-25-20
Next time adding trackers (autofilled):
Screenshot from 2024-11-02 17-25-27


DjLegolas:
Nice catch.
I think it will be better to use `splitlines` here as it will take care of all possible options.

cas-- (approved):

cas--:
I just realised it was a typo by myself during refactoring, oops!

garret1317:
:+1: thank you

PR #471, opened 2025-01-04
Update Copyright Year

LeoColman:


LeoColman:
- Update copyright string and ran "generate pot files"

PR #473, opened 2025-01-16
Add Support for "Nice Names" for Windows Network Interfaces

doadin:


doadin:
@cas-- Anything I can do to make this acceptable?

doadin:
@cas-- I know the same comment I made before probabbly still stands but something I also thought about before and is maybe a change for a different PR. Currently deluge falls back to a previous config value for network interface if it has issues. I think instead to avoid IP leaks etc it should default to 127.0.0.1 to prevent connections.

PR #477, opened 2025-04-10
[Core] Respect listen_reuse_port to randomize port on startup

GZGavinZhao:
The `listen_reuse_port` config is never respected in deluge.
Practically, this means that users cannot randomize the listen port on
startup. Before this change, this is what happens in `_set_listen_on`:

1. On run 1, `listen_random_port` is empty.
2. A random port is chosen, and its value is written into
   `listen_random_port` in `core.conf`.
3. On all subsequent runs, `deluged` reads `core.conf` and picks up
   `listen_random_port` from the previous run, so `listen_random_port`
   will never be randomized again.

To fix this, we should read `listen_reuse_port` and if it's `false`,
always disregard the current `listen_random_port` and randomly
pick a new value for it.

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -200,7 +200,7 @@ class PreferencesManager(component.Component):
     def __set_listen_on(self):
         """Set the ports and interface address to listen for incoming connections on."""
         if self.config['random_port']:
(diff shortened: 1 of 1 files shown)


cas-- (approved):
Good catch :+1:

PR #478, opened 2025-04-30
Docker support for development

mth4saurabh:
I was trying to setup development env and it was easier to do so with docker/compose, so I thought it might be useful for others as well

### Changes

- Add Dockerfile and docker-compose.yml with dev/test services
- Update deluge development documentation

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,53 @@
+# Do not include git in dockerignore since we need it for pre-commit
+
+# Python
+__pycache__/
+*.py[cod]
+*$py.class
+*.so
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# Virtual Environment
+.env
+.venv
+env/
+venv/
+ENV/
+
+# IDE
+.idea/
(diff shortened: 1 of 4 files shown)


kenstir:
Is this not a duplicate of #476 ?

mth4saurabh:
@kenstir I have added general docker setup along with dev documentation that should be helpful to anyone using docker/compose, I believe #476 is limited to vs code?

cas--:
Yes devcontainers originates from VSCode but it doesn't require VSCode, there are other tools that [support it](https://containers.dev/supporting) but it also can be run as regular docker, even using docker compose: https://containers.dev/guide/dockerfile

I usually avoid being biased to specific code editors but I like this concept, it has no lock-in and might actually be useful in CI pipeline. It would be good if you could work with @kenstir on his PR to create good developer docker option in devcontainer format.

mth4saurabh:
Sounds good! @kenstir I think we can update dev documentation and have a quick look at what's missing in #476 (w.r.t. to Calum's proposal)

kenstir:
@mth4saurabh I like the addition of the documentation.  Can you add your documentation on top of #476 ?  For VS Code, the instructions are simple:
```
code .
```
and VS Code prompts you to "reopen in container".

For command-line use, the instructions are also straightforward:
```
cd .devcontainer
docker build -t deluge-dev .
docker run -it deluge-dev bash
```

PR #479, opened 2025-05-02
Perform force recheck on torrent in error state after restart

mth4saurabh:
Should resolve [#3230](https://dev.deluge-torrent.org/ticket/3230)

Automatically recover torrents from error states (e.g., after crashes) without manual intervention.

#### Changes:
* Added auto-recheck for error-state torrents in `load_state()`
* Added test verifying auto-recheck after shutdown

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -884,6 +884,13 @@ class TorrentManager(component.Component):
                 len(state.torrents),
                 str(datetime.datetime.now() - start),
             )
+
+            # After all torrents are loaded, perform force recheck for torrents in error state
+            for torrent in self.torrents.values():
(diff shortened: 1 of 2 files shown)


cas--:
My quick response here is that setting the torrents in error state is by intention to specifically avoid that recheck and losing torrent state e.g. storage not mounted. There will be a Trac issues that going into the background of implementation in more detail

I am aware there is a need to better recover torrents that are in error state but I don't think this PR is the answer and needs more thought and planning about how that should happen.

mth4saurabh:
Thanks for the feedback! I did try to find other connected components but looks like I might need to spend more time looking at what else goes on in the background to build on the _naive_ implementation here.

PR #481, opened 2025-05-19
Fix multipart form data parsing for file upload in Web UI

qm3jp:
Per the [forum thread](https://forum.deluge-torrent.org/viewtopic.php?t=56916), adding multiple files via the Web UI was broken by the removal of `parse_multipart` in Twisted v24.

I've pulled in [python-multipart](https://github.com/kludex/python-multipart) to re-enable that functionality.

### DIFF
--- a/deluge/ui/web/server.py
+++ b/deluge/ui/web/server.py
@@ -14,6 +14,7 @@ import os
 import tempfile
 from pathlib import Path
 
+from python_multipart import parse_form
 from twisted.application import internet, service
 from twisted.internet import defer, reactor
 from twisted.web import http, resource, server, static
@@ -126,7 +127,7 @@ class Upload(resource.Resource):
             request.finish()
             return server.NOT_DONE_YET
 
-        files = request.args.get(b'file', [])
+        files = self._parse_files(request)
         filenames = []
 
         if files:
@@ -144,6 +145,28 @@ class Upload(resource.Resource):
         request.setHeader(b'content-type', b'text/html')
         request.setResponseCode(http.OK)
         return json.dumps({'success': bool(filenames), 'files': filenames}).encode()
+
+    def _parse_files(self, request):
+        """
+        Parse multipart/form-data using python-multipart,
+        and return an array of bytestrings.
+        """
+        files = []
+
+        def on_file(file):
+            # Convert from python-multipart's internal File class to simmple bytestrings.
+            file.file_object.seek(0)
+            file_bytes = file.file_object.read()
+            files.append(file_bytes)
+
+        # Convert headers to dict of ascii strings for python-multipart.
+        headers_dict = dict(request.requestHeaders.getAllRawHeaders())
+        headers_ascii = {
+            k.decode('ascii'): v[0].decode('ascii') for k, v in headers_dict.items()
+        }
+        parse_form(headers_ascii, request.content, None, on_file)
+
+        return files
 
 
 class Render(resource.Resource):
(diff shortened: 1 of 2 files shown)


cas--:
Thanks for providing this fix but this will be fixed in next release of Twisted 25.5, relevant links below:

* https://github.com/twisted/twisted/pull/12449
* https://github.com/twisted/twisted/issues/12423

Do you think your fix is still required? I would note that it would be preferable to override Twisted method than include a third-party lib to fix the issue.

qm3jp:
Oh whoops, I hadn't seen that Twisted was fixing it on their end.
How long do you think it might be until this new Twisted release makes it into a Deluge release?
If it's not going to be for a while, might it be worth merging this one now, and then remove it later was the Twisted update is incorporated?

Gauvino:
Hello, new twisted has been merged, i think it can be fixed now

Gauvino:
Can this be look at another time cause it's really annoying to not being capable of uploading multiple torrent please !

cas--:
@Gauvino There is nothing to fix in Deluge, either install Twisted 23.10 or 25.05 for deluge web.

Roxedus:
@cas-- However, a normal pip install will not pull a working Twisted with the current requirements file.

dustin-cowles:
@cas-- This is still an issue for downstream distributions (e.g. the Docker image from linuxserver.io) since they are doing a `pip install` with the requirements.txt from the repo which results in a broken version of twisted being installed. Is there any chance of getting the requirements.txt updated to require twisted 25.x?

https://github.com/linuxserver/docker-deluge/issues/226

cas--:
I can absolutely say that using this projects `requirements.txt` would pick up latest versions of twisted:

```
uv venv
uv pip install -r https://raw.githubusercontent.com/deluge-torrent/deluge/refs/heads/develop/requirements.txt | grep -i twisted
Resolved 29 packages in 7ms
Installed 29 packages in 9ms
...
 + twisted==25.5.0
... 
```

However looking at the Linuxserver Dockerfile we can see that it's using apk to install Deluge and dependencies:

```
docker run --rm -it --entrypoint "" linuxserver/deluge bash
root@a117751c0c54:/# python -c "import twisted; print(twisted.__version__)"
24.11.0
root@a117751c0c54:/# python -m pip
/usr/bin/python: No module named pip
root@a117751c0c54:/# apk list | grep twisted
...
py3-twisted-24.11.0-r0 x86_64 {py3-twisted} (MIT) [installed]
...
```

Can also verify that Alpine linux are packaging an older version Twisted: https://pkgs.alpinelinux.org/package/edge/community/x86/py3-twisted

There are different options available to the downstream maintainers

1. Request apk package manager version bump Twisted to 25.05 e.g. [upgrade to 24.11.0](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/81524)
2. If problems upgrading to 25.05 instead provide a bug fix patch to be applied to py3-twisted 24.11.0 
3. Create a bug fix patch to be applied in Dockerfile to Twisted site-packages files, it's [only a few lines in one file](https://github.com/twisted/twisted/pull/12424/files#diff-3093a8f64dec64c2305f6322f276a5cee1437f0037efc660ce6524ffa58017a1)
4. Use pip to install latest twisted version (caveat it's usually not advised to mix python dependencies installed by a package manager and pip)

The first two would benefit the wider community so would be the better route.

I wish I could help more with this but it really is the role of those packaging Deluge to manage dependency versions correctly.

PR #483, opened 2025-05-29
[Core] Work around libtorrent size limit

marcan:
When loading a torrent_info by file path, libtorrent has a hardcoded size limit that some torrents exceed. This means that the torrent can be added normally to deluge, but on restart it fails to load the file from the state directory and the torrent is lost.

Work around the hardcoded buffer size limit by reading in the file directly and passing the buffer.

See: https://libtorrent-discuss.narkive.com/wYACQDb7/libtorrent-loading-a-big-torrent-fails
Fixes: https://forum.deluge-torrent.org/viewtopic.php?t=56856

### DIFF
--- a/deluge/core/torrentmanager.py
+++ b/deluge/core/torrentmanager.py
@@ -15,6 +15,7 @@ import os
 import pickle
 import time
 from base64 import b64encode
+from pathlib import Path
 from tempfile import gettempdir
 from typing import Dict, List, NamedTuple, Tuple
 
@@ -338,8 +339,8 @@ class TorrentManager(component.Component):
         if log.isEnabledFor(logging.DEBUG):
             log.debug('Attempting to extract torrent_info from %s', filepath)
         try:
-            torrent_info = lt.torrent_info(filepath)
-        except RuntimeError as ex:
+            torrent_info = lt.torrent_info(Path(filepath).read_bytes())
+        except (RuntimeError, OSError) as ex:
             log.warning('Unable to open torrent file %s: %s', filepath, ex)
         else:
             return torrent_info


deluge/core/torrentmanager.py
```diff
@@ -338,8 +338,8 @@ def get_torrent_info_from_file(self, filepath):
         if log.isEnabledFor(logging.DEBUG):
             log.debug('Attempting to extract torrent_info from %s', filepath)
         try:
-            torrent_info = lt.torrent_info(filepath)
-        except RuntimeError as ex:
+            torrent_info = lt.torrent_info(open(filepath, "rb").read())
```
cas--: When opening files in Python it is preferable to use a `with` statement to close the file as soon as it leaves the code block so we won't be accidentally leaving too many open file descriptors.

cas-- (commented):
Thanks for providing this bug fix, I wasn't aware of this libtorrent limitation

I do wonder how this change might impact the Deluge startup time if we need to roundtrip reading the file in Python then call torrent_info rather than letting libtorrent handle it. Do you have any example public torrents to test this with?

Since this is a less common problem an alternative might be to check for the error `metadata too large` and fallback to reading the file in Python. Probably something like this:

```python
        try:
            torrent_info = lt.torrent_info(filepath)
        except RuntimeError as ex:
            error = str(ex)
        else:
            return torrent_info

        if error == "metadata too large":
            # Workaround libtorrent unable to open large torrent files
            try:
                with open(filepath, "rb") as file:
                    torrent_info = lt.torrent_info(file.read())
            except (RuntimeError, OSError) as ex:
                error = str(ex)
            else:
                return torrent_info

        log.warning('Unable to open torrent file %s: %s', filepath, error)
```

deluge/core/torrentmanager.py
marcan: This is a myth. Python is a reference counted language and the fd is closed immediately after the read call, as the fd object is immediately dropped. Try this:

```
print(os.listdir("/proc/self/fd"))
open("/bin/ls", "rb").read()
print(os.listdir("/proc/self/fd"))
```

marcan:
I don't see how this would impact startup time. In both cases you are reading the file into a buffer and using it, and all that happens in CPython C internals after the patch. There is no material difference to what is happening, the only difference is whether it happens inside CPython or inside libtorrent.

I have this patched locally in a client with over 1500 torrents and startup time is not materially affected.

[This](https://files.catbox.moe/amxubt.torrent) is one affected torrent.

deluge/core/torrentmanager.py
cas--: I am aware that gc most times will close the fd however this is not guaranteed so when opening hundred or thousands of torrents I'd rather be sure we are closing the fds. For context it has happened in the past with this codebase.

An alternative to consider would be `pathlib`, which I am wanting use more in the code, so we could one-line it with `Path(filepath).read_bytes()`

cas--:
With this change we are bypassing the efficiency of libtorrent so there is likely some impact, either on memory or timings. I would really like to avoid any potential performance impacts just for edge-case torrents. Also if the issue is fixed or improved in future libtorrent releases then the original code would handle it.

deluge/core/torrentmanager.py
marcan: It's not a GC. Python is a reference-counted language and the file is closed immediately the moment the object goes out of scope, which is immediately in this case because the object is not even assigned to a variable nor stored anywhere. This is absolutely guaranteed. There is no place for the references to be stored, and therefore no way for the files to leak. It is completely impossible given the design of (C)Python.

Python does have a GC, but its only function is to break reference cycles, which is not the case here.

marcan:
This was already reported to libtorrent 11 years ago and they chose not to fix it, there is no reason they would fix it now. libtorrent isn't magically more efficient than Python at reading a file into a buffer of bytes. There is no reason to overcomplicate this at all.

```
# time python3 -c 'import os; [open(file, "rb").read() for file in os.listdir(".") if file.endswith(".torrent")]'
python3 -c   0.025 user 0.030 system 97% cpu (0.056 wasted time).
```

It takes 0.05 seconds to read over 1500 torrent files using Python (with warm cache, of course, so we're measuring the language and syscalls and not the actual disk), and that includes interpreter startup time. You are seriously overthinking this. It's fine.

There is also no memory impact because, as I explained, Python is a reference-counted language and the byte buffer from Python is immediately freed after each call into libtorrent.

marcan:
Let's actually benchmark the libtorrent use case:

```
# python3 -m timeit -n 10 -s 'import os,libtorrent as lt' '[lt.torrent_info(file) for file in os.listdir(".") if file.endswith(".torrent")]'
10 loops, best of 5: 295 msec per loop
# python3 -m timeit -n 10 -s 'import os,libtorrent as lt' '[lt.torrent_info(open(file, "rb").read()) for file in os.listdir(".") if file.endswith(".torrent")]'
10 loops, best of 5: 304 msec per loop
```

9 millisecond difference for 1500 torrents. Do we really need to overcomplicate the code and make it more fragile (the error might change in the future) to save 9 milliseconds per 1500 torrents? I don't think so.

deluge/core/torrentmanager.py
cas--: Let's not nitpick, RC is a form of GC. It is useful to hear about these technical details but my understanding from other sources that it is not guaranteed, especially with other implementation of Python.

https://docs.python.org/3/reference/datamodel.html
> Do not depend on immediate finalization of objects when they become unreachable (so you should always close files explicitly).

It has always been best practice in Python to wrap open calls in a with statement, there is no penalty here with doing so. I ask that you please honour my request for this change

cas--:
Thank you for taking the time to benchmark that does quell the concerns I have with the change. I do need to ask the questions to check possible problems like this have been considered and addressed.

deluge/core/torrentmanager.py
marcan: Non-CPython implementations (like PyPy) could use lazy GC, yes, just not CPython.

I switched it to pathlib.

cas--:
Thanks, appreciate the fix

PR #484, opened 2025-06-07
Store Hashes of Daemon User Passwords

adenwuts:
Fixes [#2442 - Plaintext auth passwords](https://dev.deluge-torrent.org/ticket/2442) by hashing passwords daemon-side instead of storing all credentials as plaintext, primarily with the aim of reducing issues that may arise from reused passwords being stored in plaintext.

The localclient user is 'whitelisted' and still uses plaintext authentication to avoid breaking autologins.

### DIFF
--- a/deluge/core/authmanager.py
+++ b/deluge/core/authmanager.py
@@ -21,7 +21,12 @@ from deluge.common import (
     AUTH_LEVEL_READONLY,
     create_localclient_account,
 )
-from deluge.error import AuthenticationRequired, AuthManagerError, BadLoginError
+from deluge.error import (
+    AuthenticationRequired,
+    AuthManagerError,
+    BadLoginError,
+    InvalidHashError,
+)
 from deluge.security import check_password_hash, generate_password_hash
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


cas--:
I can see the benefit of this for the daemon, I don't see any downsides, other than users reverting to old daemon version and needing to reset password.

It's good to store the password hash with the per-user salt but should also store the method and parameters since those might change in future. It would also guarantee you are checking a hashed password if it starts with `$<method>$...`

There is some inspiration in passlib but [werkzeug.security](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/security.py) has a much cleaner implementation, maybe we could just copy their code (removing pbkdf2) to deluge/security.py (since common util rather than specific to core.auth), adding a BSD attribution.

One other thought, should we migrate existing plaintext passwords to hashed when user logs in?

adenwuts:
I've ported it across to use a modified version of the Werkzeug implementation to try and retain better compatibility with the existing `auth` file structure.

I played around with implementing a forced migration to hashed passwords but it would likely require some rethinking of the Account class or getting even more enthusiastic with the string parsing, which is probably better suited to a future PR. Instead I've opted to fallback to plaintext validation on a failure to parse the hash parameters (this isn't ideal but it makes this a non-breaking change while keeping the diff reasonable).

cas--:
This looks great! I'm happy to leave forced migration as future task so I'll try and make time this weekend to test and merge this

adenwuts:
> This looks great! I'm happy to leave forced migration as future task so I'll try and make time this weekend to test and merge this

Much appreciated! Let me know if you find anything I should change :smile:

cas--:
Thanks for this. I've make a few tweaks to the implementation such as 
[use PHC string format to store password hashes](https://github.com/deluge-torrent/deluge/commit/8d9c7f6b72cbf3c106d7cbe1cadb1ab1e3adf274) and [validate localclient before password hash](https://github.com/deluge-torrent/deluge/commit/8a8bf0b3d932c62d98ebd34f1cb91b363e62177e) and added tests which uncovered some minor bugs

PR #485, opened 2025-06-12
Fix pkg_resources deprecation warning in deluge-console

daanblom:
Replace pkg_resources with importlib.metadata to address deprecation warning. pkg_resources is slated for removal in 2025-11-30. This change adds importlib.metadata as the primary method for accessing entry points with fallbacks for compatibility.

### DIFF
--- a/deluge/ui/ui_entry.py
+++ b/deluge/ui/ui_entry.py
@@ -17,7 +17,21 @@ import logging
 import os
 import sys
 
-import pkg_resources
+try:
+    # Use importlib.metadata for Python 3.8+
+    from importlib.metadata import entry_points
+except ImportError:
+    # Fallback for Python < 3.8
+    try:
+        from importlib_metadata import entry_points
+    except ImportError:
+        # Last resort fallback to pkg_resources
+        import pkg_resources
+        _use_pkg_resources = True
+    else:
+        _use_pkg_resources = False
+else:
+    _use_pkg_resources = False
 
 import deluge.common
 import deluge.configmanager
@@ -35,11 +49,35 @@ def start_ui():
 
     # Get the registered UI entry points
     ui_entrypoints = {}
-    for entrypoint in pkg_resources.iter_entry_points('deluge.ui'):
+    if _use_pkg_resources:
+        # Legacy method using pkg_resources
+        for entrypoint in pkg_resources.iter_entry_points('deluge.ui'):
+            try:
+                ui_entrypoints[entrypoint.name] = entrypoint.load()
+            except ImportError:
+                # Unable to load entrypoint so skip adding it.
+                pass
+    else:
+        # Modern method using importlib.metadata
         try:
-            ui_entrypoints[entrypoint.name] = entrypoint.load()
-        except ImportError:
-            # Unable to load entrypoint so skip adding it.
+            eps = entry_points(group='deluge.ui')
+            # Handle different return types between Python versions
+            if hasattr(eps, 'select'):  # Python 3.10+
+                for entrypoint in eps.select():
+                    try:
+                        ui_entrypoints[entrypoint.name] = entrypoint.load()
+                    except ImportError:
+                        # Unable to load entrypoint so skip adding it.
+                        pass
+            else:  # Python 3.8, 3.9
+                for entrypoint in eps:
+                    try:
+                        ui_entrypoints[entrypoint.name] = entrypoint.load()
+                    except ImportError:
+                        # Unable to load entrypoint so skip adding it.
+                        pass
+        except Exception:
+            # Fallback to empty dict if something goes wrong
             pass
 
     ui_titles = sorted(ui_entrypoints)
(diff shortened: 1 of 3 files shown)


deluge/ui/ui_entry.py
```diff
@@ -17,7 +17,21 @@
 import os
 import sys
 
-import pkg_resources
+try:
+    # Use importlib.metadata for Python 3.8+
+    from importlib.metadata import entry_points
+except ImportError:
+    # Fallback for Python < 3.8
```
cas--: I have just dropped support for Python <= 3.8 so this can be removed

deluge/ui/ui_entry.py
```diff
@@ -35,11 +49,35 @@ def start_ui():
...
+            except ImportError:
+                # Unable to load entrypoint so skip adding it.
+                pass
+    else:
+        # Modern method using importlib.metadata
         try:
-            ui_entrypoints[entrypoint.name] = entrypoint.load()
-        except ImportError:
-            # Unable to load entrypoint so skip adding it.
+            eps = entry_points(group='deluge.ui')
```
cas--: The `group` attribute is passed down to `select` so later call is redundant. Also `group` is a feature of Python 3.10 so this does break on Python 3.9. 

We can simplify quite a bit by EAFP, I think the following should work

```
try:
   entrypoints = entry_points(group='deluge.ui')
except TypeError:
   try:
      entrypoints = list(set(entry_points()['deluge.ui']))
   except KeyError:
      entrypoints = []

for entrypoint in entrypoints:
...
```

I have update the CI test to use 3.9 and 3.13 so should catch any problems

cas-- (changes requested):

deluge/ui/ui_entry.py
adenwuts: @cas-- Python 3.9 is EOL in October. Can we drop support for it in favour of 3.10?

loqs:
@daanblom given that pkg_resources has now been removed in setuptools 82 should the scope of this pull request be widened to cover plugins as well. Such as in https://github.com/deluge-torrent/deluge/pull/498? There is also https://github.com/deluge-torrent/deluge/pull/488 but it appears to makes unrelated changes.

daanblom:
> @daanblom given that pkg_resources has now been removed in setuptools 82 should the scope of this pull request be widened to cover plugins as well. Such as in #498? There is also #488 but it appears to makes unrelated changes.

With setuptools 82 removing pkg_resources, this definitely needs addressing more broadly.

That said, maybe don't expand this PR to cover plugins as well. The plugin system (especially legacy egg support) likely deserves its own focused PR and discussion (e.g. #488)?

daanblom:
I've made the changes @cas-- requested, but might have made a mistake when trying to push them? i believe the updated files should appear in this thread, but they can only be seen on https://github.com/daanblom/deluge ? 

excuse me for any inconvenience

loqs:
> I've made the changes @cas-- requested, but might have made a mistake when trying to push them? i believe the updated files should appear in this thread, but they can only be seen on https://github.com/daanblom/deluge ?

What is the identifier of the commit that made the changes cas-- requested? Did you add another commit to the series and perform a push or rebase and force push?

daanblom:
> What is the identifier of the commit that made the changes cas-- requested? Did you add another commit to the series and perform a push or rebase and force push?

i think its 75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e 

https://github.com/daanblom/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e

i created another commit and pushed that

DjLegolas:
> i think its [75f5d7b](https://github.com/deluge-torrent/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e)
> 
> [daanblom@75f5d7b](https://github.com/daanblom/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e)
> 
> i created another commit and pushed that

@daanblom  you need to push it to the same branch, `daanblom:fix-pkg-resources-deprecation` for it to be shown here
this commit is on your `develop` branch

loqs:
> i think its [75f5d7b](https://github.com/deluge-torrent/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e)
> 
> [daanblom@75f5d7b](https://github.com/daanblom/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e)

[75f5d7b](https://github.com/deluge-torrent/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e) contains the fallback for python < 3.8 that cas requested removing. As deluge now only supports 3.10 plus you can also remove the python 3.8 and 3.9 fallback code.

daanblom:
[75f5d7b](https://github.com/deluge-torrent/deluge/commit/75f5d7bc8c789ab0c3936f6b8cab8cf5a8f4c99e) contains the fallback for python < 3.8 that cas requested removing. As deluge now only supports 3.10 plus you can also remove the python 3.8 and 3.9 fallback code.

thanks for pointing that out. I merged my develop branch into the fix-pkg-resources-deprecation one and removed the fallback code

PR #487, opened 2025-08-25
Cleanup type annotations and linting for Authmanager

adenwuts:
Adds better type annotations to authmanager for easier extension/development/maintenance. Also cleaned up a few implicit string concatenations used in logging and added override flags where appropriate.

Hoping to use this as a basis for integrating an automated upgrade path for plaintext passwords-->hashes as disccussed in #484.

No functional changes, all tests still passing.

### DIFF
--- a/deluge/common.py
+++ b/deluge/common.py
@@ -27,7 +27,7 @@ from datetime import datetime
 from importlib import resources
 from io import BytesIO
 from pathlib import Path
-from typing import Literal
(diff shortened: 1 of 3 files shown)


adenwuts:
Tested with wrong Python version :cry: 
I'll fix this up later today.

adenwuts:
tests now passing using `python3.9`

PR #488, opened 2025-09-30
Remove package_resources dependency

adenwuts:
Use of `pkg_resources` is deprecated in favour of
`importlib.resources` and `importlib.metadata` according to https://setuptools.pypa.io/en/latest/pkg_resources.html. Running Deluge currently triggers warnings that the `pkg_resources` library is likely to be removed as early as October 2025.

Our plugin system still relies on `.egg` files, and has relied on `pkg_resources` to expose plugin metadata and resource files. We can use the `pkg_info` to pull the metadata from the .egg files and importlib for the rest. This allows for users to continue using legacy plugins while a more modern plugin framework is developed.

This does unfortunately necessitate the inclusion of the `pkg_info` library, however this seems like a reasonable trade off given that it is (currently) well maintained. There may also be other possible solutions using the importlib.metadata library or reimplementing a minimal version of the `pkg_info` `BDist` functionality.

### DIFF
--- a/deluge/pluginmanagerbase.py
+++ b/deluge/pluginmanagerbase.py
@@ -10,10 +10,13 @@
 """PluginManagerBase"""
 
 import email
+import importlib.metadata
 import logging
 import os.path
+import sys
+from typing import NamedTuple
 
-import pkg_resources
+import pkginfo
 from twisted.internet import defer
 from twisted.python.failure import Failure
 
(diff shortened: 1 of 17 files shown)


loqs:
Have you looked at the reuse of the existing resource_filename which avoids the need to create deluge/plugins/backports.py and supports automatic resource cleanup in https://github.com/deluge-torrent/deluge/pull/498?

cas--:
@adenwuts Thanks for this, it inspired me to actually bring the egg plugin discovery and loading into the plugin manager. There are several different issues needing solved with existing plugins, such as resource files and non-zip eggs so needed a good set of tests to keep on top of it all.

PR #489, opened 2025-11-14
Change default daemon port

doadin:
Current default port is part of default reserv on windows
netsh int ipv4 show excludedportrange protocol=tcp

Protocol tcp Port Exclusion Ranges

Start Port    End Port
----------    --------
      2869        2869
      5357        5357
     28385       28385
     28390       28390
     49152       49251
     49679       49778
     49779       49878
     49879       49978
     50000       50059     *
     50060       50159
     50160       50259
     50260       50359
     50360       50459
     50460       50559
     58680       58779
     58780       58879
     63786       63885
     64228       64327
     64450       64549
     64550       64649
     64922       65021
     65028       65127
     65128       65227
     65228       65327
     65419       65518

### DIFF
--- a/deluge/core/preferencesmanager.py
+++ b/deluge/core/preferencesmanager.py
@@ -37,7 +37,7 @@ log = logging.getLogger(__name__)
 DEFAULT_PREFS = {
     'send_info': False,
     'info_sent': 0.0,
-    'daemon_port': 58846,
+    'daemon_port': 61000,
     'allow_remote': False,
     'pre_allocate_storage': False,
     'download_location': deluge.common.get_default_download_dir(),
--- a/deluge/core/rpcserver.py
+++ b/deluge/core/rpcserver.py
@@ -386,7 +386,7 @@ class RPCServer(component.Component):
     :type listen: bool
     """
 
-    def __init__(self, port=58846, interface='', allow_remote=False, listen=True):
+    def __init__(self, port=61000, interface='', allow_remote=False, listen=True):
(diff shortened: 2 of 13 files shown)


cas--:
Changing the default port would be disruptive to existing users and wouldn't reliably fix the problem anyway. There is no port we could switch to that is guaranteed to be safe on Windows with Hyper-V/Docker/WSL2 installed.

The cause is Hyper-V reserving random blocks of ports from the Windows dynamic range (49152+) on each boot. The fix needs to be applied by Windows users:

**Quick fix** (may recur after reboot):
```cmd
net stop winnat && net start winnat
```

**Permanent fix:**
```cmd
net stop winnat
netsh int ipv4 add excludedportrange protocol=tcp startport=58846 numberofports=1 store=persistent
net start winnat
```

That you felt we needed to change the default port suggests it is probably worth adding to the Windows docs?

Adding the `netsh` call to NSIS installer might also be useful:

```
Section "Install"
    ...
    nsExec::ExecToLog 'netsh int ipv4 add excludedportrange protocol=tcp startport=58846 numberofports=1 store=persistent'
SectionEnd

Section "Uninstall"
    ...
    nsExec::ExecToLog 'netsh int ipv4 delete excludedportrange protocol=tcp startport=58846 numberofports=1 store=persistent'
SectionEnd
```

doadin:
@cas-- sorry, my understanding was that 6100 was outside the range that hyper-v would select.

PR #493, opened 2026-02-06
Fix OpenSSL

doadin:


doadin:
https://github.com/deluge-torrent/deluge/commit/c7dc60571e1d07081f954315f1706d0d7f87e594

this change is not exactly right. while it does include ssl dlls they are the wrong ones. the currently published libtorrent wheels where made from openssl 1.x and what gets included is openssl 3.x

doadin:
thougt this was going to be simpler but here we are. for the openssl change the wheels currently published on pypi for libtorrent were built with openssl 1.x but pyinsatller pickes up openssl 3 likely due to openssl being removed in newer python builds. This probabbly wont be needed in next libtorrent wheel build but last one was a year ago now so who knows when that will happen.

The zope issue is beyond my understanding as far is the inerworkings but 7.2 stilll works. There was a post in twisted about droping zope but that too was in 2024 and here war are still using it so not sure where this puts deluge in the future.

cas--:
Thanks for trying to unpick this all, it'll be fine for now to not constantly upgrade to latest unless required. I'll try to find time to help with this

cas--:
I will probably push the changes soon but in the meantime this is how to get openssl working with libtorrent:

```yaml
      - name: Install OpenSSL 1.1.1
        shell: bash
        run: |
          cd $(mktemp -d)
          base_url="https://github.com/python/cpython-bin-deps/raw/refs/heads/openssl-bin/amd64"
          curl -fsSL -O "$base_url/libcrypto-1_1.dll"
          curl -fsSL -O "$base_url/libssl-1_1.dll"
          # Libtorrent searches for x64 suffixed DLLs
          cp libcrypto-1_1.dll libcrypto-1_1-x64.dll
          cp libssl-1_1.dll libssl-1_1-x64.dll
          # PATH not searched for DLL deps in Python 3.8+, so install to System32
          cp ./*.dll /c/Windows/System32/
```

doadin:
> I will probably push the changes soon but in the meantime this is how to get openssl working with libtorrent:
> 
> ```yaml
>       - name: Install OpenSSL 1.1.1
>         shell: bash
>         run: |
>           cd $(mktemp -d)
>           base_url="https://github.com/python/cpython-bin-deps/raw/refs/heads/openssl-bin/amd64"
>           curl -fsSL -O "$base_url/libcrypto-1_1.dll"
>           curl -fsSL -O "$base_url/libssl-1_1.dll"
>           # Libtorrent searches for x64 suffixed DLLs
>           cp libcrypto-1_1.dll libcrypto-1_1-x64.dll
>           cp libssl-1_1.dll libssl-1_1-x64.dll
>           # PATH not searched for DLL deps in Python 3.8+, so install to System32
>           cp ./*.dll /c/Windows/System32/
> ```

Sorry forgot to remove the twisted change before I left on a trip. However yea any way to install OpenSsl 1.x and add it to path should work.

doadin:
> I will probably push the changes soon but in the meantime this is how to get openssl working with libtorrent:
> 
> ```yaml
>       - name: Install OpenSSL 1.1.1
>         shell: bash
>         run: |
>           cd $(mktemp -d)
>           base_url="https://github.com/python/cpython-bin-deps/raw/refs/heads/openssl-bin/amd64"
>           curl -fsSL -O "$base_url/libcrypto-1_1.dll"
>           curl -fsSL -O "$base_url/libssl-1_1.dll"
>           # Libtorrent searches for x64 suffixed DLLs
>           cp libcrypto-1_1.dll libcrypto-1_1-x64.dll
>           cp libssl-1_1.dll libssl-1_1-x64.dll
>           # PATH not searched for DLL deps in Python 3.8+, so install to System32
>           cp ./*.dll /c/Windows/System32/
> ```

Does this account for x86 and x64 before we need this only for 64 but since the action runners dont have openssl 1.1 at all is it not needed for x86?

cas--:
I think dropping x86 is fine, I think the user base is pretty small and I remember there was a discussion about doing that anyway

If there is a sudden rush of users wanting 32-bit we could add it back in to release build ;)

PR #494, opened 2026-02-07
setup: Remove deprecated test, setup_requires and license classifier

ademuri:
- Remove 'test' command, which is disabled and unused.
- Remove deprecated 'License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)' classifier.

These changes resolve deprecation warnings from setuptools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ from distutils.command.install_data import install_data as _install_data
 from shutil import rmtree, which
 
 from setuptools import Command, find_packages, setup
-from setuptools.command.test import test as _test
 
 import msgfmt
 from version import get_version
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -536,7 +517,6 @@ def run(self):
 ]
 _package_data['deluge.ui.gtk3'] = ['glade/*.ui']
 
-setup_requires = ['setuptools', 'wheel']
```
cas--, line 539: Unless this is breaking something, we can keep it until properly drop setup.py

cas-- (changes requested):

setup.py
ademuri: Re-added.

ademuri:
The test failure seems to already occur at head on `develop`.

cas--:
Thanks, I've resolved the test flakiness too

kuraga:
On setup, take a look at https://dev.deluge-torrent.org/ticket/3542, please, too...

PR #500, opened 2026-03-07
Make mako optional in setup.py

ademuri:
This attempts to fix https://dev.deluge-torrent.org/ticket/3542. I'm not very familiar with the build system, but it seems like this is straightforward.


cas--:
Yeah I can't remember why I didn't do this before. It is more of a runtime dependency but probably was hesitant in case it broke packaging of webui. I'll have a think and merge this if it makes sense

PR #506, opened 2026-03-30
few minor fixes to catch errors

doadin:


nopoz:
The two core changes cancel out. With `__getitem__` returning `None` instead of raising, the new `except KeyError` in `create_torrent_status` is unreachable: `torrent` is `None`, and the next line raises an uncaught `AttributeError: 'NoneType' object has no attribute 'get_status'`.

Applied the branch and called `create_torrent_status` with an id that is not in `torrents`:

    develop:  __getitem__ raises KeyError  ->  create_torrent_status returns {}
    this PR:  __getitem__ returns None     ->  AttributeError: 'NoneType' object has no attribute 'get_status'

That makes a handled case unhandled, on a path that races whenever a client polls status while torrents are being removed.

`log.debug('Attempted to get torrent object not in torrents', torrent_id)` has no format placeholder. With debug logging enabled it emits `TypeError: not all arguments converted during string formatting` and the message is never logged.

The bare `except:` in `torrent.py` also catches `KeyboardInterrupt` and `SystemExit`; `except Exception:` keeps the intent.

What would fix it: drop the `torrentmanager.py` change and keep the rest. The `create_torrent_status` restructure works as written once `__getitem__` still raises, the traceback still goes away, which is the point of the PR, and the contract stays intact for the other callers that index the manager and immediately call a method on the result.

If you want a log line for the missing torrent, putting it in the `except KeyError` in `create_torrent_status` gives the same information without touching `__getitem__`:

```python
except KeyError:
    log.debug('Torrent %s removed before status could be fetched', torrent_id)
    return {}
```

PR #514, opened 2026-07-17
replace deprecated calls in certificate generation

adenwuts:
Minimal patch to fix a crash on launch on systems without deprecated pyopenssl versions. Instead rely on the cryptography equivalents.


loqs:
@adenwuts will you also be doing a patch for libtorrent 2.1?

